PR #473, opened 2014-01-21
diffrent level logs display diffrent color

cloudaice:
对不同级别的日志在终端可以显示为不用的颜色，这里仅仅是对日志显示的内容进行的颜色转换，但是对日志本身的时间前缀没有做颜色转换。在这里讨论一下是否需要对时间前缀也做相应的颜色转换。

### DIFF
--- a/logs/console.go
+++ b/logs/console.go
@@ -6,6 +6,25 @@ import (
 	"os"
 )
 
+type Brush func(string) string
+
+func NewBrush(color string) Brush {
+	pre := "\033["
+	reset := "\033[0m"
+	return func(text string) string {
+		return pre + color + "m" + text + reset
+	}
+}
+
+var colors = []Brush{
+	NewBrush("1;36"), // Trace      cyan
+	NewBrush("1;34"), // Debug      blue
+	NewBrush("1;32"), // Info       green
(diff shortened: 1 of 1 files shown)


astaxie:
> Hi, I am using beego log lib, but I found that the debug color is the same to info,
> and i don't found a way to set it differently, May i ask here, Is there a way to set the debug color to a different color?

what's system?

zwpaper:
@astaxie Both macOS 10.11 with oh-my-zsh and CentOS 7 show the same blue color for debug and info.

astaxie:
@zwpaper ah, it's use the same color. https://github.com/astaxie/beego/blob/master/logs/console.go#L43-L44

zwpaper:
@astaxie oops...The debug, info, error color is the most useful color for me (and maybe somebody else), maybe we can change it to a different color to mark the difference? What your idea?

astaxie:
@zwpaper fix it. https://github.com/astaxie/beego/commit/ae0a75c464e3373c89f296e698a1679f80e566f9

I use a background blue for the debug.

zwpaper:
@astaxie Thanks

PR #987, opened 2015-01-05
Update parser.go

wkrause13:
Log message grammar for un-updated packages  is now correct. Another option would be "has not been updated", but that may confuse some people into thinking the command should have updated something but failed to do so.


astaxie:
please send the pull request to develop branch. Thanks

PR #993, opened 2015-01-06
Add buffered output and flush into output context

athurg:
Add buffered output and flush support.
So you can flush content to the client while the application is running.
Like this:
```GO
func (this *MyController) Get() {
	this.Ctx.Output.Header("Content-Type", "text/html; charset=utf-8")

	for i := 1024; i>0; i-- {
		this.Ctx.Output.Write([]byte(" "))
	}

	this.Ctx.Output.Flush()

	for i := 10; i>0; i-- {
		this.Ctx.Output.Write([]byte(fmt.Sprintf("%03d...<br/>\n",i)))
		this.Ctx.Output.Flush()
		time.Sleep(1000*time.Millisecond)
	}
	this.StopRun()
}
```

### DIFF
--- a/context/output.go
+++ b/context/output.go
@@ -49,6 +49,22 @@ func (output *BeegoOutput) Header(key, val string) {
 	output.Context.ResponseWriter.Header().Set(key, val)
 }
 
+// Flush current output buffer to the client
+func (output *BeegoOutput) Flush() {
+	flusher,ok := output.Context.ResponseWriter.(http.Flusher)
+	if ok {
+		flusher.Flush()
+	}
+}
+
+// Write response content into the buffer directly
+func (output *BeegoOutput) Write(content []byte) {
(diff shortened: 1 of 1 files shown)


context/output.go
```diff
@@ -49,6 +49,22 @@ func (output *BeegoOutput) Header(key, val string) {
...
+// Flush current output buffer to the client
+func (output *BeegoOutput) Flush() {
+	flusher,ok := output.Context.ResponseWriter.(http.Flusher)
+	if ok {
+		flusher.Flush()
+	}
+}
+
+// Write response content into the buffer directly
+func (output *BeegoOutput) Write(content []byte) {
```
astaxie: There's no need to add this function. `output.Context.ResponseWriter`already has the Write method

astaxie:
has any update?

PR #1000, opened 2015-01-10
Add GroupBy to QuerySeter

pdf:
Adds support for GroupBy to QuerySeter SELECT operations.

### DIFF
--- a/orm/db.go
+++ b/orm/db.go
@@ -800,6 +800,7 @@ func (d *dbBase) ReadBatch(q dbQuerier, qs *querySet, mi *modelInfo, cond *Condi
 	tables.parseRelated(qs.related, qs.relDepth)
 
 	where, args := tables.getCondSql(cond, false, tz)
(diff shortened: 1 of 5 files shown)


astaxie:
https://drone.io/github.com/astaxie/beego/905 build error

pdf:
@astaxie sorry about that, I think I was having some problems with the postgres setup in my dev environment when I wrote this.  The problem is that postgres doesn't support partial group by, whereas mysql and sqlite do.  As such, some of the tests here do not produce valid queries for postgres, yet the code is behaving as expected.  I'm not sure what the best way to deal with this is, should I conditionally execute those tests only for mysql and sqlite, so that we maintain coverage?  I don't think I can exercise that code in a way that will be valid for every DBMS.

y2mao:
Seter or Setter?

pdf:
@y2mao `QuerySeter`, the struct is `QuerySet`, and the Go convention is to add `er` when creating an interface, hence `QuerySeter`.

PR #1016, opened 2015-01-28
set exist to true when value is integer

kykl:
Doing so will prevent it from saying PK doesn't exist when integer key is less than or equal 0.


astaxie:
IMO, PK should more than 0

PR #1025, opened 2015-02-04
Update task tpl 

kongjian:
1.task page can't run task, see also: http://stackoverflow.com/questions/27922410/beego-admin-task-page-not-working
2.task page can't see spec


kongjian:
by the way, use <br> make web page looks nice

Issue #1040, opened 2015-02-16
json api format


alexjp:
Hi,

I am kinda a newbie at golang and i have a webapp in ember data.
I successfully created a backend in api2go package ( creates a restfull server with the jsonapi format ) but i would like to try to use beego as the backend.

Is there a way to serve json like specified in : http://jsonapi.org/ ?

Thanks !

astaxie:
you can try with this way:

    go get github.com/beego/bee

after you install the bee, add you `$GOPATH/bin` into you `PATH`(`export PATH="$PATH:$GOPATH/bin"`). then get into the `$GOPATH/src` run the command:

    $ bee api myapi
    $ cd myapi
    $ bee run -gendoc=true -downdoc=true

then open you browser open the url: `http://127.0.0.1:8080/swagger/swagger-1/`

hope you enjoy it.

alexjp:
Hi astaxie, thanks for the reply. Beego with swagger is really nice, with also the router logging in the console.

But i think i didnt explain very well, the issue is not documenting the API, but formatting the json response. Using beego, the response, as an example, for get:
{
objectID : [{ object1, object2, ...}] }

What jsonapi and ember data expect is : 
{
objects: [{ object1, object2 ...}] }

in this case, is almost simillar, but then there are other cases specified in jsonapi.org and emberdata.

Is there a specific section of beego, that i could hack to format the json, and then be using in all application?

I would like to do something similar to beego as this was to sails : https://github.com/mphasize/sails-generate-ember-blueprints

Thanks !

astaxie:
I don't know what do you want? format the response with json? do you want this? http://beego.me/docs/mvc/controller/jsonxml.md#json%2C-xml-and-jsonp

evandertino:
@alexjp The functionality you want, [json api format](http://jsonapi.org/) cannot be achieved with beego. Beego is flexible and lets you design your api structure in whichever format you like.

What you should be asking @astaxie is to introduce the 
header **Content-Type: application/vnd.api+json** which json api [format uses](http://jsonapi.org/#mime-types), so that you can easily get the header in the output. Then hack your way all through.

The jsonapi format is very strict, and i have sent @alexjp a [PR](https://github.com/astaxie/beego/pull/1065) to help you integrate the format.

When i get time i will send a PR with a package handling the format to beego

alexjp:
@evandertino thanks.

I will take a look at it this weekend. 
[...]

PR #1065, opened 2015-03-05
Added support for the JSONAPI header

evandertino:
The format also supports its own official [extensions](http://jsonapi.org/extensions/#official-extensions) and [custom extensions](http://jsonapi.org/extensions/#custom-extensions)

Created to close the [issue 1040](https://github.com/astaxie/beego/issues/1040) and to support those creating APIs following this [format](http://jsonapi.org)

### DIFF
--- a/context/output.go
+++ b/context/output.go
@@ -207,6 +207,42 @@ func (output *BeegoOutput) Json(data interface{}, hasIndent bool, coding bool) e
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


astaxie:
I am still not familiar with the JSONAPI. So maybe a complete example will make sense

astaxie:
send to the master PR will close

PR #1070, opened 2015-03-06
add JsonBody

fugr:
easy to new a request with body encoding by JSON.

### DIFF
--- a/httplib/httplib.go
+++ b/httplib/httplib.go
@@ -253,6 +253,22 @@ func (b *BeegoHttpRequest) Body(data interface{}) *BeegoHttpRequest {
 	return b
 }
 
+
+// JsonBody adds request raw body encoding by JSON.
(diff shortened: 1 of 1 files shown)


mattn:
please merge this into master because [bat](https://github.com/astaxie/bat) is not possible to build.

astaxie:
@mattn it's already merger into beego's develop branch. because bat is still in process, if you want to try you can get into the beego, and `git checkout develop` then you can build bat.

mattn:
okay, but i love `go get`. :)

astaxie:
yea, I will release beego ASAP. Now it's still in develop. because I want to add proxy and auth into bat. So maybe still need to change the httplib.

astaxie:
@mattn I just remove the beego dependence. move httplib to the bat. When it stable can move to beego

mattn:
thanks

PR #1092, opened 2015-03-25
For enhancing performance, check log level before fmt.Sprintf()

toalexjin:
If writing log messages frequently, it's better to check log level before `fmt.Sprintf()`, especially for BeeLogger.Debug().


astaxie:
please send the PR to develop branch. Thanks

PR #1093, opened 2015-03-25
Added to input.go: AcceptsHtml/Xml/Json functions

byrnedo:
Check the header against a regex for simpler mult-content-type handling.

Ie. in controller

    if c.Ctx.Input.AcceptsHtml() {
    ...do html stuff

If there's already something in there to do this please let me know


astaxie:
please send the PR to develop branch, thanks

PR #1094, opened 2015-03-26
For enhancing performance, check log level before fmt.Sprintf()

toalexjin:
If writing log messages frequently, it's better to check log level before `fmt.Sprintf()`, especially for `BeeLogger.Debug()`.


astaxie:
cool, but maybe we can improve even better, `bl.writerMsg(level, msg)` change to `bl.writerMsg(msg)`, because we don't need level.

toalexjin:
I made a second commit, remove checking log level from writerMsg(). However, input parameter "logLevel" is still required by writerMsg() because a new logMsg object will be created by the function.

astaxie:
great, Thanks

PR #1095, opened 2015-03-26
Added AcceptHtml/Json/Xml function to input

byrnedo:
Checks the header against a regex for simpler mult-content-type handling. 

In controller

    if this.Ctx.Input.AcceptsHtml {
    ...

Might not be the best way to do this and I'm aware there are constants in controller.go for "application/Json" etc so maybe I should move those out somewhere common and use them in both?

### DIFF
--- a/context/input.go
+++ b/context/input.go
@@ -21,10 +21,19 @@ import (
 	"net/http"
 	"net/url"
 	"reflect"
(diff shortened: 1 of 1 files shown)


byrnedo:
Awesome :)

PR #1096, opened 2015-03-26
Set ErrorsMap to nil in Validation.Clear function

byrnedo:
A bit misleading that the errors map still hangs around after Validation.Clear


astaxie:
Thanks

PR #1103, opened 2015-03-31
Update output.go

ElvizLai:
fix cookie not work in IE


astaxie:
cool, That's intresting, I just find this article, http://mrcoles.com/blog/cookies-max-age-vs-expires/

PR #1113, opened 2015-04-03
Patch 2

ElvizLai:


astaxie:
> If you wants to use it as a session cookie, why adding an expire to it? It's life time should be equal explorer's life time.

http://mrcoles.com/blog/cookies-max-age-vs-expires/

IE 6 7 8 not support MaxAge

PR #1129, opened 2015-04-24
Add Recursively Validation support

dafang:
We are using the recursively validations, send this pull request to help others.

The details:

- Step1: Validate by v.Valid
- Step2: If pass on step1, then reflect obj's fields
- Step3: Do the Recursively validation to all struct or struct pointer fields


astaxie:
please send the PR to the develop branch

PR #1141, opened 2015-05-03
Improve and refactor ControllerRegistor.ServeHTTP.

cr7pt0gr4ph7:


router.go
```diff
@@ -701,69 +693,58 @@ func (p *ControllerRegistor) ServeHTTP(rw http.ResponseWriter, r *http.Request)
...
-	if !findrouter {
+	// a "not found" exception is thrown in case no match was found
+	if !matchFound {
 		exception("404", context)
 		goto Admin
 	}
 
-	if findrouter {
-		//execute middleware filters
+	if matchFound && (runRouter == nil || runMethod == "") {
```
astaxie: this logic is not correct

cr7pt0gr4ph7: You are right, that is an error. I'm preparing a fix.

cr7pt0gr4ph7:
I fixed the error you mentioned in 442d3128b5f53fdfa6fb91991bd8eb2097c1a88c. The second commit (b081cf44a22d170dd4c3b2f930323811d56e5b1d) contains improvements to the documentation comments, but no code changes.

router.go
```diff
@@ -701,69 +693,62 @@ func (p *ControllerRegistor) ServeHTTP(rw http.ResponseWriter, r *http.Request)
...
+				runMethod = determineRouterMethod(routerInfo, context)
 			}
 		}
 
-		// also defined runrouter & runMethod from filter
-		if !isRunable {
-			//Invoke the request handler
-			vc := reflect.New(runrouter)
+		// internal assertion to catch bugs
+		if (runRouter != nil && runMethod == "") || (runRouter == nil && runMethod != "") {
```
astaxie: `isRunable` is improtant here. if remove it will always panic when the router is `routerTypeRESTFul` and `routerTypeHandler`

astaxie: @cr7pt0gr4ph7 do you have any update for this?

cr7pt0gr4ph7: Sorry for the break. I'm planning to review and extend the tests in `router_test.go` to catch this, and to fix the problem using the tests.

cr7pt0gr4ph7:
I've also got some upates for the documentation comments in `router.go`, `beego.go` and `app.go`. Shall I include them in this pull request, or in a separate one?

cr7pt0gr4ph7:
I hit a problem/inconsistency in the current code: The filter execution sequence is like this: `BeforeStatic -> BeforeRouter -> BeforeHandler -> (Handler) -> AfterExec -> FinishRouter`.

**TL;DR:** Should `AfterExec` and `FinishRouter` filters be executed if the normal handler has already written something to the output? (If no: What is the use case for `AfterExec` and `FinishRouter` filters, if they're almost never executed?)

### Problem description

If a filter outputs something, the later stages of the sequence are not executed - **except for one problematic case**: When the `(handler)` outputs something, the first filter of `AfterExec` is executed.
If that filters `returnOnOutput` is `true`, the sequence is aborted as usual, as if the filter had written something to the output by itself.

From a users perspective, this behavior could be seen as inconsistent, and thus does not look like it is intended. So my question is:

* What is the correct behavior when the handler has written something to the output?
  (And also: What is the intended use case for `AfterExec` and `FinishRouter` handlers?)

#### Possible solutions

There are two possible behaviors (although hybrid aproaches using additional flags are possible):
  * **(a)** Always execute the `AfterExec` and `FinishRouter` handlers, even if the main handler has already written to the output.
  * **(b)** Don't execute the `AfterExec` and `FinishRouter` handlers if the main handler has written to the output.

Option **(b)** is more consistent with the current behavior of the `BeforeStatic`/`BeforeRouter`/`BeforeExec` handlers, as well as with the current behavior of `AfterExec`/`FinishRouter` (except for the bug described above).

But when viewed from another perspective, option **(b)** is almost useless in practice, because the filters are almost never going to be executed (except in the (improbable) case the handler has written nothing) - so option **(a)** might make more sense here.

#### Additional context
I discovered the inconsistency while looking at this particular piece of code in `TestFilterFinishRouter` in `router_test.go`:

```go
mux := NewControllerRegister()
mux.InsertFilter(url, BeforeRouter, beegoFilterNoOutput)
mux.InsertFilter(url, BeforeExec, beegoFilterNoOutput)
mux.InsertFilter(url, AfterExec, beegoFilterNoOutput)
mux.InsertFilter(url, FinishRouter, beegoFinishRouter1)
mux.Get(url, beegoFilterFunc)

rw, r := testRequest("GET", url)
mux.ServeHTTP(rw, r)

if strings.Contains(rw.Body.String(), "FinishRouter1") == true {
    t.Errorf(testName + " FinishRouter did not run")
}
// ...
```

The code looks like it wants to validate that `FinishRouter1` is executed, but then proceeds on to validate the opposite of that. It is mostly the error message that raised my suspicion.

router.go
astaxie: That should be great.

astaxie:
> I've also got some upates for the documentation comments in router.go, beego.go and app.go. Shall I include them in this pull request, or in a separate one?

could you send in another PR, that will be easy for me to review the code

astaxie:
@cr7pt0gr4ph7 

http://godoc.org/github.com/astaxie/beego#InsertFilter

InserFilter has the fourth params `bool`. 

The bool params is for setting the returnOnOutput value (false allows multiple filters to execute)

So when execute the filter, it will depend on the filterR.returnOnOutput
https://github.com/astaxie/beego/blob/master/router.go#L617

cr7pt0gr4ph7:
I'll send the second PR containing documentation improvements when this one is done (because it would be difficult to rebase the changes).

To your second comment: I know about `InsertFilter` and the `returnOnOutput` parameter. The question is: Should the first `AfterExec` filter be executed if the controller has written something to the output?

cr7pt0gr4ph7:
Runnable demonstration (The expected output is written below the `// Output:` comments. The output currently given by beego is written after the `// Actual output:` comments).

```go
package beego_test

import (
	"fmt"
	"net/http"
	"net/http/httptest"

	"github.com/astaxie/beego"
	"github.com/astaxie/beego/context"
)

func Example_No_ReturnOnOutput() {
	url := "/someUrl"
	rw, r := testRequest("GET", url)

	mux := beego.NewControllerRegister()
	mux.Get(url, beegoHandlerFunc)
	mux.InsertFilter(url, beego.AfterExec, beegoAfterExec1, false) // NOTE: returnOnOutput = false
	mux.InsertFilter(url, beego.AfterExec, beegoAfterExec2, false) // NOTE: returnOnOutput = false
	mux.ServeHTTP(rw, r)

	// Actual output:
	// |Handler
	// |AfterExec1
	// |AfterExec2

	// Output:
	// |Handler
}

func Example_With_ReturnOnOutput() {
	url := "/someUrl"
	rw, r := testRequest("GET", url)

	mux := beego.NewControllerRegister()
	mux.Get(url, beegoHandlerFunc)
	mux.InsertFilter(url, beego.AfterExec, beegoAfterExec1) // NOTE: returnOnOutput = true
	mux.InsertFilter(url, beego.AfterExec, beegoAfterExec2) // NOTE: returnOnOutput = true
	mux.ServeHTTP(rw, r)

	// Actual output:
	// |Handler
	// |AfterExec1

	// Output:
	// |Handler
}

func testRequest(method, path string) (*httptest.ResponseRecorder, *http.Request) {
	request, _ := http.NewRequest(method, path, nil)
	recorder := httptest.NewRecorder()

	return recorder, request
}

func beegoHandlerFunc(ctx *context.Context) {
	fmt.Println("|Handler")
	ctx.WriteString("Some output from handler")
}
func beegoAfterExec1(ctx *context.Context) {
	fmt.Println("|AfterExec1")
}
func beegoAfterExec2(ctx *context.Context) {
	fmt.Println("|AfterExec2")
}
```

router.go
cr7pt0gr4ph7: I don't get panic for `routerTypeRESTFul` and `routerTypeHandler`. They seem to work fine in the tests.

`isRunable` was set to true if and only if `runMethod` and `runRouter` were set to a non-nil value, so `isRunable` and `runMethod != "" && runRouter != nil` are equivalent.
(The only exception is the case where `context.RunMethod` and `context.RunController` are set, which was not handled by the old code).

To avoid a misunderstanding: The line `if isRunable { ...` is replaced by [line 737](https://github.com/astaxie/beego/pull/1141/files#diff-56c3bb2da6c5ecf6872e89a554944453R737) in the new code. Git's diff is confused by the additional assertion that I have inserted.

astaxie:
> When the (handler) outputs something, the first filter of AfterExec is executed

it looks this is a bug. it should not executed the first one if returnOnOutput is false

astaxie:
> What is the correct behavior when the handler has written something to the output?

if has something output, then depend on the returnOnOutput, if it's true just return, if it's false execute the filter.

> (And also: What is the intended use case for AfterExec and FinishRouter handlers?)

AfterExec and FinishRouter always used to release resources, record the request,  close the filehandle etc.

astaxie:
> (a) Always execute the AfterExec and FinishRouter handlers, even if the main handler has already written to the output.
> (b) Don't execute the AfterExec and FinishRouter handlers if the main handler has written to the output.

I would like to keep the b options. it will keep consistent with the current behavior of other filters

astaxie:
@cr7pt0gr4ph7 could you update the PR as soon as possible. I want to release the new version in the next two weeks. I hope we can release PR in this version. Thanks

cr7pt0gr4ph7:
I'll add the updated tests to this PR, but most likely not until the weekend. Sorry for the delay.

astaxie:
@cr7pt0gr4ph7 I will close the PR as there's something conflict now. Just send another PR. Thanks

PR #1142, opened 2015-05-04
Fix save config ini file

mlgd:
Save default section at first place


astaxie:
please send the PR to develop branch.

PR #1156, opened 2015-05-18
Let filter function get more params info from ctx.Input.Params

ziyel:


astaxie:

PR #1157, opened 2015-05-18
Let filter function get more params info from ctx.Input.Params

ziyel:


router.go
```diff
@@ -612,6 +612,9 @@ func (p *ControllerRegistor) ServeHTTP(rw http.ResponseWriter, r *http.Request)
 			if l, ok := p.filters[pos]; ok {
 				for _, filterR := range l {
 					if ok, p := filterR.ValidRouter(urlPath); ok {
+                        for k, v := range context.Input.Params {
+                            p[k] = v
+                        }
```
astaxie: do you use the gofmt?

ziyel: I have gofmt it just now.

PR #1163, opened 2015-05-20
Re: Fixed using RelatedSel with multiple params

Hepri:
Now in develop branch


astaxie:
conflicts

Hepri:
...
Could you apply this change by your self? It's important, but looks like I do something wrong

astaxie:
it's already fixed in develop branch.

PR #1210, opened 2015-06-13
fix multiple filters execute issue

oiooj:
The last filterFunc with returnOnOutput=ture won't be executed.

``` golang
	beego.InsertFilter("/*", beego.BeforeExec, FilterLoginCheck1,false)
	beego.InsertFilter("/*", beego.BeforeExec, FilterLoginCheck2)
```
In function  FilterLoginCheck1 , I'll write data via ResponseWriter, and w.started = true
FilterLoginCheck2 won't be executed, it should be.

### DIFF
--- a/router.go
+++ b/router.go
@@ -611,9 +611,6 @@ func (p *ControllerRegistor) ServeHTTP(rw http.ResponseWriter, r *http.Request) 
 		if p.enableFilter {
 			if l, ok := p.filters[pos]; ok {
 				for _, filterR := range l {
-					if filterR.returnOnOutput && w.started {
-						return true
-					}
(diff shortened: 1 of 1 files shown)


astaxie:
yea, that's what I design for.

```
    beego.InsertFilter("/*", beego.BeforeExec, FilterLoginCheck1,false)
    beego.InsertFilter("/*", beego.BeforeExec, FilterLoginCheck2)
```

if `FilterLoginCheck1` has output and `FilterLoginCheck2` should never execute.

oiooj:
oh,  if I have some filters that all should be executed,and just the last one can write data. right?

astaxie:
yea, if you has any filters has output and want to execute the next should set the fourth params as false.

PR #1246, opened 2015-07-06
fix example/chat i/o timeout

Skycrab:
// NextReader returns the next data message received from the peer. The
// returned messageType is either TextMessage or BinaryMessage.
func (c *Conn) NextReader() (messageType int, r io.Reader, err error) 

so, readPump() will 100% cause i/o timeout


astaxie:
please send the pull request to develop branch. thanks

Issue #1269, opened 2015-07-18
Package violates go documentation convention


wallclockbuilder:
Package documentation comments that are detailed should be put in its own `doc.go` file.

wallclockbuilder:
Heres the recommendation from the golang blog:
![golang_package_documentation_convention](https://cloud.githubusercontent.com/assets/2033945/8761205/36b636ae-2d33-11e5-981f-c9dbaa723c9b.png)

wallclockbuilder:
Beego's multi line introductory code lives in the `beego.go` file. Thats ok if it is a simple one line comment documentation. Its now 11 lines long. I suggest we move it to its own `doc.go` file as recommended.

PR #1270, opened 2015-07-18
Fix for #1269 Package documentation in its on `doc.go` file.

wallclockbuilder:


wallclockbuilder:
Fix for #1269 Package documentation in its on `doc.go` file.

wallclockbuilder:
Also fixes #1271

PR #1292, opened 2015-07-29
Implement Flush() and CloseNotify()

paulmynarsky:


router.go
```diff
@@ -922,6 +922,16 @@ func (w *responseWriter) WriteHeader(code int) {
 	w.writer.WriteHeader(code)
 }
 
+// Implements Flush()
+func (w *responseWriter) Flush() {
+        w.writer.(http.Flusher).Flush()
```
astaxie: whether we need to assert first, otherwise if the w.writer not implemented http.Flusher will crash. the same as below

astaxie:
@paulmynarsky please update the assert

PR #1320, opened 2015-08-23
Efficient String Concatenation

vefaliahmet:
Can you rewrite string concatenation in beego?

http://herman.asia/efficient-string-concatenation-in-go


astaxie:
please send the PR to develop branch. thanks

PR #1334, opened 2015-09-01
Update beego.go

ElvizLai:
Maybe the `Hard Coding` should have a higher priority

### DIFF
--- a/beego.go
+++ b/beego.go
@@ -259,6 +259,8 @@ func AddAPPStartHook(hf hookfunc) {
 // beego.Run(":8089")
 // beego.Run("127.0.0.1:8089")
 func Run(params ...string) {
+	initBeforeHttpRun()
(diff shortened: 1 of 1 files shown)


astaxie:
do you have an special example for this priority?

ElvizLai:
I have `HttpPort = 8888` in my app.conf ,but I want to change or fix my port with 8081(for some reason that the port can not be changed through config file), so I tried using:
`beego.Run(":8081")` , but that does not work, I read the source code and found this. I not quite sure this is a good feature, but someone may need this.

astaxie:
It's trick. sometimes we want configure has the top priority

ElvizLai:
Absolute right, that implement is not good enough, but for now, I can not find a better way to specify the port before beego.Run().

PR #1340, opened 2015-09-02
always use server's locale to parse date

onealtang:
When parsing the date without time, it's always using UTC date, which is
unexpected. If we want to use UTC date, it's recommend to set the
server's timezone as UTC, and keep the code flexible.


astaxie:
please send the PR to develop branch. Thanks

PR #1341, opened 2015-09-02
Update tree.go

ElvizLai:
router bug fix:  `beego.Router("/trash/?:id:int", &console.Trash{})` can only match with `/trash/123`, not work for `/trash` and `/trash/`


astaxie:
@ElvizLai could you also add the test case for the hotfix?

PR #1343, opened 2015-09-02
always use server's locale to parse date

onealtang:
When parsing the date without time, it's always using UTC date, which is
unexpected. suppose we get '2015-09-02' from page, it's always parsed as '2015-09-02 00:00:00 UTC' but by default, what we expect is '2015-09-02 00:00:00' at local time zone.

And If we want to use UTC date throughout the system, it's recommend to set all
server's timezone as UTC including DB.

### DIFF
--- a/templatefunc.go
+++ b/templatefunc.go
@@ -352,7 +352,7 @@ func ParseForm(form url.Values, obj interface{}) error {
 				if len(tags) > 1 {
 					format = tags[1]
 				}
-				t, err := time.Parse(format, value)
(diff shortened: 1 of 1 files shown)


thanzen:
It will add more confusions when people need to support multi time zones in their applications if we use local machine's setting.
I think we should always to use one format(utc) for time in beego, and adding custom translation based on specific  project's need(developer's choice).

astaxie:
@thanzen yea, it's trade-off. but in most case it just use the local timezone.

PR #1349, opened 2015-09-06
Update tree.go

ElvizLai:
fix routers for:
```
/topic/:id/?:auth
/topic/:id/?:auth:int
```


tree.go
```diff
@@ -208,6 +208,11 @@ func (t *Tree) addseg(segments []string, route interface{}, wildcards []string,
 		} else {
 			t.leaves = append(t.leaves, &leafInfo{runObject: route, wildcards: wildcards})
 		}
+		for i, v := range wildcards {
+			if v==":" {
```
astaxie: not format this line

PR #1378, opened 2015-09-22
orm: Fix for related models with string based pk

pjoe:
When using e.g. rel(fk) to a model with a pk of type string(32), it would result in trying to genreate field as varchar(0) in db, which obviously wouldn't work.

This PR fixes the issue by also taking the size of the modelinfo of the related models pk into account.


pjoe:
Gahhh should be against develolp ... when will I learn ... closing

PR #1379, opened 2015-09-22
orm: Fix handling of rel(fk) to model with string pk

pjoe:
When using e.g. rel(fk) to a model with a pk of type string(32), it would result in trying to genereate field as varchar(0) in db, which obviously wouldn't work.

This PR fixes the issue by also taking the size of the modelinfo of the related models pk into account.


wallclockbuilder:
:+1:

pjoe:
Any news on getting this reviewed/merged?

astaxie:
sorry for the later review

pjoe:
No worries just glad to get it in :)

谢谢你

PR #1401, opened 2015-10-16
Remove Unnecessary c.Ctx.ResponseWriter.WriteHeader(status)

codyayers:
The following error is thrown with this line in place: 

`http: multiple response.WriteHeader calls`


astaxie:
already fixed in develop branch.

PR #1406, opened 2015-10-26
XsrfFormHtml should also generate XSRF token

superhacker777:
I did not receive any comment on [my issue](/astaxie/beego/issues/1397), so I decided to make a pull request instead. It is very, very strange that you have to initialize `_xsrf_token` field by calling `XsrfToken()` function before calling `XsrfFormHtml()`.


astaxie:

PR #1413, opened 2015-11-02
bugfix: graceful failed when both enable http and https

FlamingTree:
when both enable http and https, graceful restart or upgrade failed. Reason:
1. https://github.com/astaxie/beego/blob/master/app.go#L90-L105 L90, L105 assign same app.Server to http and https graceServe.Server.
2. https://github.com/astaxie/beego/blob/master/grace/server.go#L263 srvPtr.Server.Addr has both same value eventually, orderArgs's index 0 element will not be assigned and remain empty string. fork process's args will be -socketorder=,:10443

### DIFF
--- a/app.go
+++ b/app.go
@@ -78,6 +78,7 @@ func (app *App) Run() {
 		}
 	} else {
 		if Graceful {
+			httpsAddr := addr
 			app.Server.Addr = addr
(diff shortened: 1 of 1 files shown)


astaxie:

PR #1422, opened 2015-11-10
fix postgres syntax error during migration

WnP:
when doing migration using bee tool and postgresql as database postgresql return the following error:

```
pq: syntax error at or near "`"
```
This patch fix this error

### DIFF
--- a/migration/migration.go
+++ b/migration/migration.go
@@ -101,12 +101,10 @@ func (m *Migration) Exec(name, status string) error {
 }
 
 func (m *Migration) addOrUpdateRecord(name, status string) error {
-	var err error
-	var p orm.RawPreparer
 	o := orm.NewOrm()
 	if status == "down" {
 		status = "rollback"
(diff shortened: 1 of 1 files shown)


migration/migration.go
```diff
@@ -101,10 +101,17 @@ func (m *Migration) Exec(name, status string) error {
 }
 
 func (m *Migration) addOrUpdateRecord(name, status string) error {
+	var err error
+	var p orm.RawPreparer
 	o := orm.NewOrm()
 	if status == "down" {
 		status = "rollback"
-		p, err := o.Raw("update migrations set `status` = ?, `rollback_statements` = ?, `created_at` = ? where name = ?").Prepare()
```
astaxie: maybe we just remove the `  for mysql. as it's not necessary

WnP:
@astaxie you're right, fixed in last commit

WnP:
@astaxie I'm going to make a new pull request because we don't need 3 commits to do something simple like that

WnP:
see #1434

PR #1457, opened 2015-11-28
Log config parsing errors

ryanchapman:
Beego currently handles the case of "conf/app.conf"
not existing, but all other errors are not logged.
This fixes that.

I ran into an issue where beego was not listening on the
correct port, and it turned out that the conf/app.conf file
had a colon ":" instead of an equal sign "=" (confusion of
INI vs YAML formats).  When there was a parsing error, beego
would give up on parsing app.conf and not log anything to the
console.

### DIFF
--- a/config.go
+++ b/config.go
@@ -307,13 +307,13 @@ func init() {
 	SetLogFuncCall(true)
 
 	err = ParseConfig()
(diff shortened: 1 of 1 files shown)


astaxie:
please send the PR to develop branch, thanks

PR #1484, opened 2015-12-14
Added MaxMemory limit to CopyBody()

KilledKenny:
Beego only uses the MaxMemory flag when using go's built in functions
for parsing forms. However the CopyBody() function have no limit an will
coppy anny amount of data into memory using ioutil.ReedAll() on the
request body whitout anny size validation or limit.

This fix wrapps input.Requst.Body in a LimitedReader using the same
memory limit as ParseFormOrMulitForm()

### DIFF
--- a/context/input.go
+++ b/context/input.go
@@ -17,6 +17,7 @@ package context
 import (
 	"bytes"
 	"errors"
+	"io"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


astaxie:
Please send the PR to develop branch. Thanks

PR #1487, opened 2015-12-15
compress method fixed

JessonChan:
in http,the deflate is zlib compress method accoding to the sec
http://tools.ietf.org/html/rfc2616#section-3.5
The "zlib" format defined in RFC 1950 [31] in combination with
        the "deflate" compression mechanism described in RFC 1951 [29].


JessonChan:
@astaxie

PR #1503, opened 2015-12-21
simplfy config.go and controller.go

fuxiaohei:
config.go

- move to if-err style

controller.go

- use one bytes.Buffer in controller.RenderBytes
- clean some assignment statement code

gofmt files


astaxie:
LGTM

PR #1505, opened 2015-12-22
Retrieve session identifier from cookie and query parameters

yydzero:
Currently beego only retrieve session identifier from cookie in HTTP request.

This pull request try to retrieve session identifier from URL query parameters if could not find from cookie.

Thus following request will works:

CURL -X GET http://example.com/api/resource?gosessionid=<SESSIONID>

Session identifier is configurable.


astaxie:
please send the PR to develop branch. thanks

PR #1507, opened 2015-12-22
Retrieve session identifier from cookie and query parameters

yydzero:
Currently beego only retrieve session identifier from cookie in HTTP request.

This pull request try to retrieve session identifier from URL query parameters if could not find from cookie.

Thus following request will works:

CURL -X GET http://example.com/api/resource?gosessionid=

Session identifier is configurable.

### DIFF
--- a/session/session.go
+++ b/session/session.go
@@ -133,63 +133,65 @@ func NewManager(provideName, config string) (*Manager, error) {
 	}, nil
 }
 
-// SessionStart Start session. generate or read the session id from http request.
-// if session id exists, return SessionStore with this id.
-func (manager *Manager) SessionStart(w http.ResponseWriter, r *http.Request) (session Store, err error) {
+// getSid retrieves session identifier from HTTP Request.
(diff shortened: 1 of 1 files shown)


astaxie:
identifier from query is easy. the problem is that we need to return the sessionID to the client if the client not support cookie.

yydzero:
I hope I understand your question correctly. Here is what we are using:

During login request (like http://example.com/api/login?username=<name>&password=<password>, we will return sessionid as part of RESTful response, like:

{
  'status': 'ok',
  'resource': {
        sessionid: <sessionid>
        ....
   }
}

astaxie:
session/session.go:161: undefined: SessionStore
session/session.go:172: manager.sessionId undefined (type *Manager has no field or method sessionId)

astaxie:
I fix it https://github.com/astaxie/beego/commit/ac3a447479046559be0dc34a22b068b5169f8504

PR #1519, opened 2015-12-29
Refactoring: Move dev & prod runmodes to const

vvelikodny:
"dev" & "prod" repeated many times, let's do code cleaner

### DIFF
--- a/config/ini_test.go
+++ b/config/ini_test.go
@@ -17,6 +17,8 @@ package config
 import (
 	"os"
 	"testing"
+
+	"github.com/astaxie/beego"
 )
 
 var inicontext = `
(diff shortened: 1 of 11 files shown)


config/ini_test.go
```diff
@@ -67,7 +69,7 @@ func TestIni(t *testing.T) {
 		t.Error(pi)
 		t.Fatal(err)
 	}
-	if iniconf.String("runmode") != "dev" {
+	if iniconf.String("runmode") != beego.DEV {
```
astaxie: I think here we still van keep use the string "dev". As `iniconf.String("")` is testing the string result. it doesn't make sense use beego.DEV

astaxie: also I think use beego.DEV will dependence with the beego package. I just want to keep the config package independence.

vvelikodny: agreed

Issue #1259, opened 2015-07-14
cache模块


myml:
memory cache的put函数timeout默认值0,到期时间是下次GC,
file cache的timeout默认值0,却是永久存储,这样设计合理吗?

lei-cao:
@myml Thank you for supporting beego. please update or submit your issue in English for a better international user friendly communication.

PR #1532, opened 2016-01-06
mem cache put function fixed

JessonChan:
when a expire duration==0,it means forever
https://github.com/astaxie/beego/issues/1259

### DIFF
--- a/cache/memory.go
+++ b/cache/memory.go
@@ -34,6 +34,14 @@ type MemoryItem struct {
 	expired    int64
 }
 
+func (mi *MemoryItem) isExpire() bool {
+	// expired==0 means never
+	if mi.expired == 0 {
+		return false
+	}
+	return time.Now().Unix()-mi.Lastaccess.Unix() > mi.expired
+}
+
 // MemoryCache is Memory cache adapter.
 // it contains a RW locker for safe map storage.
 type MemoryCache struct {
@@ -55,7 +63,7 @@ func (bc *MemoryCache) Get(name string) interface{} {
 	bc.RLock()
 	defer bc.RUnlock()
 	if itm, ok := bc.items[name]; ok {
-		if (time.Now().Unix() - itm.Lastaccess.Unix()) > itm.expired {
+		if itm.isExpire() {
 			go bc.Delete(name)
 			return nil
 		}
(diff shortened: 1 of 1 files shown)


JessonChan:
@fuxiaohei 
plz review this pr.

JessonChan:
yeap,but i think it's ok.
have a look at file-mem
https://github.com/JessonChan/beego/blob/0b390912921919df1b75894a9ff06b6b2665cc75/cache/file.go#L150

fuxiaohei:
No need to follow file cache. There is no gc in file-mem.

And maybe if `item.Expired == 0`, reduce api calls in 

https://github.com/JessonChan/beego/blob/0b390912921919df1b75894a9ff06b6b2665cc75/cache/memory.go#L58

    if itm, ok := bc.items[name]; ok && itm.Expired > 0{

JessonChan:
thanks @fuxiaohei, please review again.

cache/memory.go
```diff
@@ -52,11 +60,10 @@ func NewMemoryCache() Cache {
 // Get cache from memory.
 // if non-existed or expired, return nil.
 func (bc *MemoryCache) Get(name string) interface{} {
-	bc.lock.RLock()
-	defer bc.lock.RUnlock()
+	bc.RLock()
+	defer bc.RUnlock()
 	if itm, ok := bc.items[name]; ok {
-		if (time.Now().Unix() - itm.Lastaccess.Unix()) > itm.expired {
-			go bc.Delete(name)
```
JessonChan: as commit log says.
the memory gc goroutine will do the delele,no more goroutine here.

JessonChan:
@fuxiaohei typo fixed

cache/memory.go
fuxiaohei: LGTM

fuxiaohei:
https://github.com/JessonChan/beego/blob/develop/cache/memory.go#L65

it can:

```go
if itm, ok := bc.items[name]; ok && !itm.isExpired() {
    return itm.val
```

JessonChan:
@fuxiaohei 
i think the style  is not so readable.
hmmm,i prefer my version .

fuxiaohei:
LGTM

maybe I prefer compact style

PR #1536, opened 2016-01-07
change get sessionID logic from cookie

ysqi:
1. Destorey session cookie need check config : EnableSetCookie
2. Need get new sessionid if session cookie is nil or sessionID is emtpy or cookie is expired. 
3. Add test


astaxie:
send the PR to develop branch

PR #1559, opened 2016-01-12
Create Dockerfile

shoother:
Create a Dockerfile for beego


astaxie:
I think the Dockerfile should be generated in bee problem. it should not included in beego

pjoe:
Also if you just want to run your app directly you can make a much smaller image by statically linking and using ```FROM scratch```

shoother:
This is a Docker deploy solution for beego, if don't include the main code branch, in the beego docs repos maybe better.

PR #1565, opened 2016-01-15
logs:remove select from startLogger

fugr:
conflict fixed


JessonChan:
@fugr 
for-range is short and compact ,by i think for-select loop is much more popular indeed.

astaxie:
it's not run once. it should always running there

Issue #1580, opened 2016-01-18
Receiving Error when wildcard (*.*) is empty


hbejgel:
I am using the annotation router and it looks like this:

<pre><code>// @router /:account/pricing/prices/:id/:brand/*.* [get]
</code></pre>

When I get to the router with a blank wildcard such as with: 
<pre><code>localhost:8080/mastercatalog/pricing/prices/340502/2000170/
</code></pre>

I get the following error:
<pre><code>2016/01/18 20:54:03 [router.go:854][C] Handler crashed with error runtime error: slice bounds out of range
2016/01/18 20:54:03 [router.go:860][C] /usr/local/go/src/runtime/asm_amd64.s:437
2016/01/18 20:54:03 [router.go:860][C] /usr/local/go/src/runtime/panic.go:423
2016/01/18 20:54:03 [router.go:860][C] /usr/local/go/src/runtime/panic.go:18
2016/01/18 20:54:03 [router.go:860][C] /Users/hbejgel/Documents/Projects/go/src/github.com/astaxie/beego/tree.go:423
</code></pre>

This happens because in [tree.go](https://github.com/astaxie/beego/blob/master/tree.go) the index variable equals to 3 while len(wildcardValues) - 1 equals to 2 so you are actually indexing wildcardValues[3:2] which causes this error. Is this the intended behaviour?

astaxie:
@hbejgel I think `*.*` should be never empty. I think `*` means catchall. while `*.*` should be never empty. What do you think?

hbejgel:
Well, I tried using `*` but then in the `Params` the `:path` does not show up. I needed the whole thing to be an optional path. So I changed it to `*.*` and it worked fine, except when I leave the path blank I get this error. In my opinion, empty should match both `*` and `*.*' or at least the `:path` should come in the `Params` when using `*`. What were exactly your intentions and the differences you intended between `*` and `*.*`? If it was for it to be a file path requiring an extension then the regular expression `"([^.]+).(.+)"` found in tree.go is wrong, it should have been `"([^.]+)\.(.+)"`

PR #1581, opened 2016-01-18
Checks if index is greater than the length of the wildcards. #1580

hbejgel:


JessonChan:
@hbejgel 
the var index is always less than len(wildcardValues),so there's no need to check.

hbejgel:
Well, when you pass now values to the wildcard as shown on issue #1580
that's not what happens thus causing a panic.

Issue #1597, opened 2016-01-21
新建的 hello 项目，只在 main.go 加了一个 Notice 日志，beego.Notice("Networthicense is Correct!") 似乎没有输出任何内容


linux-mac:
新建的 hello 项目，只在 main.go 加了一个 Notice 日志
```
beego.Notice("Networthicense is Correct!") 
```
似乎没有输出任何内容
```
[root@git hello]# bee version
bee   :1.4.1
beego :1.6.0
Go    :go version go1.5.3 linux/amd64
```

好像 beego 1.5.0 是会输出的，不知道我错过了什么%...

![snip20160121_1](https://cloud.githubusercontent.com/assets/6309187/12480186/7ad41816-c07b-11e5-8e66-fa95f54d4c5a.png)

astaxie:
yea, init the log in the beego.Run

linux-mac:
sorry, I don't know what shoud i do?

ysqi:
hi linux-mac, You need exec bee.Run before use everythings about beego.
But you can use other way:
1. use fmt.Println('')
2. create  yourself logger , see :http://beego.me/docs/module/logs.md
```

astaxie:
@linux-mac it looks like beego's problem. we will fix in 1.6.1

linux-mac:
I am looking forward to the release of the 1.6.1 version.

youngsterxyf:
@astaxie Can I try this ?

astaxie:
@youngsterxyf yea, Go for it

linux-mac:
@youngsterxyf yea! You fixed it.

linux-mac:
![www1](https://cloud.githubusercontent.com/assets/6309187/12528591/2d98d6ac-c1d6-11e5-941d-b05d957c1ccd.jpg)

PR #1610, opened 2016-01-23
Fix issue1597

youngsterxyf:
R.T. #1597

### DIFF
--- a/beego.go
+++ b/beego.go
@@ -73,6 +73,8 @@ func initBeforeHTTPRun() {
 		panic(err)
 	}
 
+	SetLogFuncCall(BConfig.Log.FileLineNum)
+
 	//init hooks
 	AddAPPStartHook(registerMime)
 	AddAPPStartHook(registerDefaultErrorHandler)
--- a/beego.go
+++ b/beego.go
@@ -15,7 +15,6 @@
 package beego
 
 import (
-	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


astaxie:
@youngsterxyf build failed

youngsterxyf:
@astaxie maybe this failure is not related with my pr，but i will continue to check the reason。

image

youngsterxyf:
@astaxie I made no change, but it succeed. :smile:

astaxie:
@youngsterxyf weird. I will review later.

youngsterxyf:
@astaxie It's a bug of go or travisCI ?

astaxie:
it looks like the timeout and autoclose the connection. then call the WaitGroup.Done(). I see the same issue in beego's grace module. I still have no idea on that. but we can recover from the panic

beego.go
```diff
@@ -73,13 +72,6 @@ func initBeforeHTTPRun() {
 	if err != nil {
 		panic(err)
 	}
-	//init log
-	for adaptor, config := range BConfig.Log.Outputs {
-		err = BeeLogger.SetLogger(adaptor, config)
-		if err != nil {
-			fmt.Printf("%s with the config `%s` got err:%s\n", adaptor, config, err)
-		}
-	}
```
astaxie: why I add these init log here. As user can change the config path. So if they change that, we need to init the BeeLogger again.

PR #1611, opened 2016-01-23
fix issue #1601

youngsterxyf:
R.T. #1601

### DIFF
--- a/beego.go
+++ b/beego.go
@@ -69,10 +69,13 @@ func Run(params ...string) {
 
 func initBeforeHTTPRun() {
 	// if AppConfigPath is setted or conf/app.conf exist
+	/*
 	err := ParseConfig()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


beego.go
```diff
@@ -69,10 +69,13 @@ func Run(params ...string) {
 
 func initBeforeHTTPRun() {
 	// if AppConfigPath is setted or conf/app.conf exist
+	/*
 	err := ParseConfig()
 	if err != nil {
 		panic(err)
 	}
+	*/
```
astaxie: you can't do this as maybe user change the `AppConfigPath` in main function. So we need to run parse the config.

youngsterxyf: @astaxie Sorry, i will try to give a better solution.

youngsterxyf:
I will come back！

Issue #1613, opened 2016-01-23
upgrate to beego 1.6, the path error


coseyo:
When I upgrate to beego 1.6, the beego app just read the file under the current directory where you execute it. Including static file, conf/app.conf and route file and commentsRouter_*.files.

It is ok in beego 1.5.

PR #1616, opened 2016-01-23
fix path issue #1613

coseyo:


coseyo:
I see it, I will check it.

config.go
```diff
@@ -179,15 +181,23 @@ func init() {
 // ParseConfig parsed default config file.
 // now only support ini, next will support json.
 func ParseConfig() (err error) {
+	AppPath, _ := filepath.Abs(filepath.Dir(os.Args[0]))
+	workPath, _ := os.Getwd()
+	workPath, _ = filepath.Abs(workPath)
```
astaxie: I think these init could move to config.init(){}

PR #1625, opened 2016-01-25
Fix utils mail some bugs

miraclesu:
1. Fix utils mail from field can't including Chinese bug
2. Fix mail Chinese subject garbled bug

#1338

### DIFF
--- a/utils/mail.go
+++ b/utils/mail.go
@@ -235,11 +235,11 @@ func (e *Email) Send() error {
 	}
 
 	if len(e.From) == 0 {
-		e.From = from.String()
+		e.From = e.Username
 	}
-	// use mail's RFC 5322 to encode any string
-	sub := mail.Address{e.Subject, ""}
-	e.Subject = strings.Trim(sub.String(), " <@>")
+	// use mail's RFC 2047 to encode any string
+	sub := mail.Address{Name: e.Subject, Address: ""}
+	e.Subject = strings.TrimRight(sub.String(), " <@>")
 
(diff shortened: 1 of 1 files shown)


utils/mail.go
```diff
@@ -228,14 +225,22 @@ func (e *Email) Send() error {
...
 		return err
 	}
-	e.From = from.String()
+
+	if len(e.From) == 0 {
+		e.From = e.Username
+	}
+	// use mail's RFC 2047 to encode any string
+	sub := mail.Address{Name: e.Subject, Address: ""}
+	e.Subject = strings.TrimRight(sub.String(), " <@>")
```
astaxie: I think this solution is trick. I still like the `mime.QEncoding`. could you do some research on go1.3, is there similar function.

miraclesu: I couldn't found a similar function likes `mime.QEncoding` in go1.3 and go1.4.
Another solution is move the `mail.Address.String` code into uitls package.
[go1.3 encoding](https://github.com/golang/go/blob/release-branch.go1.3/src/pkg/net/mail/message.go#L154), [go1.4 encoding](https://github.com/golang/go/blob/release-branch.go1.4/src/net/mail/message.go#L155)

astaxie: maybe we can add mime.QEncoding into mail.go and just rename to private function.

miraclesu:
@astaxie I have moved `mime.QEncoding.Encode` logic to mail.go, and it works good.

astaxie:
LGTM

PR #1629, opened 2016-01-26
add Config.Load func

coseyo:
Add new feature for beego. 

you can load specific config file by run mode, just as below

cf, err := beego.BConfig.Load("auth.json")

It will load conf/dev/auth.json if you run in dev mode, or load conf/prod/auth.json in prod mode.

If can not find the file in above dir, it will load conf/auth.json.

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -88,7 +88,7 @@ func listConf(rw http.ResponseWriter, r *http.Request) {
 
 	data := make(map[interface{}]interface{})
 	switch command {
-	case "conf":
+	case confDir:
 		m := make(map[string]interface{})
 		m["AppConfigPath"] = AppConfigPath
 		m["AppConfigProvider"] = AppConfigProvider
--- a/beego.go
+++ b/beego.go
@@ -101,7 +101,7 @@ func initBeforeHTTPRun() {
 // TestBeegoInit is for test package init
 func TestBeegoInit(ap string) {
 	os.Setenv("BEEGO_RUNMODE", "test")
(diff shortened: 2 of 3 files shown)


astaxie:
@coseyo I don't think Load is a correct way to load config for different RunMode. I would to say we need to use include in ini config.

```
[dev]
include "dev.conf"
[prod]
include "prod.conf"
```

coseyo:
@astaxie This method can load different type of adpater file. On the other hand, if the site has many configuration files, it is much clear to put the files in specific path, no need to change the ini config.
well, just in my opinion.

astaxie:
@coseyo I think https://github.com/astaxie/beego/blob/develop/config.go#L315 is what you needed

coseyo:
Maybe there is something different, multi invoking LoadAppConfig func would exec parseConfig multi times, and it would init BConfig multi time. Sometimes, not all config file is used for the app common config.

astaxie:
BConfig only init once. if the config didn't include any variable about BConfig. it still use the old value.

coseyo:
Oh, that's it, My mistake~
For keeping the config file independent, I thnk I would still use the load method for the reason I first mention. support multi adapter and keep the code module's config independent and looks clear.

In fact that I have make a package to do this staff.

https://github.com/coseyo/beeconfig

By the way, is there a github repo to store the third party package for beego ?

astaxie:
@coseyo https://github.com/beego/contrib

coseyo:
good~

PR #1632, opened 2016-01-27
Config logic

youngsterxyf:
try to fix issue #1597 issue #1601 

Make a little adjustment to the logic of config module：

- Add `beego.LoadAppConfig`：it will parse config file immediately when called
- No more to automatically parse config file in `initBeforeHTTPRun`
- `beego.LoadAppConfig` will call `parseConfig`, the last part of `parseConfig` will try to init log writers again

### DIFF
--- a/config.go
+++ b/config.go
@@ -124,6 +124,10 @@ func init() {
 	workPath, _ = os.Getwd()
 	workPath, _ = filepath.Abs(workPath)
 
+	if workPath != AppPath {
+		os.Chdir(AppPath)
+	}
+
 	BConfig = &Config{
 		AppName:             "beego",
 		RunMode:             DEV,
@@ -195,10 +199,6 @@ func init() {
 
 // now only support ini, next will support json.
 func parseConfig(appConfigPath string) (err error) {
-	if workPath != AppPath {
-		os.Chdir(AppPath)
-	}
-
 	AppConfig, err = newAppConfig(AppConfigProvider, appConfigPath)
 	if err != nil {
 		return err
--- a/config.go
+++ b/config.go
@@ -312,7 +312,7 @@ func parseConfig(appConfigPath string) (err error) {
 }
 
 // LoadAppConfig allow developer to apply a config file
-func LoadAppConfig(configPath string, adapterName string) error {
+func LoadAppConfig(adapterName string, configPath string) error {
 	absConfigPath, err := filepath.Abs(configPath)
 	if err != nil {
 		return err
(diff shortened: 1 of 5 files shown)


youngsterxyf:
@astaxie @JessonChan

staticfile_test.go
```diff
@@ -8,9 +8,11 @@ import (
 	"io/ioutil"
 	"os"
 	"testing"
+	"path/filepath"
 )
 
-const licenseFile = "./LICENSE"
+var currentWorkDir, _ = os.Getwd()
+var licenseFile = filepath.Join(currentWorkDir, "LICENSE")
```
astaxie: why not just use workPath

config.go
```diff
@@ -286,9 +286,52 @@ func ParseConfig() (err error) {
...
+			fmt.Printf("%s with the config `%s` got err:%s\n", adaptor, config, err)
+		}
+	}
+	SetLogFuncCall(BConfig.Log.FileLineNum)
+
 	return nil
 }
 
+// LoadAppConfig allow developer to apply a config file
+func LoadAppConfig(configPath string, adapterName string) error {
```
astaxie: Could you change the sequence adapterName and configPath?

config.go
```diff
@@ -181,26 +186,19 @@ func init() {
...
-// now only support ini, next will support json.
-func ParseConfig() (err error) {
-	if AppConfigPath == "" {
-		// initialize default configurations
-		AppConfigPath = filepath.Join(AppPath, "conf", "app.conf")
-		if !utils.FileExists(AppConfigPath) {
-			AppConfig = &beegoAppConfig{config.NewFakeConfig()}
-			return
-		}
+	AppConfigPath = filepath.Join(AppPath, "conf", "app.conf")
```
astaxie: if change to this style, why we still need AppConfigPath as public ?

staticfile_test.go
youngsterxyf: using workPath, the test case will fail

config.go
youngsterxyf: Yes, not need, but the next release will be a minor version，so i think not changing the Public variable may be better

config.go
youngsterxyf: OK.

config.go
astaxie: it doesn't make sense. As if this variable still as public. but beego will never use it. As we overwrite it here. And I think we need to just build failed. and let the developer call LoadAppConfig

youngsterxyf: OK.

PR #1638, opened 2016-01-27
Add noColor option for console logger

pjoe:
- Also added simple test


astaxie:
@pjoe I think use `Color` will make more sense for the config.

pjoe:
Changed it to Color. That was actually my first instinct, but then thought NoColor would better handle defaulting to false - easily handled in constructor though :)

PR #1639, opened 2016-01-28
try to fix the little bug when calling Close or Flush in async mode

youngsterxyf:
In the Async mode, calling the `func (bl *BeeLogger) Flush() ` or `func (bl *BeeLogger) Close()`, sometimes some log message may be discard, or the order of log may be wrong. Please see the example:

```go
package main

import (
	"github.com/astaxie/beego/logs"
)

var log *logs.BeeLogger

func init() {
	log = logs.NewLogger(10000)
	log.SetLogger("console", `{"level": 10}`)
	log.SetLevel(logs.LevelDebug)
	log.EnableFuncCallDepth(true)
	log.SetLogFuncCallDepth(3)
	log.Async()
}

func main() {
	log.Debug("1 log message")
	log.Debug("2 log message")
	log.Debug("3 log message")
	log.Debug("4 log message")
	log.Debug("5 log message")

	log.Close()
}
```

the output:

image

you can see that first three times  the second log has been discard

### DIFF
--- a/logs/log.go
+++ b/logs/log.go
@@ -98,8 +98,8 @@ type BeeLogger struct {
 	loggerFuncCallDepth int
 	asynchronous        bool
 	msgChan             chan *logMsg
-	signalChan			chan string
-	wg					sync.WaitGroup
+	signalChan          chan string
+	wg                  sync.WaitGroup
 	outputs             []*nameLogger
 }
 
@@ -111,7 +111,7 @@ type nameLogger struct {
 type logMsg struct {
 	level int
 	msg   string
-	when time.Time
+	when  time.Time
 }
 
 var logMsgPool *sync.Pool
@@ -125,7 +125,6 @@ func NewLogger(channelLen int64) *BeeLogger {
 	bl.loggerFuncCallDepth = 2
 	bl.msgChan = make(chan *logMsg, channelLen)
 	bl.signalChan = make(chan string, 1)
-	bl.wg.Add(1)
 	return bl
 }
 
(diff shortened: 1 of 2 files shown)


logs/log.go
```diff
@@ -98,6 +98,8 @@ type BeeLogger struct {
 	loggerFuncCallDepth int
 	asynchronous        bool
 	msgChan             chan *logMsg
+	signalChan			chan string
+	wg					sync.WaitGroup
```
astaxie: fmt

youngsterxyf: @astaxie sorry

astaxie: use the atom or liteide. auto fmt when save

youngsterxyf: @astaxie OK

logs/log.go
```diff
@@ -345,27 +370,16 @@ func (bl *BeeLogger) Trace(format string, v ...interface{}) {
...
-		break
-	}
-	for _, l := range bl.outputs {
-		l.Flush()
-		l.Destroy()
-	}
-	bl.outputs = nil
+	bl.signalChan <- "close"
+	bl.wg.Wait()
+	bl.wg.Add(1)
```
astaxie: if we close the Log, do we still need to add?

youngsterxyf:
@astaxie with golang tip version, `go vet ./...` will raise an error.

image

astaxie:
@youngsterxyf it should be fixed in develop now. I like travis

Issue #1617, opened 2016-01-24
Context.Abort acts weird.


boombuler:
Hi,

when I call Context.Abort, the status code is ignored, then it panics with the body, which may contain a status code as string, which then will be converted to the corresponding error page, which is send to the client with status code 200. 
I was expecting the same behavior as Controller.CustomAbort

osleg:
+1

youngsterxyf:
@astaxie When the runmode is not dev,  After handling Context.Abort, the http response status code is 200, maybe this is wrong, i think it should be 500

boombuler:
@youngsterxyf it should return the statuscode passed as argument...
if i call:

```go
Ctx.Abort(http.StatusBadRequest, "")
```

I would expect the client to recieve a message with 400... Not 500 and not 200... 
And having a different behavior in dev mode don't make it any better...

youngsterxyf:
@boombuler Yes, I have fix it.

![image](https://cloud.githubusercontent.com/assets/834231/12678009/85871436-c6d7-11e5-9761-519bbf966d86.png)

boombuler:
@youngsterxyf I've seen it, thats why I deleted my last comment. Thanks for this ;)

xionglun:
Does it still not fixed yet?

astaxie:
fixed in beego1.7

PR #1640, opened 2016-01-29
fix issue#1617：make Context.Abort more like Controller.CustomAbort, not ignore status parameter

youngsterxyf:
R.T. Try to fix issue #1617

### DIFF
--- a/router.go
+++ b/router.go
@@ -862,6 +862,8 @@ func (p *ControllerRegister) recoverPanic(context *beecontext.Context) {
 			}
 			if BConfig.RunMode == DEV {
 				showErr(err, context, stack)
+			} else {
+				context.ResponseWriter.WriteHeader(500)
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


router.go
```diff
@@ -849,19 +849,24 @@ func (p *ControllerRegister) recoverPanic(context *beecontext.Context) {
...
+				for i := 1; ; i++ {
+					_, file, line, ok := runtime.Caller(i)
+					if !ok {
+						break
+					}
+					Critical(fmt.Sprintf("%s:%d", file, line))
+					stack = stack + fmt.Sprintln(fmt.Sprintf("%s:%d", file, line))
+				}
 				showErr(err, context, stack)
+				return
```
astaxie: in the dev mode, panic already output ctx.ResponseWriter.Write([]byte(body)). and when call `showErr(err, context, stack)` will that broken?

youngsterxyf: Sometimes that will occure a problem. I have fix that, thanks!

astaxie:
@youngsterxyf I don't think this fix is works well. if Abort(404,"404"), it cause mulit header

PR #1652, opened 2016-02-02
colorful console and go fmt

JessonChan:
go fmt  makes the code more **go**

### DIFF
--- a/config.go
+++ b/config.go
@@ -15,11 +15,11 @@
 package beego
 
 import (
+	"fmt"
 	"html/template"
 	"os"
 	"path/filepath"
 	"strings"
-	"fmt"
 
 	"github.com/astaxie/beego/config"
 	"github.com/astaxie/beego/session"
@@ -111,7 +111,6 @@ var (
 	// GlobalSessions is the instance for the session manager
 	GlobalSessions *session.Manager
 
-	workPath string
 	// appConfigPath is the path to the config files
 	appConfigPath string
 	// appConfigProvider is the provider for the config, default is ini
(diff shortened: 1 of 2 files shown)


logs/console.go
```diff
@@ -68,7 +68,11 @@ func (c *consoleWriter) Init(jsonConfig string) error {
 	if len(jsonConfig) == 0 {
 		return nil
 	}
-	return json.Unmarshal([]byte(jsonConfig), c)
+	err := json.Unmarshal([]byte(jsonConfig), c)
+	if runtime.GOOS == "windows" {
```
JessonChan: if the os is windows,the color filed will be useless

config.go
```diff
@@ -120,12 +119,8 @@ var (
 
 func init() {
 	AppPath, _ = filepath.Abs(filepath.Dir(os.Args[0]))
-	workPath, _ = os.Getwd()
-	workPath, _ = filepath.Abs(workPath)
 
-	if workPath != AppPath {
-		os.Chdir(AppPath)
-	}
+	os.Chdir(AppPath)
```
astaxie: current workdir is different with apppath

JessonChan: but os.Chdir(AppPath) will change current work dir to AppPath

config.go
```diff
@@ -120,12 +119,8 @@ var (
 
 func init() {
 	AppPath, _ = filepath.Abs(filepath.Dir(os.Args[0]))
-	workPath, _ = os.Getwd()
```
JessonChan: @youngsterxyf have a look here,thanks

youngsterxyf: I think the change at this position is OK. However, after starting, should beego app change current dir to AppPath? We should make a explicit hypothesis - relative path settings is relative to AppPath or current path - this decide that beego should change current dir to AppPath or not.

JessonChan: should change to AppPath for some configure files,and workPath is useless in fact.

PR #1655, opened 2016-02-02
A crash of GetFiles fixed

yesidi:
Call this function would crash when upload`s stream was nil. now fixed!

### DIFF
--- a/controller.go
+++ b/controller.go
@@ -500,8 +500,10 @@ func (c *Controller) GetFile(key string) (multipart.File, *multipart.FileHeader,
 //	}
 // }
 func (c *Controller) GetFiles(key string) ([]*multipart.FileHeader, error) {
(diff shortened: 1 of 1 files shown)


controller.go
```diff
@@ -500,8 +500,10 @@ func (c *Controller) GetFile(key string) (multipart.File, *multipart.FileHeader,
 //	}
 // }
 func (c *Controller) GetFiles(key string) ([]*multipart.FileHeader, error) {
-	if files, ok := c.Ctx.Request.MultipartForm.File[key]; ok {
-		return files, nil
+	if c.Ctx.Request.MultipartForm != nil && c.Ctx.Request.MultipartForm.File != nil {
```
youngsterxyf: Maybe just checking `c.Ctx.Request.MultipartForm != nil` is enough

astaxie:

PR #1670, opened 2016-02-09
Set it to 0 afterwards to prevent "multiple response.WriteHeader calls"

wpajqz:
I rewrite error404 in my ErrorController,in ErrorControll I call ServerJSON(), Then It's responsed "http: multiple response.WriteHeader calls"

image


astaxie:
the same as #1719

PR #1683, opened 2016-02-13
fixed #1664 check session config before use session

### LINKED ISSUE #1664: SetSession can cause null pointer exception
If parameter SessionOn=false the call to SetSession will cause a null pointer exception. Perhaps this can be handled in a more friendly way like beego.trace without causing an exception.

ysqi:
need to set SessionOn=true before use session .  this info need panic
and set session info from  context before init controller.

Note:  I'm not sure whether it is reasonable error message.

#1664

### DIFF
--- a/controller.go
+++ b/controller.go
@@ -115,6 +115,7 @@ func (c *Controller) Init(ctx *context.Context, controllerName, actionName strin
 	c.EnableXSRF = true
 	c.Data = ctx.Input.Data()
 	c.methodMapping = make(map[string]func())
+	c.CruSession = ctx.Input.CruSession
 }
 
 // Prepare runs after Init before request function execution.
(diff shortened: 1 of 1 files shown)


controller.go
```diff
@@ -115,6 +115,7 @@ func (c *Controller) Init(ctx *context.Context, controllerName, actionName strin
 	c.EnableXSRF = true
 	c.Data = ctx.Input.Data()
 	c.methodMapping = make(map[string]func())
+	c.CruSession = ctx.Input.CruSession
```
fuxiaohei: If assign c.CurSession here, no need assign again in Controller.StartSession(), no matter BConfig.WebConfig.Session.SessionOn

ysqi: Outside package can  directly acces  `CruSession‘   ,so set value in init function. and  ctx.Input.CruSession is `nil` if SessionOn = false .

I argree "no need assign again in Controller.StartSession() " , but we need keep this function for old system.

astaxie:
I don't think we need to panic the error if it's not start the session.

PR #1688, opened 2016-02-14
fixed handle config issue

ysqi:
First parse config  error need handle. so need panic error in init application. <del>Then don't end application by include other config handle failure .<del>

current app will stop if include config file not exist , e.g:
```ini
appname = demo
httpport = 8080
runmode = dev

# 不存在的配置文件
include  APPTEST.conf
```

### DIFF
--- a/config/ini.go
+++ b/config/ini.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"log"
 	"os"
 	"path"
 	"strconv"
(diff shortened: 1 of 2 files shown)


config/ini.go
```diff
@@ -134,7 +135,9 @@ func (ini *IniConfig) parseFile(name string) (*IniConfigContainer, error) {
 				}
 				i, err := ini.parseFile(otherfile)
 				if err != nil {
-					return nil, err
+					// ignore error
+					log.Printf("[warn] handle config %q error, %s \n", key, err.Error())
```
astaxie: why ignore error?

ysqi: I don't remember that reason.  :)

Issue #1598, opened 2016-01-21
c.Abort http: multiple response.WriteHeader calls


jinze1994:
I saw the v 1.6.0  fix the bug: http: multiple response.WriteHeader calls #1329.
But I still encounter this when call c.Abount("404")

astaxie:
@jinze1994 where did you call this function?

PR #1698, opened 2016-02-19
fix issue #1598

migege:
If there was an ```ErrorController``` set to beego, calling ```Controller.Abort()``` would cause ```http: multiple response.WriteHeader calls``` since that another ```WriteHeader``` will be called before ```ControllerRegister.ServeHTTP()``` function return.

The 1st ```WriteHeader``` in ```Controller.CustomAbort()``` is redundant.


astaxie:
dup with #1719

PR #1699, opened 2016-02-19
New logs adaptor which supports writing to multiple files

quxiao:
Add an new beego logs adaptor, which supports writing different files according to different log level in configuration.

Configuration example:

        {  
            "levelfiles": [{
                "levelnames": ["debug", "trace"],
                "filename": "some.log.debug"
            },{
                "levelnames": ["info"],
                "filename": "some.log"
            },{
                "levelnames": ["warning", "critical"],
                "filename": "some.log.wf"
            }]
        }

### DIFF
--- a/logs/multi_file.go
+++ b/logs/multi_file.go
@@ -1,0 +1,129 @@
+package logs
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+type multiFileLogWriter struct {
+	Maxlines int `json:"maxlines"`
+	// Rotate at size
+	Maxsize int `json:"maxsize"`
+	// Rotate daily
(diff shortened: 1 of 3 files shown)


astaxie:
@JessonChan what's your opinion?

JessonChan:
this pr is almost like #1721 but i think it's a little verbose and has some same code compared to `file.go`,such as multiFileLogWriter struct and fileLogWriter.
and this pr is not complete, i can not find  where is the definition of the function "logLevelName2Int".

quxiao:
@JessonChan sorry, I forgot to commit one new file, and the function logLevelName2Int is in it.

astaxie:
I think dup with #1721

PR #1692, opened 2016-02-15
Urgent : Correction for cookie in flash usage

mlgd:
Please ignore [Pull 1644](https://github.com/astaxie/beego/pull/1644). To fix cookie max-age problem for flash message, use this pull request for more sustainability.

**Explaination** :
Old function Cookie() in context/output.go don't use Max-Age when param set to 0 :
```
        switch v := others[0].(type) {
            case int:
            if v > 0 {
                fmt.Fprintf(&b, "; Expires=%s; Max-Age=%d", time.Now().Add(time.Duration(v) * time.Second).UTC().Format(time.RFC1123), v)
            } else if v < 0 {
                fmt.Fprintf(&b, "; Max-Age=0")
            }
            case int64:
            if v > 0 {
                fmt.Fprintf(&b, "; Expires=%s; Max-Age=%d", time.Now().Add(time.Duration(v) * time.Second).UTC().Format(time.RFC1123), v)
            } else if v < 0 {
                fmt.Fprintf(&b, "; Max-Age=0")
            }
            case int32:
            if v > 0 {
                fmt.Fprintf(&b, "; Expires=%s; Max-Age=%d", time.Now().Add(time.Duration(v) * time.Second).UTC().Format(time.RFC1123), v)
            } else if v < 0 {
                fmt.Fprintf(&b, "; Max-Age=0")
            }
        }
```

And old function Store() in flash.go :
`c.Ctx.SetCookie(FlashName, url.QueryEscape(flashValue), 0, "/")`

PR #1703, opened 2016-02-20
fix the cookie's max-age problem

sheng:
Ummm...the struct `Cookie` in `net/http` wrote:

```go
type Cookie struct {
        Name  string
        Value string

        Path       string    // optional
        Domain     string    // optional
        Expires    time.Time // optional
        RawExpires string    // for reading cookies only

        // MaxAge=0 means no 'Max-Age' attribute specified.
        // MaxAge<0 means delete cookie now, equivalently 'Max-Age: 0'
        // MaxAge>0 means Max-Age attribute present and given in seconds
        MaxAge   int
        Secure   bool
        HttpOnly bool
        Raw      string
        Unparsed []string // Raw text of unparsed attribute-value pairs
}
```

And the method `Cookie(name string, value string, others ...interface{})` in `context/output.go` wrote:

```go
...
        if maxAge > 0 {
            fmt.Fprintf(&b, "; Expires=%s; Max-Age=%d", time.Now().Add(time.Duration(maxAge)*time.Second).UTC().Format(time.RFC1123), maxAge)
        } else {
            fmt.Fprintf(&b, "; Max-Age=0")
        }
...
```

Then the method `Store(c *Controller)` in `flash.go` wrote:

```go
...
        c.Ctx.SetCookie(BConfig.WebConfig.FlashName, url.QueryEscape(flashValue), 0, "/")
...
```

and function `ReadFromRequest(c *Controller)` wrote:
```go
...
        //read one time then delete it
        c.Ctx.SetCookie(BConfig.WebConfig.FlashName, "", -1, "/")
...
```

I found **three things** when I read the above code.

**Thing 1:**
The param `-1` in method call `c.Ctx.SetCookie(BConfig.WebConfig.FlashName, "", -1, "/")` is meaningless because the `if...else...` in method `Cookie(name string, value string, others ...interface{})`.

**Thing 2:**
The Flash feature does not work properly.

**Thing 3:**
I can't use the method `SetCookie(name string, value string, others ...interface{})` in `context/context.go` properly when I trying to set the Max-Age of a Cookie to `0` or `-1`. **They got the same results.**

Not only my own code but also the method `Store(c *Controller)` and function `ReadFromRequest(c *Controller)` in `context/flash.go` have these problems.

The reason why the Flash feature does not work properly is:
```go
...
        // MaxAge=0 means no 'Max-Age' attribute specified.
        // MaxAge<0 means delete cookie now, equivalently 'Max-Age: 0'
...
```

`...delete cookie now, equivalently 'Max-Age: 0'...`

So I'm here to fix it. :-)

p.s. 英文不太好，若有语法错误还请谅解。😄

### DIFF
--- a/context/output.go
+++ b/context/output.go
@@ -97,8 +97,11 @@ func (output *BeegoOutput) Cookie(name string, value string, others ...interface
 			maxAge = v
 		}
 
-		if maxAge != 0 {
+		switch {
+		case maxAge > 0:
 			fmt.Fprintf(&b, "; Expires=%s; Max-Age=%d", time.Now().Add(time.Duration(maxAge)*time.Second).UTC().Format(time.RFC1123), maxAge)
+		case maxAge < 0:
+			fmt.Fprintf(&b, "; Max-Age=0")
 		}
 	}
 
--- a/context/output.go
+++ b/context/output.go
@@ -97,10 +97,9 @@ func (output *BeegoOutput) Cookie(name string, value string, others ...interface
 			maxAge = v
 		}
 
-		switch {
-		case maxAge > 0:
+		if maxAge > 0 {
 			fmt.Fprintf(&b, "; Expires=%s; Max-Age=%d", time.Now().Add(time.Duration(maxAge)*time.Second).UTC().Format(time.RFC1123), maxAge)
-		case maxAge < 0:
+		} else if maxAge < 0 {
 			fmt.Fprintf(&b, "; Max-Age=0")
 		}
 	}
--- a/context/output.go
+++ b/context/output.go
@@ -97,10 +97,8 @@ func (output *BeegoOutput) Cookie(name string, value string, others ...interface
 			maxAge = v
 		}
 
-		if maxAge > 0 {
+		if maxAge != 0 {
 			fmt.Fprintf(&b, "; Expires=%s; Max-Age=%d", time.Now().Add(time.Duration(maxAge)*time.Second).UTC().Format(time.RFC1123), maxAge)
-		} else {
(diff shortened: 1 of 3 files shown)


mlgd:
Your fix for cookies don't work with Flash.
Please see pull #1692

sheng:
Hi @mlgd can you tell me what browser you are using? In fact, this fix worked for me.

**My environment:**
OS X El Capitan 10.11.3
Safari 9.0.3
Go 1.6
Beego  1.6.0
Bee 1.4.1

mlgd:
You probably fix a problem with cookie but not the Flash Message problem that use cookie, because you send a expires header in past for negative max-age value (it equals to delete cookie).
But I need a correction on my patch syntactically "net / http".

**My environment:**
OS X El Capitan 10.11.3 or Windows 7
Safari 9.0.3 or Chrome or Firefox
Go 1.6
Beego 1.6.0
Bee 1.4.1

hippora:
does this fixed now?

hippora:
does this fixed now?

`c.Ctx.SetCookie("key", value, 0, "/")` didn't work correctly.

but i comment the code in `github.com/astaxie/beego/context/output.go`
```
} else {
			fmt.Fprintf(&b, "; Max-Age=0")
		}
```
and then work for me ,i don't know the reason,please check.

thanks.

sheng:
Hi @hippora,
It's fixed now, but @astaxie has not yet merge it. So please wait for some time, until he had finished the second session of the Gopher conference.😄

By the way, your "comment" is the same way like my fix.

Issue #1693, opened 2016-02-15
Collision between EnableGZip and Controller.Abort()


mlgd:
Usage of Controller.Abort() don't worked correcly when GZip is enabled and customs errors controllers are defined. 
Console send :
http: multiple response.WriteHeader calls

Sample controller :
```
func (this *MyController) Get() {
	this.Abort("401")
}
```

The only way, disable GZip in custom ErrorController :
```
func (this *ErrorController) Prepare() {
	this.Ctx.Output.EnableGzip = false
}
```

JessonChan:
i can't repeat this bug on my bee new project,can you paste your bee version output and have a test a mini project?

mlgd:
Download the project here : [http://dl.free.fr/jpiNLaBZx](http://dl.free.fr/jpiNLaBZx)
Run main.go
Access to http://localhost:8080/

Uncomment line 10 in controllers/errors.go
Run and access again

JessonChan:
@mlgd  thanks for your demo ,i have find the bug and will fixed it soon.The bug is hidden in this line https://github.com/astaxie/beego/blob/master/controller.go#L289

PR #1719, opened 2016-02-24
multiple response.WriteHeader calls

JessonChan:
#1693 
this happens when content-type is empty
https://golang.org/src/net/http/server.go#L1031

the net/http package should set header first,then set http status code,at last write the content


context/context.go
```diff
@@ -59,7 +59,10 @@ type Context struct {
 // Reset init Context, BeegoInput and BeegoOutput
 func (ctx *Context) Reset(rw http.ResponseWriter, r *http.Request) {
 	ctx.Request = r
-	ctx.ResponseWriter = &Response{rw, false, 0}
+	if ctx.ResponseWriter == nil {
+		ctx.ResponseWriter = &Response{}
+	}
+	ctx.ResponseWriter.reset(rw)
```
astaxie: why we need a reset? I think &Response{rw, false, 0} should the same.

JessonChan: I think reset method will reduce allocations

astaxie: make sense

context/context.go
```diff
@@ -181,15 +192,22 @@ type Response struct {
...
 
 // WriteHeader sends an HTTP response header with status code,
 // and sets `started` to true.
 func (w *Response) WriteHeader(code int) {
+	if w.Status > 0 {
+		return
+	}
 	w.Status = code
 	w.Started = true
-	w.ResponseWriter.WriteHeader(code)
```
astaxie: why remove this? when will set the header? that would be a big issue.

JessonChan: hi @astaxie the http status will be set when Write was called,look here:https://github.com/JessonChan/beego/blob/err_ctrler/context/context.go#L195

farther more,at the http/net package https://golang.org/src/net/http/server.go#L756
and https://golang.org/src/net/http/server.go#L791
the chunkWriter will clone the header if the WriteHeader was called ,so we will lost some header key-value if this line was not removed.
for example 
```
package main

import "net/http"

func main() {
	http.HandleFunc("/hello", func(w http.ResponseWriter, r *http.Request) {
		w.Header().Set("Before", "before")
		w.WriteHeader(200)
		w.Header().Set("After", "after")
		w.Write([]byte("Ok"))
	})
	http.ListenAndServe(":8090", nil)
}
```

image

astaxie: But if I just call `c.Ctx.ResponseWriter.WriteHeader(status)` and no output. what will happen?

JessonChan: it's a bug,i will try to fix this.

JessonChan: @astaxie I feel it's a little hard to fix this without a method like finishRequest (https://golang.org/src/net/http/server.go#L1243),but this method must be an expected method(FinishRequest).
So,what do you think about this?

astaxie: I think we just leave this `w.ResponseWriter.WriteHeader(code)`. lost some header is acceptable. As they know set header should before set status.

JessonChan: It's not so perfect,but I'd like to accept your idea.
I will still keep an eye on this.

JessonChan:
@astaxie 
I have reverted the change `c.Ctx.ResponseWriter.WriteHeader(status)` and add a typo fix `mulit` => `multi`.

controller.go
```diff
@@ -258,7 +258,7 @@ func (c *Controller) Abort(code string) {
 
 // CustomAbort stops controller handler and show the error data, it's similar Aborts, but support status code and body.
 func (c *Controller) CustomAbort(status int, body string) {
-	c.Ctx.ResponseWriter.WriteHeader(status)
+	c.Ctx.Output.Status = status
```
xionglun: Hi, @JessonChan 
I notice that you use `Output.Status` instead of `WriteHeader`, I don't know why.
And here cause `controller.CustomAbort(status, body)` failed! It always respond status code `200` instead of `status`.

JessonChan:
@xionglun thanks for your good question.  This is a known bug and has bee fixed https://github.com/astaxie/beego/pull/1805 ,and there is farther talk https://github.com/astaxie/beego/pull/1822

controller.go
JessonChan: @xionglun thanks for your good question. This is a known bug and has bee fixed #1805 ,and there is farther talk #1822

xionglun: @JessonChan Very good. Expect next release.
But, why not a hot-fix release? @astaxie

PR #1721, opened 2016-02-24
Log enhancement

JessonChan:
multi log files


logs/files.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+}
+
+// newFilesWriter create a FileLogWriter returning as LoggerInterface.
+func newFilesWriter() Logger {
+	return &filesLogWriter{}
+}
+
+func init() {
+	Register("files", newFilesWriter)
```
astaxie: I think we can change a better name. `mulitfile` will make more sense

JessonChan: ok

Issue #771, opened 2014-08-23
Inline Struct Support


dietsche:
I would like to propose adding an "inline" option that would allow inlining of structs. *Example 1* shows how the code works currently. In *Example2* I show how inlining could work in the future.

Example 1 - current
--------------
```
type User struct {
        //Common Fields
        Id      int64
        Created time.Time `orm:"auto_now_add;type(datetime)"`
        Updated time.Time `orm:"auto_now;type(datetime)"`

        //Other Fields
        UserName  string `orm:"unique"`
        FirstName string
        LastName  string
        Email     string
}
```

Example 2 - future
--------------
```
//Common Fields In a struct to promote re-use
type ModelBase struct {
        Id      int64
        Created time.Time `orm:"auto_now_add;type(datetime)"`
        Updated time.Time `orm:"auto_now;type(datetime)"`
}
type User struct {
        ModelBase `orm:"inline"`

        //Other Fields
        UserName  string `orm:"unique"`
        FirstName string
        LastName  string
        Email     string
}
```

Can this be done?
Thanks!

astaxie:
@dietsche it's support in 1.6.1 now

dietsche:
awesome, thanks!

PR #1723, opened 2016-02-24
orm: inline struct support

miraclesu:
#771


astaxie:
@miraclesu please add test case for this

miraclesu:
@astaxie test case has been add.

astaxie:
@miraclesu if you can add some test case in the basemodel with related table that would be awesome

PR #1731, opened 2016-03-01
fix bug: session id undecoded when destroy and sesssion memory provider push wrong

math345:
1. session id undecoded when destroy.
2. sesssion memory provider push wrong.

### DIFF
--- a/session/sess_mem.go
+++ b/session/sess_mem.go
@@ -102,7 +102,6 @@ func (pder *MemProvider) SessionRead(sid string) (Store, error) {
 	pder.lock.RUnlock()
 	pder.lock.Lock()
 	newsess := &MemSessionStore{sid: sid, timeAccessed: time.Now(), value: make(map[interface{}]interface{})}
-	// fix bug: new session should be pushed into the head of list(more fresh)
 	element := pder.list.PushFront(newsess)
 	pder.sessions[sid] = element
 	pder.lock.Unlock()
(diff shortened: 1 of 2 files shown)


session/sess_mem.go
```diff
@@ -102,7 +102,8 @@ func (pder *MemProvider) SessionRead(sid string) (Store, error) {
 	pder.lock.RUnlock()
 	pder.lock.Lock()
 	newsess := &MemSessionStore{sid: sid, timeAccessed: time.Now(), value: make(map[interface{}]interface{})}
-	element := pder.list.PushBack(newsess)
+	// fix bug: new session should be pushed into the head of list(more fresh)
```
astaxie: why add the comments here? you can explain why you modify it in the PR

session/session.go
```diff
@@ -201,7 +201,9 @@ func (manager *Manager) SessionDestroy(w http.ResponseWriter, r *http.Request) {
 	if err != nil || cookie.Value == "" {
 		return
 	}
-	manager.provider.SessionDestroy(cookie.Value)
+	// fix bug: cookie.Value has been urlencoded, so should be decoded here
+	sid, _ := url.QueryUnescape(cookie.Value)
+	manager.provider.SessionDestroy(sid)
```
astaxie: sid is use `hex.EncodeToString` and `hextable = "0123456789abcdef"`

So sid should only have these char. urlencode will change nothing

math345: yes, I check sessionID method: you are right. But it will be danger when sessionID method changed. So urlencode will be strong. thanks.

astaxie: I don't think we will change the session generate rule. and what's your concern?

math345: I don't think it's grace and unify. Because getSid/SessionRegenerateID methods has used url.QueryUnescape method.

Issue #1737, opened 2016-03-02
Url path is decoded before routering


simpleton:
If the parameter contain slash will cause router module didn't work.
eg.
I register a path like `/search/:keyword`. if `keyworkd = "你好/世界"`, the slash will make router module don't work.

simpleton:
Maybe we can change `r.URL.Path` to `r.URL.String()`

simpleton:
Here is the [reference of RawPath](https://github.com/golang/go/commit/874a605af0764a8f340c3de65406963f514e21bc)

astaxie:
I don't think `/search/你好/世界` can match `/search/:keyword`. if you want to match this router. you should set the route `/search/*`

simpleton:
@astaxie That' only the simple example. In the fact, there is a api `/v1/tags/:tag/posts`. The tag name is made by user which can be the stuff like `你好/世界`。 

The client will urlencdoe the tag name, but router module use the decode string. 
I found some framework use the `EscapedPath` as the input to router module, but this api was introduced by Go1.5.

zhlicen:
i agree
```/search/你好%2f世界``` should match ```/search/:keyword```

PR #1738, opened 2016-03-02
Change get urlPath method.

simpleton:
Use r.URL.String() instread of r.URL.Path
Fix Url path is encoded before routering #1737


astaxie:

simpleton:
ok~

PR #1742, opened 2016-03-03
自定义log输出

sipt:
1.添加 const DefaultLoggerFuncCallDepth=2
2.开放 logger输出方法
3.添加用户自定义输出级别LeveCustom
4.提取重得代码formatLogTime

### DIFF
--- a/logs/log.go
+++ b/logs/log.go
@@ -45,7 +45,8 @@ import (
 
 // RFC5424 log message levels.
 const (
-	LevelEmergency = iota
+	LevelCustom = iota //level for custom
+	LevelEmergency
 	LevelAlert
 	LevelCritical
 	LevelError
@@ -53,6 +54,7 @@ const (
 	LevelNotice
 	LevelInformational
 	LevelDebug
+	DefaultLoggerFuncCallDepth = 2
 )
 
 // Legacy loglevel constants to ensure backwards compatibility.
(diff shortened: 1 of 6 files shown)


logs/log.go
```diff
@@ -199,6 +201,15 @@ func (bl *BeeLogger) writeMsg(logLevel int, msg string) error {
...
 		msg = "[" + filename + ":" + strconv.FormatInt(int64(line), 10) + "]" + msg
 	}
+	bl.OutputMsg(msg, logLevel)
+}
+
+//open output log for custom
+func (bl *BeeLogger) OutputMsg(msg string, logLevel int) {
+	when := time.Now()
+	if LevelCustom != logLevel {
+		msg = formatLogTime(when) + msg
```
JessonChan: msg can't append a time header,check the code in logs/es/es.go L57~60

logs/log.go
```diff
@@ -45,7 +45,8 @@ import (
 
 // RFC5424 log message levels.
 const (
-	LevelEmergency = iota
+	LevelCustom = iota //level for custom
```
JessonChan: what happens when the log level set to a LevelCustom?

sipt: Users can customize the level of log;
For example,
the level of business log: [INFO][DEBUG][ERROR]....  
the level of database log:[SQL] [PARAMS] [RESULT] [INSERT] [SELECT] [UPDATE]...
their can be different.

logs/log.go
sipt: missed it....

sipt: missed it.
Custom time format is a problem

logs/log.go
JessonChan: It's a good idea,but the code is not so elegant.

sipt:
@JessonChan 
I have restored "formatLogTime".

Issue #1748, opened 2016-03-04
并发时异常退出。


zhangchun:
程序是bee new的一个test工程。
单用户浏览没有问题，用ab压下就出异常了。
`2016/03/04 10:12:14 [asm_amd64.s:1998][I] http server Running on :8200
2016/03/04 10:12:15 [router.go:828][D] | GET        | /                                        | 1.302665ms       | match      | /                                        |
2016/03/04 10:12:15 [router.go:828][D] | GET        | /                                        | 1.708893ms       | match      | /                                        |
2016/03/04 10:12:15 [router.go:828][D] | GET        | /                                        | 751.649µs        | match      | /                                        |
2016/03/04 10:12:15 [router.go:828][D] | GET        | /                                        | 1.628675ms       | match      | /                                        |
2016/03/04 10:12:15 [router.go:828][D] | GET        | /                                        | 1.988417ms       | match      | /                                        |
2016/03/04 10:12:15 [router.go:828][D] | GET        | /                                        | 687.008µs        | match      | /                                        |
fatal error: concurrent map read and map write

goroutine 9 [running]:
runtime.throw(0xaa50a0, 0x21)
        /usr/local/go/src/runtime/panic.go:530 +0x90 fp=0xc8204b7478 sp=0xc8204b7460
runtime.mapaccess1_faststr(0x87a5c0, 0xc82014e150, 0xa2e860, 0x9, 0xc820123a88)
        /usr/local/go/src/runtime/hashmap_fast.go:202 +0x5b fp=0xc8204b74d8 sp=0xc8204b7478
github.com/astaxie/beego.(*Controller).RenderBytes(0xc8204340d0, 0x0, 0x0, 0x0, 0x0, 0x0)
        /home/gowork/src/github.com/astaxie/beego/controller.go:265 +0x1893 fp=0xc8204b7760 sp=0xc8204b74d8
github.com/astaxie/beego.(*Controller).Render(0xc8204340d0, 0x0, 0x0)
        /home/gowork/src/github.com/astaxie/beego/controller.go:183 +0x4e fp=0xc8204b77c0 sp=0xc8204b7760
github.com/astaxie/beego.(*ControllerRegister).ServeHTTP(0xc82014e240, 0x7f2fb7500130, 0xc820434000, 0xc8203fa000)
        /home/gowork/src/github.com/astaxie/beego/router.go:784 +0x1d7c fp=0xc8204b7b60 sp=0xc8204b77c0
[...]

PR #1752, opened 2016-03-04
fix template  read-write lock race

JessonChan:
#1748 
this is crash is caused by the beeTemplates read-write race in "DEV" mode.

1.I rename the BeeTemplates to beeTemplates,because this var should be unexpected.so is BeeTemplateExt
2.I extract some code to function "renderTemplate" and "executeTemplate"

### DIFF
--- a/config.go
+++ b/config.go
@@ -16,7 +16,6 @@ package beego
 
 import (
 	"fmt"
-	"html/template"
 	"os"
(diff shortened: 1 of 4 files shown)


astaxie:
@JessonChan I mearged this now. But I still think we need to add section test case.

PR #1762, opened 2016-03-07
Implemented possibility to add custom template engines

saturn4er:
Go templates are good, but jade/haml is better, I think. So here is resolution to add custom html compilers.

### DIFF
--- a/template.go
+++ b/template.go
@@ -167,9 +167,11 @@ func BuildTemplate(dir string, files ...string) error {
 		for _, file := range v {
 			if buildAllFiles || utils.InSlice(file, files) {
 				templatesLock.Lock()
-				fileExt := filepath.Ext(file)[1:]
+				ext := filepath.Ext(file)
 				var t TemplateRenderer
-				if fn, ok := beeTemplateEngines[fileExt]; ok {
+				if len(ext) == 0 {
+					t, err = getTemplate(self.root, file, v...)
+				} else if fn, ok := beeTemplateEngines[ext[1:]]; ok {
 					t, err = fn(self.root, file, beegoTplFuncMap)
 				} else {
 					t, err = getTemplate(self.root, file, v...)
--- a/template.go
+++ b/template.go
@@ -31,13 +31,13 @@ import (
 
 var (
 	beegoTplFuncMap = make(template.FuncMap)
-	// beeTemplates caching map and supported template file extensions.
+// beeTemplates caching map and supported template file extensions.
 	beeTemplates = make(map[string]TemplateI)
 	templatesLock sync.RWMutex
-	// beeTemplateExt stores the template extension which will build
+// beeTemplateExt stores the template extension which will build
 	beeTemplateExt = []string{"tpl", "html"}
-	// BeeTemplatePreprocessors stores associations of extension -> preprocessor handler
-	BeeTemplateEngines = map[string]func(root, path string) (TemplateI, error){}
+// BeeTemplatePreprocessors stores associations of extension -> preprocessor handler
+	BeeTemplateEngines = map[string]func(root, path string, funcs template.FuncMap) (TemplateI, error){}
 )
 
 func executeTemplate(wr io.Writer, name string, data interface{}) error {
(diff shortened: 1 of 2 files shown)


template.go
```diff
@@ -31,11 +31,13 @@ import (
 
 var (
 	beegoTplFuncMap = make(template.FuncMap)
-	// beeTemplates caching map and supported template file extensions.
-	beeTemplates  = make(map[string]*template.Template)
+// beeTemplates caching map and supported template file extensions.
```
astaxie: it looks you didn't use gofmt with the file

saturn4er:
@astaxie Fixed

template.go
```diff
@@ -32,10 +32,12 @@ import (
 var (
 	beegoTplFuncMap = make(template.FuncMap)
 	// beeTemplates caching map and supported template file extensions.
-	beeTemplates  = make(map[string]*template.Template)
+	beeTemplates  = make(map[string]TemplateI)
 	templatesLock sync.RWMutex
 	// beeTemplateExt stores the template extension which will build
 	beeTemplateExt = []string{"tpl", "html"}
+	// BeeTemplatePreprocessors stores associations of extension -> preprocessor handler
+	BeeTemplateEngines = map[string]func(root, path string, funcs template.FuncMap) (TemplateI, error){}
```
astaxie: define `func(root, path string, funcs template.FuncMap)` as a function type?

astaxie: And I think `BeeTemplateEngines` don't need to export

saturn4er:
@astaxie Ok?

template.go
```diff
@@ -32,10 +32,12 @@ import (
 var (
 	beegoTplFuncMap = make(template.FuncMap)
 	// beeTemplates caching map and supported template file extensions.
-	beeTemplates  = make(map[string]*template.Template)
+	beeTemplates  = make(map[string]TemplateI)
 	templatesLock sync.RWMutex
 	// beeTemplateExt stores the template extension which will build
 	beeTemplateExt = []string{"tpl", "html"}
+	// BeeTemplatePreprocessors stores associations of extension -> preprocessor handler
```
astaxie: please also fix the comments name. It still use BeeTemplatePreprocessors

template.go
```diff
@@ -156,11 +161,18 @@ func BuildTemplate(dir string, files ...string) error {
 		fmt.Printf("filepath.Walk() returned %v\n", err)
 		return err
 	}
+	buildAllFiles := len(files) == 0
```
JessonChan: we should avoid creating intermediate variable

template.go
```diff
@@ -32,10 +32,12 @@ import (
 var (
 	beegoTplFuncMap = make(template.FuncMap)
 	// beeTemplates caching map and supported template file extensions.
-	beeTemplates  = make(map[string]*template.Template)
+	beeTemplates  = make(map[string]TemplateI)
```
JessonChan: interface  in golang is named with -er ,such as io.Writ**er**,io.Read**er**,http.Flush**er**

JessonChan:
@saturn4er maybe you can git rebase to merge some commits into fewer commits(squash or fixup the go-formating commit)

template.go
saturn4er: On each loops iterations Go will recalculate len(), isn't it?

JessonChan: It's not much of a problem.len() is fast enough as a  builtin function and this BuildTemplate function  will be called by the init method in product environment.

astaxie:
@saturn4er could you also help to provider the example how to add jade/haml?

saturn4er:
@astaxie https://gist.github.com/saturn4er/05e545984922784ace9b

template.go
```diff
@@ -156,11 +162,18 @@ func BuildTemplate(dir string, files ...string) error {
...
 		return err
 	}
+	buildAllFiles := len(files) == 0
 	for _, v := range self.files {
 		for _, file := range v {
-			if len(files) == 0 || utils.InSlice(file, files) {
+			if buildAllFiles || utils.InSlice(file, files) {
 				templatesLock.Lock()
-				t, err := getTemplate(self.root, file, v...)
+				fileExt := filepath.Ext(file)[1:]
```
astaxie: if the template has no ext. it will crash here.

astaxie: @saturn4er could you try to fix this?

PR #1766, opened 2016-03-08
fix json extra newline

aolu11:
POST json body now use json.Encoder#Encode()

This will result a trailing newline. It's mostly for json stream. Not useful in HTTP and will cause some problem（like break data parse protocol）.

Should use json.Marshal() instead.


astaxie:

PR #1782, opened 2016-03-11
BeeLogger can be the writer of  http server's log 

JessonChan:
the change is to make BeeLogger implements io.Writer,so the net/http package will output to the beego's logger

### DIFF
--- a/app.go
+++ b/app.go
@@ -16,6 +16,7 @@ package beego
 
 import (
 	"fmt"
+	"log"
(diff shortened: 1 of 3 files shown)


app.go
```diff
@@ -23,6 +23,8 @@ import (
 	"path"
 	"time"
 
+	"log"
+
```
astaxie: `log` also need to assign with other packages

PR #1791, opened 2016-03-13
Issue1787： if `/m` is a static path mapping to a static dir，and we set `beego.BConfig.WebConfig.DirectoryIndex = true`, then it should redirect to `/m/`

youngsterxyf:
R.T.

### DIFF
--- a/context/output.go
+++ b/context/output.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"html/template"
+	"io"
 	"mime"
 	"net/http"
 	"path/filepath"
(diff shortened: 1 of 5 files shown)


context/output.go
```diff
@@ -72,10 +73,11 @@ func (output *BeegoOutput) Body(content []byte) error {
 	if output.Status != 0 {
 		output.Context.ResponseWriter.WriteHeader(output.Status)
 		output.Status = 0
+	} else {
+		output.Context.ResponseWriter.Started = true
```
JessonChan: i don't think we need to this else-line

youngsterxyf: Oh，no, this pr has been closed.

PR #1792, opened 2016-03-13
fix issue#1787: if `/m` is a static path mapping to a static dir，and we set `beego.BConfig.WebConfig.DirectoryIndex = true`, then it should redirect to `/m/`

youngsterxyf:
R.T.

fix issue #1787


JessonChan:
LGTM

PR #1799, opened 2016-03-15
Router Filter Improve

JessonChan:
I change the `ControllerRegister.filters` from map to array,which makes filters-access  7.5 times faster than before , I think. Here is test code 

```
package main

import (
	"fmt"
	"time"
)

func main() {
	fm := map[int][]int{
		0: []int{0},
		1: []int{0, 1},
		2: []int{0, 1, 2},
		3: []int{0, 1, 2, 3},
		4: []int{0, 1, 2, 3, 4},
	}
	fi := [5][]int{{0}, {0, 1}, {0, 1, 2}, {0, 1, 2, 3}, {0, 1, 2, 3, 4}}
	max := int(1e7)
	{
		t := time.Now()
		for i := 0; i < max; i++ {
			for l := 0; l < 5; l++ {
				if _, ok := fm[l]; ok {
				}
			}
		}
		fmt.Println(time.Now().UnixNano() - t.UnixNano())
	}
	{
		t := time.Now()
		for i := 0; i < max; i++ {
			for l := 0; l < 5; l++ {
				if len(fi[l]) > 0 {
				}
			}
		}
		fmt.Println(time.Now().UnixNano() - t.UnixNano())
	}
}
```

The result
image

### DIFF
--- a/router.go
+++ b/router.go
@@ -114,7 +114,7 @@ type controllerInfo struct {
 type ControllerRegister struct {
 	routers      map[string]*Tree
 	enableFilter bool
-	filters      map[int][]*FilterRouter
+	filters      [5][]*FilterRouter
 	pool         sync.Pool
 }
 
@@ -122,7 +122,6 @@ type ControllerRegister struct {
 func NewControllerRegister() *ControllerRegister {
 	cr := &ControllerRegister{
 		routers: make(map[string]*Tree),
-		filters: make(map[int][]*FilterRouter),
 	}
 	cr.pool.New = func() interface{} {
 		return beecontext.NewContext()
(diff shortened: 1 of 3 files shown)


router.go
```diff
@@ -617,11 +615,10 @@ func (p *ControllerRegister) ServeHTTP(rw http.ResponseWriter, r *http.Request)
 		context.Output.Header("Server", BConfig.ServerName)
 	}
 
-	var urlPath string
+	var urlPath = r.URL.Path
+
 	if !BConfig.RouterCaseSensitive {
 		urlPath = strings.ToLower(r.URL.Path)
```
astaxie: `urlPath = strings.ToLower(urlPath)`

PR #1803, opened 2016-03-16
fix rotate  read-write lock race

JessonChan:
rewrite the  double-lock check for go1.6
 https://misfra.me/optimizing-concurrent-map-access-in-go/


JessonChan:
@youngsterxyf @ysqi

Issue #1230, opened 2015-06-23
Gzip compression level 


mlgd:
Add property in config to set Gzip compression level

JessonChan:
thank you for the issue.
we have a plan to support this after the beego1.6 release.

mlgd:
What's about gzip compression level in version 1.6.x ?

JessonChan:
@mlgd maybe 1.6.3

JessonChan:
@mlgd I have add compression level future https://github.com/astaxie/beego/pull/1808
I would appreaciate hearing your thoughts on this.

PR #1808, opened 2016-03-17
Gzip improve

JessonChan:
https://github.com/astaxie/beego/issues/1230

### DIFF
--- a/context/acceptencoder.go
+++ b/context/acceptencoder.go
@@ -25,7 +25,34 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+
+	"github.com/astaxie/beego/config"
 )
+
+var (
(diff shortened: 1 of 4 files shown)


context/acceptencoder.go
```diff
@@ -25,8 +25,36 @@ import (
...
+	//Default size==20B same as nginx
+	defaultGzipMinLength = 20
+	//The compression level used for deflate compression. (0-9).
+	gzipCompressLevel int
+	//List of HTTP methods to compress. If not set, only GET requests are compressed.
+	includedMethods map[string]bool
+	getMethodOnly   bool
 )
 
+func InitGzip(cf config.Configer) {
```
astaxie: it lead into the config package into context. actually I want to keep all modules independence. is it possible we just pass the value while not config.Configer.

Issue #1807, opened 2016-03-17
orm.read() 在模型中主键为负数时，调用返回错误missed pk value？


zuoruochen:
type Channel struct {
	ChannelId	int	`orm:"pk";"column(channel_id)"`
	Abc		float64	`orm:"column(Abc)"`
}

channel := Channel{ChannelId:-777}
err := o.Read(&channel)
会返回missed pk value，将-777改为正数就没问题

miraclesu:
It is a bug, we will fix it soon.
Thank you for your feedback :)
@astaxie

PR #1810, opened 2016-03-17
orm: fix miss pk when pk is negative

miraclesu:
#1807


astaxie:
@miraclesu could you add the test case for negative pk

miraclesu:
@astaxie the test case have been add.

PR #1821, opened 2016-03-21
Log errors of session update process

tamoshius:
It is hard to determine the root of the problems of configuaration, etc. if return is nil. The errors should be logged and displayed.


astaxie:
I want to add a logger in session. So user can set the logger to file or other resource. @youngsterxyf please send a PR for this

youngsterxyf:
@astaxie OK，i will try to do it.

Issue #1798, opened 2016-03-14
responseWriter wrapper for http.ResponseWriter in rourter.go


goodloop:
in 
```
func (p *ControllerRegistor) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
    ...
}
```
there are wrapper for http.ResponseWriter. and there are a field named **started**.

here are my questions about this:
1. why this field are needed? for concurrency issue?
2. why can't update response's header after calling wrapper's WriteHeader function? Is that by design?

Eager to get your answer!
Best Wishes,
-gavin

JessonChan:
1.This field is a flag to know whether the writer has been called.If you call the writer to write the response,beego will check this flag and do not write anything to the response.
2.May be you can call this is a bug  of beego,to be exact,a defect of package http/net.  For more discuss https://github.com/astaxie/beego/pull/1719,I will try to fix in later.

goodloop:
thanks for quick reply!
So seems to be some programming pattern for http header issue: we must put WriteHeader function after all of the Header settings.
Is that make sense?

JessonChan:
That's right. https://golang.org/src/net/http/server.go?s=47192:47244#L1615

JessonChan:
https://github.com/astaxie/beego/pull/1822   try to fix this

PR #1822, opened 2016-03-21
just for TALK not for MERGE

JessonChan:
talk about remove beego/context.Response(http.ResponseWriter) to beego package and rename to an unexpected struct.
**IT'S  EXPERIMENTED**


JessonChan:
https://github.com/astaxie/beego/issues/1798

PR #1823, opened 2016-03-21
Develop

e154:
with a large number of configuration files in the directory dev, staging, prod, convenient to use a single configuration file

### DIFF
--- a/config.go
+++ b/config.go
@@ -103,6 +103,8 @@ var (
 	BConfig *Config
 	// AppConfig is the instance of Config, store the config information from file
 	AppConfig *beegoAppConfig
+	// Common config
+	CommonConfig *beegoAppConfig
 	// AppPath is the absolute path to the app
 	AppPath string
 	// GlobalSessions is the instance for the session manager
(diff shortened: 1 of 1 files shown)


config.go
```diff
@@ -180,6 +182,15 @@ func init() {
 	}
 
 	appConfigPath = filepath.Join(AppPath, "conf", "app.conf")
+	commonConfigPath := filepath.Join(AppPath, "conf", "common.conf")
+	if utils.FileExists(commonConfigPath) {
+		CommonConfig, err := newAppConfig(appConfigProvider, commonConfigPath)
```
ysqi: Why not offter the test case ? CommonConfig is local variables  , Not use `:=`

JessonChan:
This is not a very good idea in my opinion.

ysqi:
hey @e154  ,Why need  common.conf  ?  You can use config item  `include` to load config from other fiel. such as like this:
```ini
AppName = crm
include "/var/app/crm/crm.config"
```

e154:
Hi friends

: = - Yes it is a typo

This patch allows you to use separate directory configurations.
if you can use this architecture by standard means, I'll be glad to hear how to do it

config

Issue #1046, opened 2015-02-26
[Feature Request] ORM: Ability to Manually Insert Id


mewben:
Use Case:

```
o := orm.NewOrm()
var user User
user.Id = 479
user.Name = "slene"
user.IsActive = true

id, err := o.Insert(&user)
if err == nil {
    fmt.Println(id)
}
```

This allows you to define the Id if you don't want Id to be auto_increment. This is useful too if your primary key is uuid.

Is this possible? Thanks.

mewben:
Thanks. More power.

hsen-dev:
++1

PR #1826, opened 2016-03-23
orm: support insert a specified value to auto field

miraclesu:
#1046


astaxie:
@miraclesu I think you also need to update these function:

1. InsertStmt
2. InsertMulti

I think these function also call collectValues. And add test case for these two functions

miraclesu:
@astaxie `InsertMulti` has updated.
`InsertStmt` never can be used, because orm doesn't provide a `InsertStmt` function.
I will implement it later.

PR #1832, opened 2016-03-24
Log enhancement

JessonChan:
1、move log function(Debug/Error...)to beego/logs
2、make the log function  more convenient
3、add a function to get official log.Logger
4、set console as default logger if none logger was set

 Example  code

```
package main

import "github.com/astaxie/beego/logs"

func main() {
	logs.Warn("this is a warn message")

	l := logs.GetLogger("HTTP")
	l.Println("this is a message of http")

	logs.GetLogger("orm").Println("this is a message of orm")

	logs.Debug("my book is bought in the year of ", 2016)
	logs.Info("this %s cat is %v years old", "yellow", 3)
	logs.Error(1024, "is a very", "good", 2.5, map[string]int{"id": 1})
	logs.Critical("oh my god")
}
```

image


log.go
```diff
@@ -32,83 +32,75 @@ const (
 	LevelDebug
 )
 
-// BeeLogger references the used application logger.
-var BeeLogger = logs.NewLogger(100)
```
astaxie: I think we still need BeeLogger. As there's so many application rely on this variable.

astaxie: var BeeLogger = logs.GetBeeLogger()

JessonChan: Yes,I have just added this line

Issue #1836, opened 2016-03-28
some puzzled


BPing:
Please answer these questions before submitting your issue. Thanks!

1. What version of Go and beego are you using (`bee version`)?
  1.6.1

2. What operating system and processor architecture are you using (`go env`)?
```
set GOARCH=amd64
set GOBIN=
set GOEXE=.exe
set GOHOSTARCH=amd64
set GOHOSTOS=windows
set GOOS=windows
set GOPATH=F:\workplace\library\gopath;F:\workplace\www\gowork
set GORACE=
set GOROOT=F:\workplace\resources\Go
set GOTOOLDIR=F:\workplace\resources\Go\pkg\tool\windows_amd64
set GO15VENDOREXPERIMENT=1
set CC=gcc
set GOGCCFLAGS=-m64 -mthreads -fmessage-length=0
set CXX=g++
set CGO_ENABLED=1
```
3. What did you do?

beego changes go's currently running path  with the dir of "os.Args[0]"
```go
AppPath, _ = filepath.Abs(filepath.Dir(os.Args[0]))
os.Chdir(AppPath)
```
the value of os.Args[0] :
C:\Users\DREAM-~1\AppData\Local\Temp\go-build364476898\command-line-arguments\_obj\exe\main.exe

but my workplace‘s path is :F:\workplace\www\gowork\src

so i can't open the file with relative path,
like:"app/conf/app.conf"(F:\workplace\www\gowork\src\app\conf\app.conf)

i don't want to modify the source code .

JessonChan:
https://github.com/astaxie/beego/issues/1789

BPing:
the problem is  present  when I run the application,not for testing

BPing:
it will have the same problem,even if I build a .exe file to my working directory,

JessonChan:
@BPing go build will be ok and this has been fixed on develop branch

PR #1840, opened 2016-03-29
To support `go run`

youngsterxyf:
After running, beego does not change current dir to `AppPath`, and firstly try to parse config file in `workPath`, then `AppPath`.

When using `go run`, the `AppPath` is a temp dir, under that path, there is no config file.

### DIFF
--- a/config.go
+++ b/config.go
@@ -106,6 +106,8 @@ var (
 	AppConfig *beegoAppConfig
 	// AppPath is the absolute path to the app
 	AppPath string
+	// workPath is the absolute path of the current dir
+	workPath string
 	// GlobalSessions is the instance for the session manager
 	GlobalSessions *session.Manager
 
@@ -118,7 +120,8 @@ var (
 func init() {
 	AppPath, _ = filepath.Abs(filepath.Dir(os.Args[0]))
 
-	os.Chdir(AppPath)
+	workPath, _ = os.Getwd()
+	workPath, _ = filepath.Abs(workPath)
 
 	BConfig = &Config{
 		AppName:             "beego",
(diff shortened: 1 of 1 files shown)


config.go
```diff
@@ -118,7 +120,8 @@ var (
 func init() {
 	AppPath, _ = filepath.Abs(filepath.Dir(os.Args[0]))
 
-	os.Chdir(AppPath)
+	workPath, _ = os.Getwd()
+	workPath, _ = filepath.Abs(workPath)
```
JessonChan: no need filepath.Abs(workPath)

config.go
```diff
@@ -106,6 +106,8 @@ var (
 	AppConfig *beegoAppConfig
 	// AppPath is the absolute path to the app
 	AppPath string
+	// workPath is the absolute path of the current dir
+	workPath string
```
JessonChan: local variable is ok

config.go
```diff
@@ -118,7 +120,8 @@ var (
 func init() {
 	AppPath, _ = filepath.Abs(filepath.Dir(os.Args[0]))
```
JessonChan: L121 can be moved after L185 (closer to where the variable used)

config.go
youngsterxyf: yes，thanks

config.go
```diff
@@ -180,10 +178,16 @@ func init() {
 		},
 	}
 
-	appConfigPath = filepath.Join(AppPath, "conf", "app.conf")
+	AppPath, _ = filepath.Abs(filepath.Dir(os.Args[0]))
+	workPath, _ = os.Getwd()
```
astaxie: you can change to `workPath, err : = os.Getwd()`

remove the workPath declare. and also I think we need to check all the err. if has the err just panic at start the application.

youngsterxyf: ok

JessonChan:
https://github.com/astaxie/beego/issues/1836

PR #1842, opened 2016-03-30
delete not used variable (status int) in output.go functions

mishudark:


astaxie:
please send the PR to develop

PR #1843, opened 2016-03-30
Beelog bug fix

JessonChan:
And make the format time header at least %5 faster than before on my mac
It's a feature bug.


logs/logger.go
```diff
@@ -36,43 +36,46 @@ func (lg *logWriter) println(when time.Time, msg string) {
...
-	buf[0] = byte('0' + y)
+
+	//change to '3' after 984 years, LOL
+	buf[0] = '2'
+	//change to '1' after 84 years,  LOL
+	buf[1] = '0'
+	buf[2] = y1[y-2000]
+	buf[3] = y2[y-2000]
 	buf[4] = '/'
-	if mo > 9 {
```
JessonChan: there is a bug when mo==10

Issue #1547, opened 2016-01-08
Missing mapping for mysql Time type


b055:
There is no mapping in beego orm for the mysql data type Time.
http://dev.mysql.com/doc/refman/5.7/en/time.html

I've checked the beego documentation at http://beego.me/docs/mvc/model/models.md#mysql
and the model fields at 
https://github.com/astaxie/beego/blob/master/orm/models_fields.go#L34

I see nothing that maps to the Time data type in mysql.

My model is defined as 
```type Times struct {
	Open         time.Time `orm:"column(open);type(time)"`
 }```
and my connection string already has the parameter ```parseTime=true```

When trying to get a row of data from the database with a time column, the following error shows up:
```
panic: Raw value: `[48 53 58 51 48 58 48 48]` convert to `*time.Time` failed, field: main/juice/models.Times.Open err: parsing time "05:30:00" as "2006-01-02": cannot parse "0:00" as "2006"

goroutine 1 [running]:
github.com/astaxie/beego/orm.(*dbBase).setColsValues(0xc8200d9100, 0xc8201b72c0, 0xc8201d44d8, 0xc820109080, 0x7, 0x8, 0xc8202a6230, 0x7, 0x7, 0xc8201f7140)
	/Users/blah/workspace/blah/_vendor/src/github.com/astaxie/beego/orm/db.go:1032 +0x428
github.com/astaxie/beego/orm.(*dbBase).ReadBatch(0xc8200d9100, 0x4f85bd8, 0xc8201e1d80, 0xc8201f7bc0, 0xc8201b72c0, 0xc8202a5100, 0x466c260, 0xc8202a5120, 0xc8201f7140, 0x0, ...)
	/Users/blah/workspace/blah/_vendor/src/github.com/astaxie/beego/orm/db.go:852 +0x2610
github.com/astaxie/beego/orm.(*querySet).All(0xc8201f7bc0, 0x466c260, 0xc8202a5120, 0x0, 0x0, 0x0, 0x4f85cb0, 0x0, 0x0)
	/Users/blah/workspace/blah/_vendor/src/github.com/astaxie/beego/orm/orm_queryset.go:174 +0xfa
```

There seems to be no way to map mysql data type time to golang time.Time via beego orm

PR #1856, opened 2016-04-03
added functionality for column type time

b055:
Here is the original issue https://github.com/astaxie/beego/issues/1547

This enabled the use of the time column type in databases. Tested sqlite3 and mysql. I don't have a postgres setup, but travis should take care of that.
updated the model_fields to cater for the time field type

### DIFF
--- a/namespace_test.go
+++ b/namespace_test.go
@@ -61,8 +61,8 @@ func TestNamespaceNest(t *testing.T) {
 	ns.Namespace(
 		NewNamespace("/admin").
 			Get("/order", func(ctx *context.Context) {
-			ctx.Output.Body([]byte("order"))
-		}),
+				ctx.Output.Body([]byte("order"))
+			}),
 	)
 	AddNamespace(ns)
 	BeeApp.Handlers.ServeHTTP(w, r)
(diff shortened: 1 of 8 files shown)


namespace_test.go
```diff
@@ -124,8 +124,8 @@ func TestNamespaceFilter(t *testing.T) {
 		ctx.Output.Body([]byte("this is Filter"))
 	}).
 		Get("/user/:id", func(ctx *context.Context) {
-		ctx.Output.Body([]byte(ctx.Input.Param(":id")))
-	})
+			ctx.Output.Body([]byte(ctx.Input.Param(":id")))
+		})
```
astaxie: fmt error....

orm/models_fields.go
```diff
@@ -46,9 +47,9 @@ const (
 
 // Define some logic enum
 const (
-	IsIntegerField         = ^-TypePositiveBigIntegerField >> 4 << 5
-	IsPositiveIntegerField = ^-TypePositiveBigIntegerField >> 8 << 9
-	IsRelField             = ^-RelReverseMany >> 14 << 15
+	IsIntegerField         = ^-TypePositiveBigIntegerField >> 5 << 6
+	IsPositiveIntegerField = ^-TypePositiveBigIntegerField >> 9 << 10
+	IsRelField             = ^-RelReverseMany >> 15 << 16
```
astaxie: Why change these value?

lluishernandez: It looks he did it having in mind the addition of the `TypeTimeField` in the group definition above, so all the indexes will be altered by +1.

b055: Exactly what @lluishernandez said.
Adding `TypeTimeField` to the list of iota constants at line 28 of that same file means that those definitions needed to be changed by +1

astaxie: @b055 @lluishernandez thanks for update. that make sense

PR #1874, opened 2016-04-11
change limit 1000 to 1,reduce the amount the data

Maxgis:
The limit default value is 1000. I think the num of record is 1.

### DIFF
--- a/orm/orm_queryset.go
+++ b/orm/orm_queryset.go
@@ -192,15 +192,10 @@ func (o *querySet) All(container interface{}, cols ...string) (int64, error) {
 // query one row data and map to containers.
 // cols means the columns when querying.
 func (o *querySet) One(container interface{}, cols ...string) error {
-	if o.limit != 1 {
-		o.limit = 2
-	}
+	o.limit = 1
 	num, err := o.orm.alias.DbBaser.ReadBatch(o.orm.db, o, o.mi, o.cond, container, o.orm.alias.TZ, cols)
 	if err != nil {
 		return err
-	}
-	if num > 1 {
-		return ErrMultiRows
 	}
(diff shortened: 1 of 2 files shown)


miraclesu:
LGTM

Maxgis:
@miraclesu ok, I add test case torommow

astaxie:
@Maxgis test failed. please make sure you have tested in local

Maxgis:
@astaxie updated and test success.

astaxie:
@Maxgis this update looks trick. @miraclesu could you help to refactor this function. I think One need to limit=1. but looks the test case failed.

Maxgis:
@astaxie I think that is is not need to throw ErrMultiRows.It will update the function and test case

astaxie:
@Maxgis yea. I agree. please help to update the function and test case.

Maxgis:
If not not specified collation, different databases return different.

orm/orm_queryset.go
```diff
@@ -192,13 +192,11 @@ func (o *querySet) All(container interface{}, cols ...string) (int64, error) {
...
 // cols means the columns when querying.
 func (o *querySet) One(container interface{}, cols ...string) error {
+	o.limit = 1
 	num, err := o.orm.alias.DbBaser.ReadBatch(o.orm.db, o, o.mi, o.cond, container, o.orm.alias.TZ, cols)
 	if err != nil {
 		return err
 	}
-	if num > 1 {
-		return ErrMultiRows
-	}
```
astaxie: if we set limit to 1, and then the num return more than 1. I think should be return the `ErrMultiRows `

Maxgis:
@astaxie You are right

Issue #1466, opened 2015-12-02
orm.RunSyncdb create varchar field instead of json or jsonb


fereidani:
Hi ,
i'm trying to make a postgres jsonb column with beego orm.RunSyncdb but orm create field as varchar(255) instead .
```
type Users struct {
	Id     int
	Birthday     time.Time
	Name         string `orm:"size(50)"`
	Metadata	string `orm:"type(jsonb)"`
}
```
i only need to add column to database as jsonb and retrieve it as string for further processing ...
is this a bug ?

astaxie:
it's not support jsonb now

fereidani:
thank you for your answer
isn't it better to make type as developer input tags instead of forced type selection?
for example i just want make json column type and seeing it as string is enough for me .
as i know pq support this behavior (retrieving json value as string).

astaxie:
@fereidani we will add this proposal into our next version planning

fereidani:
it will be great , thank you , let me know if i can provide any help ...

mailbaoer:
wo, happy to see this answer！:star2:

astaxie:
support in beego1.7.0 now

PR #1875, opened 2016-04-11
orm: add json & jsonb type support

miraclesu:
#1466

### DIFF
--- a/orm/cmd_utils.go
+++ b/orm/cmd_utils.go
@@ -90,6 +90,18 @@ checkColumn:
 		} else {
 			col = fmt.Sprintf(s, fi.digits, fi.decimals)
 		}
+	case TypeJsonField:
+		if al.Driver != DRPostgres {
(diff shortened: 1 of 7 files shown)


orm/cmd_utils.go
```diff
@@ -90,6 +90,18 @@ checkColumn:
 		} else {
 			col = fmt.Sprintf(s, fi.digits, fi.decimals)
 		}
+	case TypeJSONField:
+		if al.Driver != DRPostgres {
```
astaxie: mysql also support JSON now. https://dev.mysql.com/doc/refman/5.7/en/json.html.

vefaliahmet: can you add custom column type support?

miraclesu:
@vefaliahmet I think the `Fielder` interface is what you want:
https://github.com/astaxie/beego/blob/develop/orm/models_test.go#L81

PR #1897, opened 2016-04-21
add/get session id into/from http header, check the session name in http header

yuyongsheng:
1. remove the invalid comments.
2. allow the user to config "Enable" store/get sessionId into/from http header
3. allow the user to config "Enable" get sessionId from Url query
4. when enable sessionId in http header, check the sessionName format as CanonicalMIMRHeaderKey, then panic if not.


astaxie:
👍

PR #1900, opened 2016-04-22
close socket when http client request over

redfoxli:
Transport use link pool, so we use http client to send request, but we don't close the socket.
Until the server reach keepalive_timeout, and close the socket.  If you use tcpdump, you will find the phenomenon.

Set MaxIdleConnsPerHost = -1 can solve the problem

Test code:
``` go
package main

import (
	"fmt"
	"github.com/astaxie/beego/httplib"
	"log"
	"net/http"
)

func HttpClientTest(w http.ResponseWriter, r *http.Request) {
	req := httplib.Get("http://x.x.x.x:x")
	str, err := req.String()
	if err != nil {
		fmt.Println(err)
		return
	}
	fmt.Println(str)
}

func main() {
	http.HandleFunc("/", HttpClientTest)
	err := http.ListenAndServe(":7123", nil)
	if err != nil {
		log.Fatal("ListenAndServe: ", err)
	}
}

```

### DIFF
--- a/httplib/httplib.go
+++ b/httplib/httplib.go
@@ -409,9 +409,10 @@ func (b *BeegoHTTPRequest) DoRequest() (*http.Response, error) {
 	if trans == nil {
 		// create default transport
 		trans = &http.Transport{
-			TLSClientConfig: b.setting.TLSClientConfig,
-			Proxy:           b.setting.Proxy,
-			Dial:            TimeoutDialer(b.setting.ConnectTimeout, b.setting.ReadWriteTimeout),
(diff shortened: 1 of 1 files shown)


astaxie:
@redfoxli `b.setting.Transport` is for personal config. I think maybe we can update the logic line 419-429

redfoxli:
@astaxie Yes, b.setting.Transport is for personal config, but if it don't set by user, will lead socket leak. 
Now, the default value is not reasonable. I just want to  set up reasonable defaults

PR #1909, opened 2016-04-26
Start timer task Multiple times

albertma:
Fix start time task twice when call   toolbox.StartTask() twice. One is in Admin module of beego, and the other is called by me. And I did not aware of admin module of beego call the StartTask(). 

So I fix it  as regardless of times that StartTask() has been called, it only start one time, create one routine to handle the elapsed event.

### DIFF
--- a/toolbox/task.go
+++ b/toolbox/task.go
@@ -389,11 +389,16 @@ func dayMatches(s *Schedule, t time.Time) bool {
 
 // StartTask start all tasks
 func StartTask() {
+    if isstart {
+        //If already started， no need to start another goroutine.
+        return
+    }
 	isstart = true
 	go run()
 }
 
 func run() {
+
 	now := time.Now().Local()
 	for _, t := range AdminTaskList {
 		t.SetNext(now)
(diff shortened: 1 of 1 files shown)


toolbox/task.go
```diff
@@ -389,11 +389,16 @@ func dayMatches(s *Schedule, t time.Time) bool {
...
+    if isstart {
+        //If already started， no need to start another goroutine.
+        return
+    }
 	isstart = true
 	go run()
 }
 
 func run() {
+
```
astaxie: please remove the empty line

toolbox/task.go
```diff
@@ -417,7 +422,7 @@ func run() {
 				if e.GetNext() != effective {
 					break
 				}
-				go e.Run()
+        go e.Run()
```
astaxie: please gofmt

astaxie:
@albertma please update your PR

albertma:
just gofmt, why fail in orm test, since I did not touch any code of this module. Does your unit test have any issue?

astaxie:
it looks like the network issue on travis.

httplib_test.go:60: Post http://httpbin.org/post: read tcp 10.128.0.44:46024->54.175.219.8:80: i/o timeout

PR #1924, opened 2016-05-05
implement some features

ysqi:
1. Validation Error implement Error Interface 
2. Change Comment Router file name logic

+  First

current  filename is too long ,such like as `commentsRouter_git.coding.net_ysqi_borrowCenter_apicontroller.go` , some time we need not a full path from $GOPATH , Only need relative path for project root. just short `commentsRouter_apicontroller.go`  

+ Second

the default is to generate router file in the root of project directory , but project directory structure is not  the same one all aways. such like as:
```root
├── conf 
├── main.go
├── src
│   ├── api
│   ├── controllers
│   ├── models
│   ├── routers
│   └── service
├── static 
├── tests 
└── views 
```

The default is to generate a file in the root directory , but i want generate file into `/src/routers/` , so i change logic : find routers dir  up from the same of controller file directory until the root of project directory.


astaxie:
> current filename is too long ,such like as commentsRouter_git.coding.net_ysqi_borrowCenter_apicontroller.go , some time we need not a full path from $GOPATH , Only need relative path for project root. just short commentsRouter_apicontroller.go


What happen for these structure?
```
├── conf
│   └── app.conf
├── controllers
│   ├── admin
│   │         ├── apicontroller.go
│   ├── user
│   │         ├── apicontroller.go
├── docs
│   ├── doc.go
│   └── docs.go
├── main.go
├── models
│   ├── object.go
│   └── user.go
├── routers
│   └── router.go
```

ysqi:
will create comment router file in root/routers ,
``` 
├── routers
│   └── commentsRouter_controllers_admin.go
│   └── commentsRouter_controllers_user.go
│   └── router.go

PR #1947, opened 2016-05-13
fix log module bugs

nullne:
the log file permission can not be set. for example in a config json, `"perm":0644` can not be parsed. so i changed it to `"perm": "0644"`,   add some helper code and it works.


astaxie:

PR #1948, opened 2016-05-14
fix bug with file permission in log module

nullne:
the log file permission can not be set. for example in a config json, "perm":0644 can not be parsed. so i changed it to "perm": "0644", add some helper code and it works.

### DIFF
--- a/logs/file.go
+++ b/logs/file.go
@@ -55,7 +55,6 @@ type fileLogWriter struct {
 	Level int `json:"level"`
 
 	Perm string `json:"perm"`
-	perm os.FileMode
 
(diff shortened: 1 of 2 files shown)


logs/file.go
```diff
@@ -53,7 +54,8 @@ type fileLogWriter struct {
 
 	Level int `json:"level"`
 
-	Perm os.FileMode `json:"perm"`
+	Perm string `json:"perm"`
+	perm os.FileMode
```
astaxie: Perm just change to string, but I don't think we need another param perm. just convert it to os.FileMode in `createLogFile`

nullne: @astaxie ok, seems a bit more brief

PR #1953, opened 2016-05-20
add ssdb session provider

lcbluestorm:
a ssdb provider for session


lcbluestorm:
Maybe the test file cause the problems, I have deleted the test file. Could you please check if it is ok?

PR #1955, opened 2016-05-21
Ctx.Redirect patch

gitchs:
[http.Redirect](https://golang.org/src/net/http/server.go?s=48569:48637#L1653) follow more RFCs than beego's implementation. So maybe we should use package "net/http"


astaxie:

PR #1966, opened 2016-05-27
modify the error log for registerModel.

wy65701436:
Tell user the default hard code PK is 'id' if not set.

As some one doesn't use id as primary key, here needs to tell the user the hard code value in beego. The log is more friendly for debugging, otherwise, has to go into details to go through the source code.


wy65701436:
@astaxie  how to avoid the CI error?

PR #1976, opened 2016-06-01
Fxied bug and Optimized code

ysqi:
1. Go format other commits.
2. Print complete URL address after running . this is useful for some shell tool, e.g iTerm .I can quick click the URL and visit it.
3. Fixed a bug for parse INI file with empty space line. 
4. Sport Error Exception for outside , we can use Exception and not use Abort()
5. Add a new function to find router info for context and clean the router code.

### DIFF
--- a/router.go
+++ b/router.go
@@ -603,6 +603,7 @@ func (p *ControllerRegister) ServeHTTP(rw http.ResponseWriter, r *http.Request) 
 		findRouter bool
 		runMethod  string
 		routerInfo *controllerInfo
+		isRunnable bool
 	)
 	context := p.pool.Get().(*beecontext.Context)
 	context.Reset(rw, r)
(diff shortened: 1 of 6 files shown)


ysqi:
Sport Error Exception for outside , we can use Exception and not use Abort() , such as like : 
```go
func (c *ContextExend) Forbidden(content string) {
	c.Input.SetData("content", content)
	beego.Exception(403, c.Context)
}
```

router.go
```diff
@@ -666,136 +667,127 @@ func (p *ControllerRegister) ServeHTTP(rw http.ResponseWriter, r *http.Request)
...
+			context.Input.SetParam(strconv.Itoa(k), v)
 		}
-
 	}
 
-	//if no matches to url, throw a not found exception
-	if !findRouter {
-		exception("404", context)
+	//store router pattern into context
+	context.Input.SetData("RouterPattern", routerInfo.pattern)
```
ysqi: PS:store router pattern into context

PR #1979, opened 2016-06-03
ignore case of tag and fixed bug for columName

ysqi:


orm/utils.go
```diff
@@ -181,18 +181,36 @@ func ToInt64(value interface{}) (d int64) {
 	return
 }
 
-// snake string, XxYy to xx_yy
+// snake string, XxYy to xx_yy , XxYY to xx_yy
```
astaxie: Why change the behaves `XxYY` to `xx_yy`, I think it should be `xx_y_y`

ysqi: - ID  ---> id
- NO --->no
- OrderID ----> order_id

more see test case.

astaxie: but our rules is all the capital will change to `_char`, I admire that would make more sense. but that change the behaves maybe will broken the exist database.

ysqi: I can't think anyone use that column name `n_o` or `i_d`  . like `URL` will set tag `column(URL)`

astaxie: @ysqi could you revert this change? it looks this break a lot old implement

ysqi: @astaxie  Are You sure?

astaxie: @ysqi yes. Here is the exmaple: `GComment` need to `g_comment`, but now it turn to `gc_omment`

Issue #1019, opened 2015-01-30
NullTime like NullString etc.


ptman:
Is there an easy way to use a NullTime type in Beego ORM? Or does support need to be added to the ORM? There is at least such types for most popular databases:  http://godoc.org/github.com/lib/pq#NullTime http://godoc.org/github.com/go-sql-driver/mysql#NullTime

astaxie:
how to use this? can you provide some example code.

ptman:
That's what I'm asking. I would like to use a NULLable timestamp in my schema, but Beego ORM makes it very tricky. Beego ORM supports NullString, NullBool, etc. But not NullTime. How would I use NullTime with Beego ORM?

ptman:
I'll have to test if simply specifying the nullable time column as *time.Time helps, but I would like for some NullTime type to exist.

ptman:
struct Foo {
        Id int
        Time *time.Time
    }

results in:
    
    panic: reflect: call of reflect.Value.Interface on zero Value

PR #2006, opened 2016-06-22
Add support for time.Time pointer in struct types

victorpopkov:
This allows using pointers `*time.Time` as a type in struct in combination with `orm` tags.
```go
type User struct {
	ID         int        `orm:"column(id);auto" json:"id"`
	Username   string     `orm:"column(username);size(255);unique" json:"username"`
	Email      string     `orm:"column(email);size(255);unique" json:"email"`
	CreatedAt  time.Time  `orm:"column(created_at);auto_now_add;type(datetime)" json:"created_at"`
	UpdatedAt  time.Time  `orm:"column(updated_at);auto_now;type(datetime)" json:"updated_at"`
	DeletedAt  *time.Time `orm:"column(deleted_at);type(datetime);null" json:"deleted_at,omitempty"`
}
```
This is useful when there is a need to treat `time.Time` fields as "empty" in JSON marshaling/unmarshaling using `json:"omitempty"` tag. Without pointer, `omitempty` tag will be ignored since it's not a `nil` value and will result in `0001-01-01T00:00:00Z` when unmarshaling.

There is a good answer on [StackOverflow](http://stackoverflow.com/a/32646035) about `omitempty` with `time.Time`.

The should solve https://github.com/astaxie/beego/issues/1019 where the issue was first mentioned.

### DIFF
--- a/orm/db.go
+++ b/orm/db.go
@@ -243,6 +243,9 @@ func (d *dbBase) collectFieldValue(mi *modelInfo, fi *fieldInfo, ind reflect.Val
 				if fi.isFielder {
 					f := field.Addr().Interface().(Fielder)
 					f.SetRaw(tnow.In(DefaultTimeLoc))
+				} else if field.Kind() == reflect.Ptr {
+					v := tnow.In(DefaultTimeLoc)
+					field.Set(reflect.ValueOf(&v))
 				} else {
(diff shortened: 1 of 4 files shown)


astaxie:
Cool

PR #2014, opened 2016-06-27
Changed how status is set.

GrimTheReaper:
Changed to ResponseWriter to match line 270. Doing this should allow headers to properly be written.


GrimTheReaper:
@JessonChan ```c.Ctx.Output.Status = status``` doesn't seem to set the status in the current version for me, which is why I'm suggesting this change. I suggest this change until you get the need for ResponseWriter completely removed and tested.

GrimTheReaper:
I made a PR to the wrong branch.

PR #2029, opened 2016-07-06
use keyed fields to pass go vet

ShevYan:
if use unkeyed fields to generate routers, these routers could not pass 'go vet'. Test results:
```
...
		beego.ControllerComments{
			Method: "Start",
			Router: `/start`,
			AllowHTTPMethods: []string{"post"},
			Params: nil})
...
```


astaxie:

PR #2062, opened 2016-07-24
减少没有必要的代码以及静态前缀前面有/

Maxgis:


config.go
```diff
@@ -233,9 +233,9 @@ func assignConfig(ac config.Configer) error {
 		sds := strings.Fields(sd)
 		for _, v := range sds {
 			if url2fsmap := strings.SplitN(v, ":", 2); len(url2fsmap) == 2 {
-				BConfig.WebConfig.StaticDir["/"+strings.TrimRight(url2fsmap[0], "/")] = url2fsmap[1]
+				BConfig.WebConfig.StaticDir["/"+strings.Trim(url2fsmap[0], "/")] = url2fsmap[1]
 			} else {
-				BConfig.WebConfig.StaticDir["/"+strings.TrimRight(url2fsmap[0], "/")] = url2fsmap[0]
+				BConfig.WebConfig.StaticDir["/"+strings.Trim(url2fsmap[0], "/")] = url2fsmap[0]
```
astaxie: please check the trim function. That's the wrong usage

Maxgis: The function is similar to TrimRight. I check the function is right. It was not considered the case that  the letter of the prefix is backslash

PR #2071, opened 2016-07-30
route variables should not have underline

Maxgis:


mmindenhall:
I just tried v1.7.0, and this commit broke things for me as I use `:snake_case` for all of my path variables.  I switched to `:camelCase` and all of my tests now pass.

It was not immediately apparent that this was the cause, so it might be good to call out as a possible breaking change in the 1.7.0 release notes.

Thanks!

astaxie:
@mmindenhall I am really sorry for that. I have revert this PR in develop. And will release next version soon.

PR #2084, opened 2016-08-05
update the default value for "CopyRequestBody"

xnsryx:
The default value for the "CopyRequestBody" in the development document is true, but the default value in the source code is false.
url:http://beego.me/docs/mvc/controller/config.md


astaxie:
@xnsryx thanks for point out the error on documents. I think the default value should be false. As most of case we don't need to Copy the request body. Appreciate that if you can create a pr to fix the documents.

Issue #1550, opened 2016-01-08
Using filter with splat (*) breaks docs router


pjoe:
When using a filter added with splat, docs router breaks, e.g.
```go
beego.InsertFilter("*", beego.BeforeStatic, RedirectHTTP)
```
From what I could see the issue is that ctx.Input.Params[":splat"] was set when matching for the filter and never cleared before running the docs router.

As a workaround I added this to my filter:
```go
// make sure :splat gets cleared in order to not confuse /docs
delete(ctx.Input.Params, ":splat")
```

pjoe:
This gets even worse in 1.6 where the workaround won't work as ctx.Input.Params is now an accessor function instead of having direct access to delete the param.

Think I will make a PR with something like ctx.Input.RemoveParam

pjoe:
Tried using:
```go
ctx.Input.SetParam(":splat", "")
```

But turns out SetParam won't overwrite existing params

pjoe:
Added PR to fix SetParam: https://github.com/astaxie/beego/pull/1637

The above workaround now works

danielscottt:
👋 

I'm running into this issue too, where my filter is using the `BeforeExec` state in the pipeline. We could clear params in router's `execFilter`, I think, so they were reset on context as it passes through each router.  Does that sound reasonable or is it bad form to change the state of input params as the context passes through the pipeline?

danielscottt:
Ok -- Disregard that above comment.

After further exploration, I see that this is likely a side effect of using a `BeforeExec` on a route that also uses splat, because it's the router which sets these params, and a filter which is at the `BeforeExec` stage in the pipeline is the last writer!  The router doesn't have a chance to reset params again.

I'm moving my filter to `BeforeRouter` for now.  If I end up needing session state down the road I'll revisit then.

Thanks!

PR #2085, opened 2016-08-05
adds ability to reset params after a filter runs

danielscottt:
When a filter is run _after_ the router completes, it's input params,
such as `":splat"` will have been overwritten by the filter's router pass.
This commit adds the ability to tell the router to revert to the previous input
params after running a filter.

### DIFF
--- a/context/input.go
+++ b/context/input.go
@@ -301,6 +301,14 @@ func (input *BeegoInput) SetParam(key, val string) {
 	input.pnames = append(input.pnames, key)
 }
 
+// ResetParams clears any of the input's Params
+// This function is used to clear parameters so they may be reset between filter
+// passes.
+func (input *BeegoInput) ResetParams() {
+	input.pnames = input.pnames[:0]
+	input.pvalues = input.pvalues[:0]
+}
+
 // Query returns input data item string by a given string.
 func (input *BeegoInput) Query(key string) string {
 	if val := input.Param(key); val != "" {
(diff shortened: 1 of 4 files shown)


router.go
```diff
@@ -581,12 +588,22 @@ func (p *ControllerRegister) geturl(t *Tree, url, controllName, methodName strin
...
 		if filterR.returnOnOutput && context.ResponseWriter.Started {
 			return true
 		}
+		if filterR.resetParams {
+			preFilterParams = context.Input.Params()
+		}
 		if ok := filterR.ValidRouter(urlPath, context); ok {
 			filterR.filterFunc(context)
+			if filterR.resetParams {
+				context.Input.Reset(context)
```
astaxie: `Input.Reset` has a log impact. while you only recover the params, but the Reset will reset all the data:

```
input.CruSession = nil
input.pnames = input.pnames[:0]
input.pvalues = input.pvalues[:0]
input.data = nil
input.RequestBody = []byte{}
```

danielscottt: Yeesh.  Yeah, good call.  I isolated just params.

danielscottt:
Tagging #1550.

This _could_ close this issue given an update to any `InsertFilter` call.

context/input.go
```diff
@@ -301,6 +301,12 @@ func (input *BeegoInput) SetParam(key, val string) {
 	input.pnames = append(input.pnames, key)
 }
 
+// ResetParams clears any of the input's Params
```
astaxie: Could you add more comments why add this function? That would be great to call back when refactor these function in future.

PR #2100, opened 2016-08-18
update swagger define

tnextday:
add array items define

reference: http://swagger.io/specification/#itemsType


tnextday:
I'm sorry, item should be Property,  not []Property, i will fix it later.

tnextday:
I have fixed it.

astaxie:
Cool, thanks

PR #2103, opened 2016-08-20
update session

Maxgis:
1. delete not necssary code

1.  fix bug, one session can be overwrited by another request at the same time

### DIFF
--- a/controller.go
+++ b/controller.go
@@ -28,7 +28,6 @@ import (
 	"strings"
 
 	"github.com/astaxie/beego/context"
-	"github.com/astaxie/beego/session"
 )
 
 //commonly used mime-types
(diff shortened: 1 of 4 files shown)


controller.go
```diff
@@ -499,46 +495,28 @@ func (c *Controller) SaveToFile(fromfile, tofile string) error {
...
-	return c.CruSession
-}
-
 // SetSession puts value into session.
 func (c *Controller) SetSession(name interface{}, value interface{}) {
-	if c.CruSession == nil {
-		c.StartSession()
-	}
-	c.CruSession.Set(name, value)
+	c.Ctx.Input.CruSession.Set(name, value)
```
astaxie: if c.Ctx.Input.CruSession is nil, it will cause panic here and above

astaxie:
@Maxgis any updated on it?

Maxgis:
Sorry, I will fixed it today.

Maxgis:
Only session is off , c.Ctx.Input.CruSession  will be nil. So I think it is not a bug.

astaxie:
@Maxgis no. We should never panic inside beego. As if the session is turn off, but you still call this function it should never panic

astaxie:
I will close this PR as you didn't have any update on it

Issue #2107, opened 2016-08-22
BeforeStatic filters have a nil session object


GrimTheReaper:
**1. What version of Go and beego are you using (`bee version`)?**
```
bee   :1.4.1
beego :1.6.1
Go    :go version go1.6.2 linux/amd64
```

**2. What operating system and processor architecture are you using (`go env`)?**
```
GOARCH="amd64"
GOBIN=""
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/user/go"
GORACE=""
GOROOT="/usr/local/go"
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GO15VENDOREXPERIMENT="1"
CC="gcc"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0"
CXX="g++"
CGO_ENABLED="1"
```

**3. What did you do?**
I went to the static file that I wanted to be user restricted.
(Code below)

**4. What did you expect to see?**
A redirection to `/login`

**5. What did you see instead?**
![alt text](http://i.imgur.com/3Dc8RHR.png?1 )

**Output Log:**
```
The context is null!
2016/08/22 14:21:37 [router.go:854][C] the request url is  /static/test/secrets.txt 
2016/08/22 14:21:37 [router.go:855][C] Handler crashed with error runtime error: invalid memory address or nil pointer dereference 
2016/08/22 14:21:37 [router.go:861][C] /usr/local/go/src/runtime/asm_amd64.s:472 
2016/08/22 14:21:37 [router.go:861][C] /usr/local/go/src/runtime/panic.go:443 
2016/08/22 14:21:37 [router.go:861][C] /usr/local/go/src/runtime/panic.go:62 
2016/08/22 14:21:37 [router.go:861][C] /usr/local/go/src/runtime/sigpanic_unix.go:24 
2016/08/22 14:21:37 [router.go:861][C] /home/user/go/src/~~~~~~~/routers/router.go:41 
2016/08/22 14:21:37 [router.go:861][C] /home/user/go/src/github.com/astaxie/beego/router.go:588 
2016/08/22 14:21:37 [router.go:861][C] /home/user/go/src/github.com/astaxie/beego/router.go:634 
2016/08/22 14:21:37 [router.go:861][C] /usr/local/go/src/net/http/server.go:2081 
[...]

PR #2108, opened 2016-08-22
Changed the location where Sessions are inited

GrimTheReaper:
Putting the session init here allows for static filters to access sessions. Resolves my previous issue, [here](https://github.com/astaxie/beego/issues/2107) where beego.BeforeStatic filters would panic if sessions were accessed.


GrimTheReaper:
@astaxie can I ask why this was turned down?

PR #2109, opened 2016-08-23
feature: add api controller error handler

ansiz:
1.add ControllerError for standard controller error
  2.add ApiAbort funtion for controller


astaxie:
I don't think this function should put into the controller.

PR #2132, opened 2016-09-01
beego.ParseForm() improvement

Zaaksam:
Support for anonymous struct

### DIFF
--- a/templatefunc.go
+++ b/templatefunc.go
@@ -280,15 +280,8 @@ func AssetsCSS(src string) template.HTML {
 }
 
 // ParseForm will parse form values to struct via tag.
-func ParseForm(form url.Values, obj interface{}) error {
-	objT := reflect.TypeOf(obj)
-	objV := reflect.ValueOf(obj)
-	if !isStructPtr(objT) {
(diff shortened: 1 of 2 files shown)


templatefunc.go
```diff
@@ -280,22 +280,23 @@ func AssetsCSS(src string) template.HTML {
...
-	objT := reflect.TypeOf(obj)
-	objV := reflect.ValueOf(obj)
-	if !isStructPtr(objT) {
-		return fmt.Errorf("%v must be  a struct pointer", obj)
-	}
-	objT = objT.Elem()
-	objV = objV.Elem()
-
+// Support for anonymous struct.
+func parseFormToStruct(form url.Values, objT reflect.Type, objV reflect.Value) error {
```
astaxie: why add a new function `parseFormToStruct`, it looks still the same logic for ParseForm. you just add a Anonymous

Zaaksam: it seems inelegant, i just want to support the multi-nested anonymous struct

astaxie:
@Zaaksam could you  add same test case for it?

Zaaksam:
Already added

astaxie:
awesome. thx

PR #2155, opened 2016-09-11
add feature: get the remaining time by cache key

App1905:
# Cache
i add a new function for what i want to get the remaining time of a cache.
`func (rc *Cache) GetRestDuration(key string) time.Duration`  
eg:

|key |value |expired time |
|:---- |:---- |:---- |
|bang.{phone} |{count} |{duration} |

sometime i need get the remaining time of a banged user who has tried {count} times ,while the system has a Maximum number of attempts.i think it`s more convenient if i can get the get the remaining time of the cache rather then store the remaining time in another cache.

in the code,i didn`t finish the implement of memcache or ssdb. so i return a zero:
```
func (rc *Cache) GetRestDuration(key string) time.Duration {
	return 0
}
```

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -49,6 +49,8 @@ import (
 type Cache interface {
 	// get cached value by key.
 	Get(key string) interface{}
+	// get rest duration.
+	GetRestDuration(key string) time.Duration
 	// GetMulti is a batch version of Get.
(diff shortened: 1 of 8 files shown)


astaxie:
It's not reasonable.

PR #2156, opened 2016-09-11
add client request ip

qAison:
1、access log add client request ip.
2、orm log delete repetition time.
3、const config dir and file.


utils/http.go
```diff
@@ -0,0 +1,23 @@
+package utils
+
+import (
+	"net/http"
+	"strings"
+)
+
+// 获取 客户端 IP
+func GetRequestIP(r *http.Request) string {
+	if ips := r.Header.Get("X-Forwarded-For"); ips != "" {
```
sergeylanzman: 👍

router.go
```diff
@@ -835,18 +835,19 @@ Admin:
 		statusColor := logs.ColorByStatus(iswin, statusCode)
 		methodColor := logs.ColorByMethod(iswin, r.Method)
 		resetColor := logs.ColorByMethod(iswin, "")
+		rip := utils.GetRequestIP(r)
```
astaxie: you can get the IP from `ctx.Input.IP()`. https://godoc.org/github.com/astaxie/beego/context#BeegoInput.IP

astaxie:
@qAison leave the comments please update it

utils/http.go
```diff
@@ -0,0 +1,23 @@
+package utils
+
+import (
+	"net/http"
+	"strings"
+)
+
+// 获取 客户端 IP
```
xiaopeng163: comments should be English I think.

astaxie: the problem is we don't need this. As it's already implement in context.Input

astaxie:
@qAison I have updated the code in develop. As you didn't update this PR as I suggested. I made these change.

1、access log add client request ip.
2、orm log delete repetition time.

 I will close this PR

PR #2171, opened 2016-09-20
Support of Foreign Keys for MySQL is added

vsdutka:


astaxie:
@vsdutka make sure the test case passed

Issue #2182, opened 2016-09-27
Emails are not sent when username ( email config ) is not a correct email


rahal:
Please answer these questions before submitting your issue. Thanks!
1. What version of Go and beego are you using (`bee version`)?

├── Beego     : 1.7.0
├── GoVersion : go1.7.1
1. What operating system and processor architecture are you using (`go env`)?

GOARCH="amd64"
GOHOSTOS="darwin"
GOOS="darwin"
1. What did you do?

my "username" provided for the mailer function is not a correct email.
1. What did you expect to see?

My email sent and no errors
1. What did you see instead?
   No email sent and I see this error in logs : "mail: missing phrase"

PR #2183, opened 2016-09-27
Emailer : Should use config Username only if no From is provided.

rahal:
Should also fix the case where Username is not an email.
( issue #2182  )


astaxie:

rahal:
done.

PR #2184, opened 2016-09-27
Emailer : Should use config Username only if no From is provided.

rahal:
Should fix the case where Username is not an email.


astaxie:
thanks

PR #2197, opened 2016-09-28
policies implementation

OlegFX:
Hi,

Please review this implementation of Policies logic. To test it add to test project file "policies/isAdmin.go" with similar content:

```
package policies

import (
    "fmt"
    "github.com/astaxie/beego/context"
)

func IsAdmin(ctx *context.Context) {
    fmt.Println("Admin policy")
}
```

Then add comment to any controller or method of controller:

`// @policy IsAdmin`

Thank you,
Oleg Demchenko

### DIFF
--- a/policy.go
+++ b/policy.go
@@ -1,0 +1,97 @@
+// Copyright 2016 beego authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package beego
+
+import (
+	"strings"
+
+	"github.com/astaxie/beego/context"
+)
+
+// PolicyFunc defines a policy function which is invoked before the controller handler is executed.
+type PolicyFunc func(*context.Context)
+
+// FindRouter Find Router info for URL
+func (p *ControllerRegister) FindPolicy(cont *context.Context) []PolicyFunc {
+	var urlPath = cont.Input.URL()
+	if !BConfig.RouterCaseSensitive {
+		urlPath = strings.ToLower(urlPath)
+	}
+	httpMethod := cont.Input.Method()
+	isWildcard := false
+	// Find policy for current method
+	t, ok := p.policies[httpMethod]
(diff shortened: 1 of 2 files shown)


astaxie:
@OlegFX what's the difference with the plugins. all the logic is the same. It's still the same as BeforeExec

OlegFX:
@astaxie currently if you want to add **IsAdmin** policy to a method of controller (for example, "POST"), you should add a BeforeExec filter for path to this controller and manually check HTTP Request Method for all incoming requests inside; if it's a "POST" - you should start IsAdmin logic execution. Another approach is to add one BeforeExec root filter for all policies and search for appropriate controller/method inside.

Both approaches can lead to excessive entanglement code that is critical for big projects.

With the new @policy logic you should just create **IsAdmin** policy and add @policy annotation to corresponding controller methods.

astaxie:
@OlegFX I think policy only append the method to special filter. Maybe we can improve current plugins to support method. what do you think?

OlegFX:
@astaxie Yes, InsertFilter for BeforeRouter with the method is similar to the policy with one "but".

Usually policy - is the list of policies for a method, so we also need to be able to add an arbitrary amount of policies to any method of controller. In addition we need to be able to add policies to the controller (for all methods) and in the case when we have some policy list for controller and for method of this controller - we should check all policies for current method, but skip controller's policies, if found any.

So, policy tree works as logic "OR", choosing just one most appropriate rule. This is different from filters, where we use "AND" logic, executing all appropriate filters.

And another great thing - to be able add @policy annotation to controller or method comments and get all required policy rules in auto-generated .go file.

In my PR all these things implemented.

OlegFX:
@astaxie So, as result - filters are good for common cases. For example, we have own implementation for Sessions. So we can add BeforeRouter filter where we can get user's session with all detailed information, and then we can move to the policies - find one rule for current API method call and check if current user has enough privileges to call this method.

astaxie:
I think it still the same as filter. Just add another router to the logic, that's redundant.

euknyaz:
@astaxie, I see that proposed policies implementation may be useful for better structuring the code and defining policies configuration as annotations. I would agree that it may look redundant counting that it's all just alteration of filters, but with pure filters there will be no separate well-defined layer of policies.

Lets look at SailsJS/nodejs framework policies implementation and documentation at http://sailsjs.org/documentation/concepts/policies
In this doc they stated "When it comes down to it, policies are really just Connect/Express middleware functions which run before your controllers.", so in their case they have extended middleware functions logic (filters in case of beego) and implemented policies on top of it for better structuring of the application. @OlegFX proposing to bind policies  to methods with annotations, which is the key aspect of implementation. This approach looks even better than what SailsJS proposing with policies config file.

Please think of it one more time. It looks like a good idea.

smacker:
looks interesting for me too. Because of imperfection of filters we had to build this layer inside controllers, and it's very wrong.

PR #2211, opened 2016-10-08
Default values

ihippik:


ihippik:
perform with:
https://github.com/beego/bee/pull/292

PR #2216, opened 2016-10-10
HTTPOnly Configurable

GrimTheReaper:
While most people will leave HTTPOnly turned on, we should be allowed to configure it, and turn it off if we need to. Very useful for single page JavaScript applications that are not being served by beego.

### DIFF
--- a/config.go
+++ b/config.go
@@ -94,7 +94,7 @@ type SessionConfig struct {
 	SessionCookieLifeTime   int
 	SessionAutoSetCookie    bool
 	SessionDomain           string
-	DisableHTTPOnly         bool // used to allow for cross domain cookies/javascript cookies.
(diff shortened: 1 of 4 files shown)


config.go
```diff
@@ -94,6 +94,7 @@ type SessionConfig struct {
 	SessionCookieLifeTime   int
 	SessionAutoSetCookie    bool
 	SessionDomain           string
+	DisableHTTPOnly         bool // used to allow for cross domain cookies/javascript cookies.
```
astaxie: Could we use the special name, like `SessionCookieHTTPOnly`

GrimTheReaper: Would SessionDisableHTTPOnly work? I think that the field should be Disable rather than Enable, because if something happens and the field isn't set, booleans default to false.

astaxie: LGTM

GrimTheReaper: Alright, making the change now.

PR #2228, opened 2016-10-17
Rework getFlatParams for time.Time

Hepri:
Fixed bug when all "time.Time" params in raw sql queries formatted as time. (e.g. "00:00:00"). This happens because raw sql params don't have fieldType for the arguments


Hepri:
The second commit not supposed to be is this pull request, but still important could you separate them into two pull requests?

Hepri:
@astaxie any news for merge?

PR #2248, opened 2016-10-28
Add RouterPattern to context.Input

smacker:
Right now beego adds this param only in dev mode, but I noticed that it's very useful to have in prod environment to.
My current use case - filter that sends logs in newrelic. Pattern there will help a lot to generate correct transaction name.

### DIFF
--- a/router.go
+++ b/router.go
@@ -712,10 +712,8 @@ func (p *ControllerRegister) ServeHTTP(rw http.ResponseWriter, r *http.Request) 
 	}
 
 	if routerInfo != nil {
-		if BConfig.RunMode == DEV {
-			//store router pattern into context
-			context.Input.SetData("RouterPattern", routerInfo.pattern)
-		}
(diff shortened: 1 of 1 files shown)


smacker:
actually I realized it's impossible to build general solution without RouterPattern :(

sergeylanzman:
I writer solution for newrelic and beego,  yes router pattern this good case.  @smacker  check my solution

smacker:
@sergeylanzman it doesn't work for us. Because we use UUID instead of integer ids. Of course I made a work around for our specific case, but it won't work in general case. Consider url pattern:
```
/api/entity/:uuid/products/:type(any string)
```

You can't replace `someWeirdStringFromDatabase` to `:type`. (okay, you can with a complicated logic based on `ctx.Input.Params()`, but better don't do like that)

sergeylanzman:
You are right, and this PR fixed this problem

PR #2265, opened 2016-11-07
Fix#2263

szyhf:
add "`" while create Sql of InsertOrUpdate


astaxie:
postgresql not support `

szyhf:
Oh, is any advice to use InsertOrUpdate in mysql and avoid this issue?

PR #2277, opened 2016-11-13
beego/session: return proper error when session is not found

mgenov:
Parent errs was returned instead of err which is returned from the last statement.


astaxie:
@mgenov please send the PR to develop branch

PR #2278, opened 2016-11-13
beego/session: return proper error when session is not found

mgenov:
Parent errs was returned instead of err which is returned from the last statement.


mgenov:
There is a previous commit which need to be fixed.

mgenov:
@astaxie PR is now on the development branch.

astaxie:
thanks

Issue #2317, opened 2016-12-08
AppPath regression


mlgd:
Please answer these questions before submitting your issue. Thanks!

1. What version of Go and beego are you using (`bee version`)?
go version go1.7.4 windows/amd64
Beego 1.7.2

2. What operating system and processor architecture are you using (`go env`)?
windows amd64

3. What did you do?
I try to read a config file stored in "conf" subfolder.

4. What did you expect to see?
No error

5. What did you see instead?
I see an error file not found because AppPath is incorrect when i launch app with the commande : go run
In this case, AppPath is set to the temp directory not the project's root folder
This commit https://github.com/astaxie/beego/commit/0a822209c8c1f356716ddf8072d5aa8e0f1ab377 generates this regresssion because it's set the current directory with the variable named AppPath that contain the temp folder

Someone can rollback this commit ?

PR #2319, opened 2016-12-09
Remove a regression on AppPath

mlgd:
The application path is incorrect on Windows with the command line "go run". AppPath is assigned to the temp directory instead the folder project


mlgd:
Issue https://github.com/astaxie/beego/issues/2317

PR #2323, opened 2016-12-13
replace kill() with Signal(syscall.SIGINT)

matthewgao:
Graceful use process.Kill() to kill the parent process, which will make the postSignalHandler can not been correctly called, because the main process was forced to exit.


astaxie:

PR #2326, opened 2016-12-14
Changing content type for template

kabab:
In the rendering of the template, the render function will overwrite the content type, this PR, will enable the dev to alter the content type even if he uses templating, as use case of this functionality, the ability to generate dynamic JS or CSS files.


astaxie:

kabab:
@astaxie done

PR #2339, opened 2016-12-22
enrich the funtion of toolbox.NewTask

zmw2004110:
if i want a task start another thread to do it's job after current job shutting down, current api cannot support my our feeds. so i suggest adding another param delayType, to make the work. if we do this, a task T can start a thread t1, and start another thread t2 after t1 it's done in period we set.


astaxie:
please send the pr to develop branch

PR #2342, opened 2016-12-23
replace kill() with Signal(syscall.SIGINT) 

matthewgao:
Kill signal will force the process exit, so it won't work in graceful mode

### DIFF
--- a/grace/server.go
+++ b/grace/server.go
@@ -65,7 +65,9 @@ func (srv *Server) ListenAndServe() (err error) {
 			log.Println(err)
 			return err
 		}
-		err = process.Kill()
+		//Do not use process.Kill() it will force the pprocess exit.
+		//which lead to all the post handler not work at all.
+		err = process.Signal(syscall.SIGINT)
 		if err != nil {
 			return err
 		}
@@ -120,7 +122,7 @@ func (srv *Server) ListenAndServeTLS(certFile, keyFile string) (err error) {
 			log.Println(err)
 			return err
 		}
-		err = process.Kill()
+		err = process.Signal(syscall.SIGINT)
 		if err != nil {
 			return err
 		}
--- a/grace/server.go
+++ b/grace/server.go
@@ -173,7 +173,8 @@ func (srv *Server) handleSignals() {
 	pid := syscall.Getpid()
 	for {
 		sig = <-srv.sigChan
-		srv.signalHooks(PreSignal, sig)
+		srv.wg.Add(1)
+		go srv.signalHooks(PreSignal, sig)
 		switch sig {
 		case syscall.SIGHUP:
 			log.Println(pid, "Received SIGHUP. forking.")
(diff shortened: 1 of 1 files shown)


grace/server.go
```diff
@@ -120,7 +123,7 @@ func (srv *Server) ListenAndServeTLS(certFile, keyFile string) (err error) {
 			log.Println(err)
 			return err
 		}
-		err = process.Kill()
+		err = process.Signal(syscall.SIGINT)
```
amrfaissal: Sending SIGINT is not supported on Windows.

grace/server.go
```diff
@@ -187,11 +191,15 @@ func (srv *Server) handleSignals() {
...
 			log.Printf("Received %v: nothing i care about...\n", sig)
 		}
-		srv.signalHooks(PostSignal, sig)
+		go srv.signalHooks(PostSignal, sig)
+		srv.wg.Done()
 	}
 }
 
 func (srv *Server) signalHooks(ppFlag int, sig os.Signal) {
+	srv.wg.Add(1)
```
amrfaissal: This should be outside of the function

grace/server.go
```diff
@@ -240,6 +248,7 @@ func (srv *Server) serverTimeout(d time.Duration) {
 			break
 		}
 		srv.wg.Done()
```
amrfaissal: Why you're still calling `srv.wg.Done()` when it's already deferred at the beginning.

grace/server.go
```diff
@@ -240,6 +248,7 @@ func (srv *Server) serverTimeout(d time.Duration) {
...
+		runtime.Gosched()
```
amrfaissal: We are already using a `WaitGroup` for that purpose.

grace/server.go
matthewgao: But in UNIX like OS Kill will force to kill the process, it won't wait for the parent gracefully exit, if INT is not support on windows then we can change it to another one.

grace/server.go
matthewgao: If we don't add the wait group, then we have a change which the process exit before any pre/post handler get called.

grace/server.go
matthewgao: I may not catch your point, the defer is used for handling a panic, not decrease a waitGroup

grace/server.go
matthewgao: It is a corner case, if we don't goshced here, then we have a change to loop again without srv.state been changed to StateTerminate, this race condition will make the waitGroup step == -1, that's a panic.

astaxie: which race condition? I want to know the root cause here while not use Gosched to ignore the race condition

matthewgao: StateTerminate is not protected by any mutex and not atomic, so the this for loop has a change not jump out, so the Done() will be called additionally, so the 'step' in waitGroup become -1

astaxie:
merged in another PR

PR #2343, opened 2016-12-24
Modify func camelString() to be more robust

legendtkl:
1. Strength function to cover more cases, such as
"pic_url_1" return "PicUrl1" (with the old version,
it will return "PicUrl_1").
2. More test cases is as follow:

"hello_world" -> "HelloWorld"
"HeLLo_world" -> "HeLLoWorld"
"hello_world__" -> "HelloWorld"
"hello__world" -> "HelloWorld"
"hello_world_1" -> "HelloWorld1"

PR #2346, opened 2016-12-25
Modify func camelString() to be more robust

legendtkl:
Hello, in the previous pull request #2343 , I mix the history commits,  so I close it, and make a new one. Sorry for that.

Here is the points.
1. In previous edition, for case "pic_url_1", the func camelString() will return
"PicUrl_1", but "PicUrl1" seems to be more reasonable. So I make this change.
2. For my commit, test cases are as follow:
* "pic_url" -> "PicUrl" 
* "hello_world_" -> "HelloWorld"
* "hello__World" -> "HelloWorld"
* "_HelLO_Word" -> "HelLOWord"
* "pic_url_1" -> "PicUrl1"
* "pic_url__1" -> "PicUrl1"

about the test, please refer to file utils_test.go

Thanks.

### DIFF
--- a/orm/utils.go
+++ b/orm/utils.go
@@ -219,22 +219,17 @@ func snakeString(s string) string {
 // camel string, xx_yy to XxYy
 func camelString(s string) string {
 	data := make([]byte, 0, len(s))
-	j := false
-	k := false
-	num := len(s) - 1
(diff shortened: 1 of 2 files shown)


astaxie:
👍

PR #2358, opened 2016-12-31
EnableGZip=true,then content-length header missing

kbynd:
This results in responses with Content-Type as gzip as opposed to original content type.
This affects ServeJSON() function.

Also content-type is set wrong in this case , resulting in loss of functionality to gzip encode JSON output.


astaxie:
@kbynd please send the pull request to develop branch.

PR #2365, opened 2017-01-06
[WIP]Have Required validator trim strings to fix #2361

### LINKED ISSUE #2361: Space character can be used to get passed `Required` validation
#### bee version
Beego     : 1.7.2
GoVersion : go1.7.3
GOOS      : linux
GOARCH    : amd64
NumCPU    : 2
GOPATH    : /home/chesedo/go-code
GOROOT    : /home/chesedo/go
Compiler  : gc
Date      : Wednesday, 4 Jan 2017

#### go env
GOARCH="amd64"
GOBIN=""
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/chesedo/go-code"
GORACE=""
GOROOT="/home/chesedo/go"
GOTOOLDIR="/home/chesedo/go/pkg/tool/linux_amd64"
CC="gcc"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build049135941=/tmp/go-build -gno-record-gcc-switches"
CXX="g++"
CGO_ENABLED="1"

#### What I did
When one creates a validation element with the `Required` validation it will pass when the value is only spaces. Eg:
```go
type Test struct {
	Field string `form:"Required"`
}

func main() {
	// Create test with empty `Field`
	test := Test{}

	v := valid.Required(test.Field, "field")
	fmt.Printf("Validation has error '%s' as expected for missing value\n", v.Error.Message)

	// Make `Field` a space value
	test.Field = " "
	v = valid.Required(test.Field, "field")
	fmt.Printf("Validation has error '%v' which is unexpected for space value\n", v.Error)

	// Output:
	// Validation has error 'Can not be empty' as expected for missing value
	// Validation has error '' which is unexpected for space value
}
```

#### Expected
I expected the validation to also fail when the value was just spaces

#### Results
When the value was just spaces it passed the validation

If this is a bug then I would be happy to submit a PR.

chesedo:
This will cause the Required validator not to consider fields that has
only spaces or new lines to be regarded as valid. This is done by
checking if the trimmed version of the string is valid.

### DIFF
--- a/validation/validation_test.go
+++ b/validation/validation_test.go
@@ -35,6 +35,12 @@ func TestRequired(t *testing.T) {
 	if valid.Required("", "string").Ok {
 		t.Error("\"'\" string should be false")
 	}
+	if valid.Required(" ", "string").Ok {
+		t.Error("\" \" string should be false") // For #2361
+	}
+	if valid.Required("\n", "string").Ok {
+		t.Error("new line string should be false") // For #2361
+	}
 	if !valid.Required("astaxie", "string").Ok {
 		t.Error("string should be true")
 	}
--- a/validation/validators.go
+++ b/validation/validators.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"reflect"
 	"regexp"
+	"strings"
 	"time"
 	"unicode/utf8"
 )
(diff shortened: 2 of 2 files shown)


chesedo:
Check my next [PR](2367) where I highlighted that this is also possible in `MaxSize`, `MinSize` and `Length`. That solution will fix it for all validators where this is possible, but it only works for structs called by reference on the `validator.Valid()` method.

This one will work for all validators called directly (eg. `validator.Required()`), but will still leave the field 'dirty' and it would probably go into the DB with the padding spaces.

So I would like advise on which solution you would prefer (or both maybe)? If this one will be used, then I would like to also add it for the validators mentioned above before it is merged.

PR #2367, opened 2017-01-06
Add trim ability in validator.Valid() for all string fields

chesedo:
This will allow all string fields for structs that are passed by
reference to validator.Valid() to have their value trimmed
automatically.

This will prevent a bypass on all validators (like Length,
MinSize and Maxsize) by just adding enough spaces. And it will help
to prevent DBs from becoming polluted with padding spaces.

### DIFF
--- a/validation/validation.go
+++ b/validation/validation.go
@@ -321,6 +321,15 @@ func (v *Validation) Valid(obj interface{}) (b bool, err error) {
 
 	for i := 0; i < objT.NumField(); i++ {
 		var vfs []ValidFunc
+		field := objV.Field(i)
+		// Trim spaces from string fields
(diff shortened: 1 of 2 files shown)


astaxie:
I don't think trim the string in validation is a good idea. validation should check the original data. If the validate need to valid the empty. How can we do in future?

chesedo:
@astaxie not doing it validation makes sense - I never though of that. But where else would you suggest it happens (maybe when the POST/GET data is parsed)?

Note that the way it currently is, an empty will still validate as normal since the value is only changed when the trim is successful.

astaxie:
@chesedo I think maybe we need to add a new validation for trim. Or before pass the struct to validation user need to handle the values.

PR #2371, opened 2017-01-09
beego/config:remove temp file after parse done which ParseData created

fugr:
temp files should removed after works done.

### DIFF
--- a/config/ini.go
+++ b/config/ini.go
@@ -184,6 +184,8 @@ func (ini *IniConfig) ParseData(data []byte) (Configer, error) {
 	if err := ioutil.WriteFile(tmpName, data, 0655); err != nil {
 		return nil, err
 	}
+	defer os.Remove(tmpName)
+
 	return ini.Parse(tmpName)
 }
 
--- a/config/xml/xml.go
+++ b/config/xml/xml.go
@@ -81,6 +81,8 @@ func (xc *Config) ParseData(data []byte) (config.Configer, error) {
 	if err := ioutil.WriteFile(tmpName, data, 0655); err != nil {
 		return nil, err
 	}
+	defer os.Remove(tmpName)
+
 	return xc.Parse(tmpName)
 }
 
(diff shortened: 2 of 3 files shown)


config/ini.go
```diff
@@ -184,6 +184,8 @@ func (ini *IniConfig) ParseData(data []byte) (Configer, error) {
 	if err := ioutil.WriteFile(tmpName, data, 0655); err != nil {
 		return nil, err
 	}
+	defer os.Remove(tmpName)
```
amrfaissal: @fugr I think it would be better if you handle the returned `error` from `os.Remove(tmpName)`:
```
defer func() {
     if err := os.Remove(tmpName); err != nil {
            return nil, err
     }
}()
```

config/xml/xml.go
```diff
@@ -81,6 +81,8 @@ func (xc *Config) ParseData(data []byte) (config.Configer, error) {
 	if err := ioutil.WriteFile(tmpName, data, 0655); err != nil {
 		return nil, err
 	}
+	defer os.Remove(tmpName)
```
amrfaissal: Same as here.

config/yaml/yaml.go
```diff
@@ -69,6 +69,8 @@ func (yaml *Config) ParseData(data []byte) (config.Configer, error) {
 	if err := ioutil.WriteFile(tmpName, data, 0655); err != nil {
 		return nil, err
 	}
+	defer os.Remove(tmpName)
```
amrfaissal: and here.

fugr:
i think the value of os.Remove(tmpName) does not affect ParseData(data []byte) (Configer, error)

amrfaissal:
Well, it's still considered an error that needs to be reported to the caller. As you mentioned, writing `os.Remove()` without error handling will pass silently but the end user will end up (soon or later) with a full `/tmp` folder which might cause the application to crash.

fugr:
@amrfaissal it's unnecessary to do that.@astaxie what you think?

fugr:
opened another PR #2373 ,so this is closed.

PR #2373, opened 2017-01-09
avoid creating new file to implements Config

fugr:
There is no need to create new file in ParseData(data []byte) (Configer, error).
let's make code simply.

when handle include "other.conf", the otherfile is either absolute dir or in current directory.


fugr:
@amrfaissal will you review this?

amrfaissal:
@fugr Nice!

PR #2389, opened 2017-01-13
Package env for working with environment variables inside Beego

amrfaissal:
The package env makes it trivial to work with environment variables.
It allows getting values with defaults as a fallback.
New ENV variables can be set safely on the current process environment.


astaxie:
maybe this package can move into config?

amrfaissal:
Good idea!

config/env/env.go
```diff
@@ -0,0 +1,89 @@
...
+var env struct {
+	data map[string]string
+	lock *sync.RWMutex
+}
+
+func init() {
+	env.data = make(map[string]string)
+	env.lock = &sync.RWMutex{}
+	for _, e := range os.Environ() {
+		splits := strings.Split(e, "=")
```
astaxie: I would like to check the splits's len here, it maybe cause panic if access the index 0

config/env/env.go
```diff
@@ -0,0 +1,89 @@
...
+
+	if val, ok := env.data[key]; ok {
+		return val
+	}
+	return defVal
+}
+
+// MustGet returns a value by key.
+// If the key does not exist, it will return an error.
+func MustGet(key string) (string, error) {
```
astaxie: it looks mustGet and Get need to change the params. like https://golang.org/pkg/html/template/#Must

it will never return error.

config/env/env.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	env.data[key] = value
+	return nil
+}
+
+// GetAll returns all keys/values in the current child process environment.
+func GetAll() map[string]string {
+	env.lock.RLock()
+	defer env.lock.RUnlock()
+	return env.data
```
astaxie: I think here need to return the env.data's copy, otherwise will cause the cocurrent access map.

config/env/env.go
amrfaissal: I don't think it's necessary. `os.Envrion()` always returns elements in the form `key=value`.
https://golang.org/src/syscall/env_unix.go

config/env/env.go
amrfaissal: I am using `Must*()` just to tell the user that this function **will return an error** if the element does not exist. This is an example from `env_test.go`:
```
	_, err = MustGet("NOEXISTVAR")
	if err == nil {
		t.Error("expected error to be non-nil")
	}
```

config/env/env.go
amrfaissal: I think I'll use `BeeMap` in `safemap.go` instead of regular `map[string]string` as the thread-safety is already implemented. What do you think?

config/env/env.go
astaxie: I understand what you wanted, but in Go convention,`Must` always return no error.  I think we also need to follow the convention.

config/env/env.go
astaxie: yea, but you need to convert interface to string. otherwise it's not convenience.

config/env/env.go
astaxie: make sense

config/env/env.go
amrfaissal: Yes I agree. I am using the same signature as the one in `regexp` package:
```
func MustCompile(str string) *Regexp {}
```
AFAIK, `Must()` and `MustCompile()` panic in case of an error. So instead of panicking (which is bad), the error is returned which gives more control to the developer.

PR #2399, opened 2017-01-18
Add GetCond func to querySet

kerwin:
Export raw cond object on querySet allow more enhanced query filter.


amrfaissal:
@kerwin Can you please send the request to `develop` branch? Thanks. https://github.com/astaxie/beego/blob/master/CONTRIBUTING.md

PR #2402, opened 2017-01-19
support set Asset func for the template

uptoknow:
Resolve this issue https://github.com/beego/bee/issues/310

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .idea
+.vscode
 .DS_Store
 *.swp
 *.swo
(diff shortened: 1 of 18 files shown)


uptoknow:
This is the beego part for go-bindata support, we need to enable set the Asset func for retrieve the template data, when we use go-bindata, this will generate the a func, if we use it, we can set the func for the template, will prepare a pr for the bee tool to enable go-bindata support when I have time, thanks.

amrfaissal:
Hi @uptoknow. Can you please send the pull request to `develop` branch? Thanks.

PR #2406, opened 2017-01-21
fix the bug to prevent rewrite t

xqbumu:
the t have parsed in line 212,and t is a point , and it will rewrite in the getTplDeep function, so  the origin date will be removed.
the t.New(file) operate in line 212 product a new template from t and extend some information about t.
if you debug it when the tpl file has line like 
{{ template "" .}}
the file info of parent tpl  will become of the file above of.

### DIFF
--- a/template.go
+++ b/template.go
@@ -224,7 +224,7 @@ func getTplDeep(root, file, parent string, t *template.Template) (*template.Temp
 			if !HasTemplateExt(m[1]) {
 				continue
 			}
-			t, _, err = getTplDeep(root, m[1], file, t)
+			_, _, err = getTplDeep(root, m[1], file, t)
 			if err != nil {
 				return nil, [][]string{}, err
 			}


astaxie:
@xqbumu I can't get the point.

> the file info of parent tpl will become of the file above of.

what's this mean?

xqbumu:
@astaxie 在getTplDeep方法中，t已经在[https://github.com/xqbumu/beego/blob/e32d173b0d43c3987d0f2906467f999be5ba4d7a/template.go#L212](url)生成了，并传入了相应的tpl文件，而在[https://github.com/xqbumu/beego/blob/e32d173b0d43c3987d0f2906467f999be5ba4d7a/template.go#L227](url)又被新的指针替换，那么最后得到的是子视图里解析后的tpl，而父级视图文件也就不存在了，所以需要在解析子视图时，不用再对t指针重新赋值。go 自带的template引擎会自动将传入的子tpl归属到其父tpl中。

astaxie:
@xqbumu 谢谢详细的解释

PR #2409, opened 2017-01-28
make ViewsPath options for controller

sergeylanzman:
If you need set controller with views from package(module system). Today this very hard, because ViewsPath this global param in all application. Path to view from package this vendor/../../.. and not views. Need set this in controller.

### DIFF
--- a/controller.go
+++ b/controller.go
@@ -69,6 +69,7 @@ type Controller struct {
 
 	// template data
 	TplName        string
+	ViewsPath      string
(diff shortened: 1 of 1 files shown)


astaxie:
This is a good proposal. But I have concern that the vendor also need to deploy to production. Not like the binary with view folder.

sergeylanzman:
better solution here https://github.com/astaxie/beego/pull/2417

PR #2417, opened 2017-02-07
support for multiple view paths

eyalpost:
Added support for multiple view paths in addition to BConfig.WebConfig.ViewsPath
To use another path, call beego.AddViewPath() with the additional path and set the controller.ViewPath accordingly

### DIFF
--- a/template.go
+++ b/template.go
@@ -32,7 +32,8 @@ import (
 
 var (
 	beegoTplFuncMap = make(template.FuncMap)
-	// beeViewPathTemplates caching map and supported template file extensions.
+	beeViewPathTemplateLocked = false
+	// beeViewPathTemplates caching map and supported template file extensions per view
 	beeViewPathTemplates  = make(map[string]map[string]*template.Template)
 	templatesLock sync.RWMutex
 	// beeTemplateExt stores the template extension which will build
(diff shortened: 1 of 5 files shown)


template.go
```diff
@@ -149,6 +156,11 @@ func AddTemplateExt(ext string) {
 	beeTemplateExt = append(beeTemplateExt, ext)
 }
 
+func AddViewPath(viewPath string) error {
+	beeViewPathTemplates[viewPath] = make(map[string]*template.Template)
```
astaxie: As `AddViewPath` is  a public function, is it possible the user add the viewPath during the application running? That would cause map panic.

sergeylanzman: @astaxie 
Can add mutex lock/unlock in this case. WDYT?

eyalpost:
I'll add a flag to disable calls to AddVieePath after init. Sounds ok?

Eyal

On 10 Feb 2017, at 18:42, Sergey Lanzman <notifications@github.com> wrote:

*@sergeylanzman* commented on this pull request.
------------------------------

In template.go <https://github.com/astaxie/beego/pull/2417>:

> @@ -149,6 +156,11 @@ func AddTemplateExt(ext string) {
 	beeTemplateExt = append(beeTemplateExt, ext)
 }

+func AddViewPath(viewPath string) error {
+	beeViewPathTemplates[viewPath] = make(map[string]*template.Template)

@astaxie <https://github.com/astaxie>
Can add mutex lock/unlock in this case. WDYT?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/astaxie/beego/pull/2417>, or mute the thread

template.go
astaxie: It's a tradeoff. If we add mutex here all the template render will add lock. I think maybe we just add the documents for the caller.

astaxie: most of the case it will not happen during the application running.

eyalpost: I added a flag to deny access to AddViewPath after app is running

PR #2422, opened 2017-02-09
Add config field EnableErrorsRender

ansiz:
Add config field EnableErrorsRender to disable errors output
  with template data, sometimes we do not want errors output with
  template data even in dev mode, especially in API projects.


astaxie:
please send pr to develop branch

PR #2423, opened 2017-02-09
Add config field EnableErrorsRender

ansiz:
Add config field EnableErrorsRender to disable errors output
with template data, sometimes we do not want errors output with
template data even in dev mode, especially in API projects.


astaxie:
@ansiz please also update the beego/beedoc for the new config

ansiz:
@astaxie Please review: https://github.com/beego/beedoc/pull/558

astaxie:
@ansiz thanks

PR #2466, opened 2017-02-27
Parse form time by its length

### LINKED ISSUE #2451: Beego cannot parse time "2006-01-02"
I see this issue in my web app, beego cannot parse the time format "2006-01-02", but beego/orm can work.

Beego is only support the format `time.RFC3339`, in
https://github.com/astaxie/beego/blob/master/templatefunc.go#L355

The beego/orm is handling it better. 
https://github.com/astaxie/beego/blob/master/orm/db.go#L1263

gouyang:
Fix #2451

Signed-off-by: Guohua Ouyang <guohuaouyang@gmail.com>

### DIFF
--- a/templatefunc.go
+++ b/templatefunc.go
@@ -26,6 +26,12 @@ import (
 	"time"
 )
 
+const (
+	formatTime     = "15:04:05"
+	formatDate     = "2006-01-02"
+	formatDateTime = "2006-01-02 15:04:05"
+)
+
 // Substr returns the substr from start to length.
 func Substr(s string, start, length int) string {
 	bt := []rune(s)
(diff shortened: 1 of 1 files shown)


templatefunc.go
```diff
@@ -352,11 +358,25 @@ func parseFormToStruct(form url.Values, objT reflect.Type, objV reflect.Value) e
...
-				format := time.RFC3339
-				if len(tags) > 1 {
-					format = tags[1]
+				var (
+					t   time.Time
+					err error
+				)
+				if len(value) >= 19 {
+					value = value[:19]
+					t, err = time.ParseInLocation(formatDateTime, value, time.Local)
```
astaxie: how about `2006-01-02T15:04:05Z07:00`? if the string with the timezone, this Parse would be error.

gouyang: Maybe I take this too simple, You can see I was copy the codes from beego/orm. 
If it's only parsing time from html form, I think the three format would be enough.

astaxie: this code dup three as below. I think you still can put it back `t, err = time.ParseInLocation(formatDateTime, value, time.Local)`

gouyang: I don't catch your point much, are you saying the line you mentioned are duplicated?  It's handling three different time format, not dup. And are you meaning to include time format `2006-01-02T15:04:05Z07:00` as well?

astaxie: > I don't catch your point much, are you saying the line you mentioned are duplicated? It's handling three different time format, not dup.

Ah, you are right. I am sorry for it. It's really a bit confused for the code.




> And are you meaning to include time format 2006-01-02T15:04:05Z07:00 as well?

Yea, Could you also support this format?

PR #2491, opened 2017-03-10
Fix routing bug for splat matching

chendx79:
Bug fix. When match url `/http://customer/123/` to router `/*`，param splat will be `http:/customer/123` instead of `http://customer/123/` because of wildcards splitting and rejoining.


chendx79:
it should be /url/*

PR #2512, opened 2017-03-17
add unconverted support

sergeylanzman:


sergeylanzman:
unconverted not support go1.4 => https://github.com/astaxie/beego/pull/2511

PR #2536, opened 2017-03-28
Fix for IndexExists in SQLite driver

ChristophPech:
they added the "origin" and "partial" columns to the index_list pragma.

see:  https://www.sqlite.org/src/info/2743846cdba572f6


ChristophPech:
See the documentation: https://sqlite.org/pragma.html#pragma_index_list
otherwise I'll get a "sql: expected 5 destination arguments in Scan, not 3" from rows.Scan()

PR #2586, opened 2017-04-22
Automatic Parameter Router

eyalpost:
This is a POC for a feature i've been working on that I'd like to get reviewed before I complete the work (to make sure the direction is ok)
The idea is to make Comments Controller even more "magical" by translating request parameters to method parameters. 

[See updated information below]

Let me know if the direction looks ok and I will complete the missing parts. Also if you think this is good, please review the changes.
Note that most changes are in new files and there should be no changes to current flows. Changes in existing files are minimal.


astaxie:
That is a good ideas. Could you move the params and response to the context package? I think context packgage is used for that, And also I think we already have parse params to type there

eyalpost:
@astaxie I can move it under context but will it be ok if I keep it as sub folders of context? 
The reason is that it makes code shorter and more readable this way e.g.:
param.Bool vs context.BoolParam 
or
response.StatusCode(400) vs context.ResponseStatusCode(400)
What do you think?

Also, Can I delete Params map from ControllerComments struct? It doesn't seem to be used anywhere

eyalpost:
@astaxie This is now ready for review. It has an accompanying in bee: https://github.com/beego/bee/pull/418 which also add automatic swagger generation for method parameters.

What does it do?
- Add parameters to your controller method and the router will convert request parameter (query, form ,body or header) and pass to your function. This works in coordination with the @param comment so the router knows where to look for the parameter
- if your function has a result, it will automatically be converted to json and returned in the http response. If you have an `error` result, it will be translated to an http error (example below)
- if parameters are declared as pointers, you will get null if they don't appear in the request. unless if @param defines a parameter as required: then an error will occur if it's missing
- all built in types are supported (well.. except Complex)
- struct types are marshalled as json (except time.Time which is converted to\from RFC3339 string)
- arrays in body are treated as json. in query or formData: as comma separated list


End result: 
It allows you to convert code like this:
```
// @Param	id		path 	string		true	"The id you want to update"
// @Param	task	body 	models.Task	true	"body for Task content"
// @Success 200 {object} Status
// @router /task/:id [put]
func (c *TaskController) PutTaskById() {
	id, err := c.GetInt64(":id")
	if err != nil {
		c.Ctx.Output.SetStatus(400)
		c.Data["json"] = err.Error()
		c.ServeJSON()
		return
	}

	v := models.Task{}
	if err := json.Unmarshal(c.Ctx.Input.RequestBody, &v); err != nil {
		c.Ctx.Output.SetStatus(400)
		c.Data["json"] = Status{Status: err.Error()}
		c.ServeJSON()
		return
	}
	if err := models.UpdateTask(id, &v); err == nil {
		c.Data["json"] = Status{Status: "OK"}
	} else {
		c.Ctx.Output.SetStatus(400)
		c.Data["json"] = Status{Status: err.Error()}
	}
	c.ServeJSON()
}
```
To this:
```
// @Param	id		path 	string		true	"The id you want to update"
// @Param	task	body 	models.Task	true	"body for Task content"
// @Success 200 {object} Status
// @router /task/:id [put]
func (c *TaskController) PutTaskById(id int64, task *models.Task) (*Status, error) {
	if err := models.UpdateTask(id, task); err != nil {
		return nil, err
	}
	return &Status{Status: "OK"}, nil
}
```
Note how all of the boilerplate code is gone and all that is left is pure business logic

context/param/conv.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	"fmt"
+	"reflect"
+
+	beecontext "github.com/astaxie/beego/context"
+	"github.com/astaxie/beego/logs"
+)
+
+func ConvertParams(methodParams []*MethodParam, methodType reflect.Type, ctx *beecontext.Context) (result []reflect.Value) {
```
astaxie: please add comments

context/param/methodparams.go
```diff
@@ -0,0 +1,67 @@
+package param
+
+import (
+	"fmt"
+	"strings"
+)
+
+//Keeps param information to be auto passed to controller methods
+type MethodParam struct {
```
astaxie: golints need start the comments with `MethodParam`

astaxie:
@eyalpost This is a great feature. But I will close it as I add golint and ineffassign tool to check the code in the CI. please fix the comments and then send a new PR. Thanks

eyalpost:
Can you leave this PR open? I will do the fixes and it will run the build again automatically

astaxie:
@eyalpost yea, I can try. But it looks will not automaticlly use the latest travis config. let's try

eyalpost:
I can sync my branch so I will have the latest travis

astaxie:
@eyalpost thannks

eyalpost:
@astaxie i've made the golint fixes and updates with the latest travis. build passes

context/response/responses.go
```diff
@@ -0,0 +1,53 @@
+package response
+
+import (
+	beecontext "github.com/astaxie/beego/context"
+)
+
+// JSON renders value to the response as JSON
+func JSON(value interface{}, encoding ...bool) Renderer {
```
astaxie: It's already have context response, why create a new package response here. That's really confuse

eyalpost: Where there is already package response?

eyalpost:
@astaxie what are the next steps here? Would love to see it as part of beego! :)

context/response/responses.go
eyalpost: Maybe rename to httpResponse?

context/param/conv.go
```diff
@@ -0,0 +1,81 @@
...
+
+func getParamValue(param *MethodParam, ctx *beecontext.Context) string {
+	switch param.location {
+	case body:
+		return string(ctx.Input.RequestBody)
+	case header:
+		return ctx.Input.Header(param.name)
+		// if strValue == "" && strings.Contains(param.name, "_") { //magically handle X-Headers?
+		// 	strValue = ctx.Input.Header(strings.Replace(param.name, "_", "-", -1))
+		// }
```
astaxie: please remove these comments code.

context/param/methodparams.go
```diff
@@ -0,0 +1,69 @@
...
+	"fmt"
+	"strings"
+)
+
+//MethodParam keeps param information to be auto passed to controller methods
+type MethodParam struct {
+	name     string
+	location paramLocation
+	required bool
+	defValue string
```
astaxie: please use `defaultVal` or `defaultValue` while not `defValue`.

context/param/methodparams.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+//MethodParam keeps param information to be auto passed to controller methods
+type MethodParam struct {
+	name     string
+	location paramLocation
```
astaxie: http://swagger.io/specification/#parameter-object-44

I would like use the type here. The same as swagger defined. what do you think?

context/param/parsers.go
```diff
@@ -0,0 +1,149 @@
...
+	"strconv"
+	"strings"
+	"time"
+)
+
+type paramParser interface {
+	parse(value string, toType reflect.Type) (interface{}, error)
+}
+
+func getParser(param *MethodParam, t reflect.Type) paramParser {
```
astaxie: If can add some test for this function would be awesome. I am not confident about this for all type

context/response/responses.go
astaxie: As context.Output already have a lot functions based on response. I think the seperate response package does not make sense here.

eyalpost: This is mainly for code readability
The idea is that a controller method can be written like this:
```
func(c *TaskController) PutTaskById(id int64, task *models.Task) (*Status, error) {
	if /*not found */ {
		return nil, response.StatusCode(404)
	} else if /* needs redirect */ {
		return nil, response.Redirect("/somePath")
        } 

	return &Status{Status: "OK"}, nil
}
```
What do you think?

context/param/methodparams.go
eyalpost: You mean like this?

type MethodParam struct {
	name         string
	**type         paramType**
	required     bool
	defaultValue string
}

astaxie: yea, does that works?

eyalpost: Don't know.. maybe it will confuse that this is the go type of the param (like int, float etc.)? 
Also I see swagger ref you linked also treats this as "location". 
I'm OK with both options: What do you decide?

context/param/parsers.go
eyalpost: Definitely! Added now and will expand later with more tests

context/param/methodparams.go
astaxie: swagger use `in` for this type. We can keep the same as swagger.

```
"parameters": [
{
"name": "petId",
"in": "path",
"description": "ID of pet to return",
"required": true,
"type": "integer",
"format": "int64"
}
],
```

```
type MethodParam struct {
name string
in paramType
required bool
defaultValue string
}
```

eyalpost: done

context/param/methodparams.go
eyalpost: done

eyalpost:
All comments addressed. Let's decide what to do with response package? 
My last comment: https://github.com/astaxie/beego/pull/2586#discussion_r116009391

context/response/responses.go
astaxie: Yea, I understand the readability. But why not just use the ctx.Output

```
func(c *TaskController) PutTaskById(id int64, t *Task) (*Status, error) {
	if /*not found */ {
		return nil, c.Ctx.StatusNotFound
	} else if /* needs redirect */ {
		return nil, c.Ctx.Output.Redirect("/somePath")
        } 

	return &Status{Status: "OK"}, nil
}
```

eyalpost: StatusNotFound is simply a const - I don't think it should be placed on the context object.
Redirect returns a renderer. It does not changes directly the output - so I don't think it belongs on the Output object itself. also c.Ctx.Output.Redirect feels to verbose IMO

eyalpost: Also, response.JSON can not be put on Output because Output already has JSON. and repsonse JSON again returns a renderer which eventually calls Output.JSON. 
The point of renderers what that writing to output is always done after the function is finished and not within the code.

franzwilhelm:
@eyalpost Fantastic work here! Would love to use this in my projects!

eyalpost:
Will sure do!

context/response/responses.go
astaxie: > c.Ctx.Output.Redirect feels to verbose IMO

When I design the Ctx, we just name request as input and response as output. we need to keep consistency with the design. 

> Also, response.JSON can not be put on Output because Output already has JSON. and repsonse JSON again returns a renderer which eventually calls Output.JSON.

If you return the render from the JSON, I would like to rename it to `JSONRenderer`

eyalpost: OK, I refactored the code a bit and hopefully it will look better now.
One thing I want to keep in mind: I don't want controller code to rely on c.Ctx and want it to be as "stateless" as possible. The reason is that it makes testing the controller method much easier. 
No need to setup context to test the method if it simply accepts all parameters as function parameters and returns simple objects
Please review now and let me know if this is better

eyalpost:
@astaxie I updated the code and also added documentation here: https://github.com/beego/beedoc/pull/580

httpResponse/response.go
```diff
@@ -0,0 +1,52 @@
+package httpResponse
```
astaxie: Could you also move this package to context?

eyalpost:
I think it doesn't belong in context. It's an implementation of the Renderer interface which also a user can do within his own packages

astaxie:
> I think it doesn't belong in context. It's an implementation of the Renderer interface which also a user can do within his own packages

I don't think so. httpResponse means response.

eyalpost:
No it's a helper for generating responses. Just like implementing an error interface does not mean it should be in the error package

eyalpost:
And it's similar to all helper method on the controller (ServeJson(), GetInt() etc. ) - they handle request\response but they're not part of the context package.

astaxie:
> No it's a helper for generating responses. Just like implementing an error interface does not mean it should be in the error package

yea, they are helper functions. But they just works for response. While response should be in context package. Like the `context/param/`. While this two helper response are response not other common function can reuse by other functions.

eyalpost:
So move it under context, not inside context? That's ok with me. Moved

astaxie:
> So move it under context, not inside context? That's ok with me. Moved

My idea is moving the file to context package. What's your concern?

eyalpost:
We have quite a lot of projects using beego and almost none of our code imports context package. The fact that it will not start importing it means (for me) these helpers are not in the correct place. 
Also, as I said before, it's for readability reasons:
`return nil, httpResponse.NotFound`
instead of 
`return nil, context.HttpResponseNotFound`
If you prefer, I can delete these helpers and implement them in my code. But I think they will be useful for other developers too.

astaxie:
> We have quite a lot of projects using beego and almost none of our code imports context package. The fact that it will now start importing it means (for me) these helpers are not in the correct place.

If you already use controller, context package already imported.



> Also, as I said before, it's for readability reasons:
> return nil, httpResponse.NotFound
> instead of
> return nil, context.HttpResponseNotFound
> If you prefer, I can delete these helpers and implement them in my code. But I think they will be useful for other developers too.

It still confuse for me that httpResponse have a `Redirect`. But context also have a `Redirect`. The other statuscode also should be parts of reponse. 

`return nil, context.NotFound`

eyalpost:
> If you already use controller, context package already imported.
I mean explicitly in code. We don't have explicit import of "github.com/astaxie/beego/context"

>It still confuse for me that httpResponse have a Redirect. But context also have a Redirect. The other statuscode also should be parts of response.
So how do you want to call it?

astaxie:
> I mean explicitly in code. We don't have explicit import of "github.com/astaxie/beego/context"

that's what I think is you don't need to import that now. and also don't need to import httpResponse. As you can just call these functions like `c.Ctx.Redirect` or `c.Ctx.Output.NotFound`

eyalpost:
> you can just call these functions like c.Ctx.Redirect or c.Ctx.Output.NotFound

But these functions require an actual Context object while the way they are built now is that they don't need a context. You can call the controller method in a test, and get a response all without needing to setup a context object. If I call c.Redirect in a test then it needs the entire http infrastructure (responseWriter etc.) but in my case it doesn't need anything because it just returns an "instruction" to redirect. The redirection is executed later

astaxie:
> But these functions require an actual Context object while the way they are built now is that they don't need a context. You can call the controller method in a test, and get a response all without needing to setup a context object. If I call c.Redirect in a test then it needs the entire http infrastructure (responseWriter etc.) but in my case it doesn't need anything because it just returns an "instruction" to redirect. The redirection is executed later

you can still rename current `httpResponse.Redirect` to `RedirectRender`. I think it should be a render while not the Redirect

eyalpost:
maybe `return httpResponse.RedirectResult(..)` ?

eyalpost:
or RedirectRender**er** ?

astaxie:
I like `return c.Ctx. RedirectRenderer(..) As we already define many renderer now. What do you think?

eyalpost:
Making it a context method again loses ability to test it. I will delete this helper for now to push this forward

astaxie:
Could you also move context/httpResponse/response.go to context? or remove it

eyalpost:
Done

astaxie:
@eyalpost thanks so much on this PR. Appreciated.

PR #2595, opened 2017-04-25
Add exported PrintTree function

ansiz:
We can do more thing with an exported PrintTree function, such as collect routers' information to set role base access control, create our own requests statistics and so on.

I know it is not necessary for all people, but I think we should have a way to get all routers information, I did not find any function or method can do that.(if there has been some functions can do that, hope you can point it out) so I add a wrapped printTree function named PrintTree.

Thank you anyway for your patience of reading this pull request. : )

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -105,29 +105,12 @@ func listConf(rw http.ResponseWriter, r *http.Request) {
 		tmpl.Execute(rw, data)
 
 	case "router":
-		var (
-			content = map[string]interface{}{
-				"Fields": []string{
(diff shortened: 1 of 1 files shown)


amrfaissal:
@ansiz It's better to create a feature branch based on `develop` first, then raise your pull request based on it. Can you also make sure `go test -v ./...` is green.

ansiz:
I did not find any errors reported by Travis CI in my environment，I will close this PR and re-test it.

Issue #2612, opened 2017-05-01
How to get the authenticated user name?


hsluoyz:
As far as I know, Beego already supports HTTP basic authentication. As [this docs](https://github.com/astaxie/build-web-application-with-golang/blob/master/en/14.4.md) said, Beego seems to also support other auth ways like HTTP digest authentication and OAuth. But I don't find a way to get the authenticated user name in a controller. Is there an official way to get the authenticated user name that works for all authentication ways?

Here I found a function in https://github.com/astaxie/beego/blob/master/plugins/auth/basic.go#L80 that can be used to get the user name for HTTP basic authentication. But it doesn't work for other auth ways.

```golang
// CheckAuth Checks the username/password combination from the request. Returns
// either an empty string (authentication failed) or the name of the
// authenticated user.
// Supports MD5 and SHA1 password entries
func (a *BasicAuth) CheckAuth(r *http.Request) string {
	s := strings.SplitN(r.Header.Get("Authorization"), " ", 2)
	if len(s) != 2 || s[0] != "Basic" {
		return ""
	}

	b, err := base64.StdEncoding.DecodeString(s[1])
	if err != nil {
		return ""
	}
	pair := strings.SplitN(string(b), ":", 2)
	if len(pair) != 2 {
		return ""
	}

	if a.Secrets(pair[0], pair[1]) {
		return pair[0]
	}
	return ""
}
```

zjjott:
use `c.Ctx.Request.BasicAuth()` can get the username and password,implement by `net/http` package 
BTW,username also can store by `ctx.Input.SetData(key,username)`

astaxie:
Add a user in context does not make sense. I think this should be handle in middleware which can store the data in `Input.Data`

hsluoyz:
``Input.Data`` is a very good place to store username.  I think all authentication middlewares like the current HTTP basic authentication or OAuth in the future should store the username into the same place. So that next module (for example, the RBAC handler) will exactly know where to find the username to enforce its access control. What about just storing it in ``Input.Data`` as key: ``username``?

PR #2613, opened 2017-05-01
Add authenticated user name to the context, left blank if it's an anonymous user.

hsluoyz:
This PR solves https://github.com/astaxie/beego/issues/2612.

Currently, it works with the [built-in HTTP basic authentication](https://github.com/astaxie/beego/blob/master/plugins/auth/basic.go). Other authentication methods like OAuth should follow this design by setting ``context.User`` to the authenticated user name or ID or token.


astaxie:
please see the #2612 discussion

PR #2618, opened 2017-05-02
Suggest all authentication modules to store the user name at the same place

hsluoyz:
Store the user name authenticated by HTTP basic authentication into the context input as key "username", all other authentication modules should follow this design.

This is to solve https://github.com/astaxie/beego/issues/2612

### DIFF
--- a/plugins/auth/basic.go
+++ b/plugins/auth/basic.go
@@ -59,7 +59,6 @@ func NewBasicAuthenticator(secrets SecretProvider, Realm string) beego.FilterFun
 	return func(ctx *context.Context) {
 		a := &BasicAuth{Secrets: secrets, Realm: Realm}
 		if username := a.CheckAuth(ctx.Request); username == "" {
-			ctx.Input.SetData("username", "")
 			a.RequireAuth(ctx.ResponseWriter, ctx.Request)
 		} else {
 			ctx.Input.SetData("username", username)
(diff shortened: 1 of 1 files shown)


plugins/auth/basic.go
```diff
@@ -59,7 +59,10 @@ func NewBasicAuthenticator(secrets SecretProvider, Realm string) beego.FilterFun
 	return func(ctx *context.Context) {
 		a := &BasicAuth{Secrets: secrets, Realm: Realm}
 		if username := a.CheckAuth(ctx.Request); username == "" {
+			ctx.Input.SetData("username", "")
```
zjjott: It is not necessary do this.a.RequireAuth will case redirect

hsluoyz: OK. I have updated the PR: https://github.com/astaxie/beego/pull/2618

astaxie:
https://golang.org/pkg/net/http/#Request.BasicAuth

hsluoyz:
BasicAuth() only gets the username when the authentication method is HTTP basic authentication. This means that if the auth way is another for example OAuth, the underlying access control module doesn't have a standard way to get the username, BasicAuth() won't work because HTTP basic authentication is never used at that time.

astaxie:
If it's OAuth I would like the plugin to handle this while not the framework do it

hsluoyz:
I agree with you that plugin should handle this and provide the username. In fact, what this PR modifies is just the HTTP basic authentication plugin (in path: ``plugins/auth/basic.go``).

The thing is that this HTTP basic authentication plugin what I tried to patch here happens to be in this main repo, together with the framework, but it doesn't make it a part of the framework.

Or maybe you want to separate this HTTP basic authentication plugin into a standalone repo, in a umbrella project like ``beego-plugins``. So the modification won't have anything to do with this repo.

Issue #740, opened 2014-08-14
beego 有没有权限处理的模块


zihuxinyu:
RBAC？

lei-cao:
@zihuxinyu Thank you for supporting beego. please update or submit your issue in English for a better international user friendly communication.

lei-cao:
@astaxie  close?
duplicated https://github.com/astaxie/beego/issues/432

hsluoyz:
Hi, I'm the author of [casbin](https://github.com/hsluoyz/casbin). It is an authorization library that supports models like MAC, RBAC, ABAC.

Related to RBAC, casbin has several advantages:
1. roles can be cascaded, aka roles can have roles.
2. support resource roles, so users have their roles and resource have their roles too. role = group here.
3. the permission assignments (or policy in casbin's language) can be persisted in files or database.

In fact, a light-weight project named [pybbs-go](https://github.com/tomoya92/pybbs-go) which is based on Beego already integrates casbin.

So please consider using casbin when beego implements RBAC security. Also let me know if there's any question:)

astaxie:
👍  I will add casbin to https://beego.me/docs/contributed/

hsluoyz:
I have submitted a PR here: https://github.com/astaxie/beego/pull/2620. It adds an authorization plugin, which
 supports ACL, RBAC, ABAC. Test is also included.

The interface is similar to the built-in HTTP basic authentication plugin and very easy to use. Welcome review!

PR #2620, opened 2017-05-03
Add an authorization plugin that supports ACL, RBAC based on casbin.

hsluoyz:
Add an authorization plugin that supports ACL, RBAC based on [casbin](https://github.com/hsluoyz/casbin).

It requires the built-in HTTP basic authentication by default, so the authz can get the user name and enforce the access control.

Test is also added in ``authz_test.go``.

This PR is to solve: https://github.com/astaxie/beego/issues/740


astaxie:

PR #2635, opened 2017-05-11
ISSUE2630  使用sqlite，orm中通过filter后的delete删除不成功

guanly:
https://github.com/astaxie/beego/issues/2630

### DIFF
--- a/orm/db.go
+++ b/orm/db.go
@@ -833,7 +833,11 @@ func (d *dbBase) DeleteBatch(q dbQuerier, qs *querySet, mi *modelInfo, cond *Con
 		if err := rs.Scan(&ref); err != nil {
 			return 0, err
 		}
-		args = append(args, reflect.ValueOf(ref).Interface())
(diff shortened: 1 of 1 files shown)


orm/db.go
```diff
@@ -833,7 +833,11 @@ func (d *dbBase) DeleteBatch(q dbQuerier, qs *querySet, mi *modelInfo, cond *Con
 		if err := rs.Scan(&ref); err != nil {
 			return 0, err
 		}
-		args = append(args, reflect.ValueOf(ref).Interface())
+		pkValue, err := d.convertValueFromDB(mi.fields.pk, reflect.ValueOf(ref).Interface(), tz)
+		if err != nil {
+			panic(fmt.Errorf("get pk value failed:  `%s` ", ref))
```
astaxie: if have the err, return the err, never panic inside the orm

guanly: Hmm, right. What about the issue? Does this code work? @astaxie

astaxie: yea, I think it make sense.

PR #2637, opened 2017-05-12
allow o.Raw(sql).QueryRows(&container) pass nested struct

alexsunxl:
users表

id | name
--- | --- 
2| sunxinle 

user_extra表

user_id | nick
--- | --- 
2| alex

```golang
type Users struct {
  Id int `json:"id" orm:"column(id);auto"`
  Name string `json:"name" orm:"column(name)"`
}

type UserExtra struct {
  Users
  Nick string `json:"nick" orm:"column(nick)"`
}

func main() {
  qb, _ = orm.NewQueryBuilder("mysql")
  sql := qb.Select("u.*", "e.nick").
    From("users as u").
    LeftJoin("user_extra as e").On("e.user_id = u.id").
    String()
  o := orm.NewOrm()
  var l []UserExtra

  o.Raw(sql).QueryRows(&l)
}

```
I used the similar code in my project , but can not get the 'name' field in the query result
I see the source find the setFiledValue and found that there is no special deal to Struct.
so i try to fix it by recusive

我在项目中使用了类似的代码，但是没办法获得UserExtra嵌套的Struct中的字段，比如User中的'name'
然后在源码中找到setFiledValue这个地方，发现确实没有对Struct做特殊的处理，所以现在把原来的代码放入一个函数中，通过递归去处理setFiledValue

### DIFF
--- a/orm/orm_raw.go
+++ b/orm/orm_raw.go
@@ -493,19 +493,33 @@ func (o *rawSet) QueryRows(containers ...interface{}) (int64, error) {
 					}
 				}
 			} else {
-				for i := 0; i < ind.NumField(); i++ {
-					f := ind.Field(i)
-					fe := ind.Type().Field(i)
-					_, tags := parseStructTag(fe.Tag.Get(defaultStructTagName))
-					var col string
-					if col = tags["column"]; col == "" {
-						col = snakeString(fe.Name)
-					}
-					if v, ok := columnsMp[col]; ok {
-						value := reflect.ValueOf(v).Elem().Interface()
-						o.setFieldValue(f, value)
+				// define recursive function
+				var recursiveSetField func(rv reflect.Value)
+				recursiveSetField = func(rv reflect.Value) {
+					for i := 0; i < rv.NumField(); i++ {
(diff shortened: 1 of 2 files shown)


astaxie:
Could you add the test case for this change?

alexsunxl:
@astaxie  I had added a test case  ^-^

orm/orm_raw.go
```diff
@@ -493,19 +493,33 @@ func (o *rawSet) QueryRows(containers ...interface{}) (int64, error) {
...
-					var col string
-					if col = tags["column"]; col == "" {
-						col = snakeString(fe.Name)
-					}
-					if v, ok := columnsMp[col]; ok {
-						value := reflect.ValueOf(v).Elem().Interface()
-						o.setFieldValue(f, value)
+				// define recursive function
+				var recursiveSetField func(rv reflect.Value)
+				recursiveSetField = func(rv reflect.Value) {
```
astaxie: why not just use `recursiveSetField:=....`

zzfuuu:
why not modify function o.Raw(sql).QueryRow as while?

PR #2640, opened 2017-05-13
Moved Security to Operation struct to support swagger API security auth

franzwilhelm:
# To support https://github.com/beego/bee/pull/436

Also changed `Security` field from 
```golang
Security map[string][]string
```
to 
```golang
Security []map[string][]string
```
as specified here: http://swagger.io/specification

### DIFF
--- a/swagger/swagger.go
+++ b/swagger/swagger.go
@@ -32,7 +32,6 @@ type Swagger struct {
 	Paths               map[string]*Item    `json:"paths" yaml:"paths"`
 	Definitions         map[string]Schema   `json:"definitions,omitempty" yaml:"definitions,omitempty"`
 	SecurityDefinitions map[string]Security `json:"securityDefinitions,omitempty" yaml:"securityDefinitions,omitempty"`
-	Security            map[string][]string `json:"security,omitempty" yaml:"security,omitempty"`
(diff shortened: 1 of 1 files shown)


swagger/swagger.go
```diff
@@ -32,7 +32,6 @@ type Swagger struct {
 	Paths               map[string]*Item    `json:"paths" yaml:"paths"`
 	Definitions         map[string]Schema   `json:"definitions,omitempty" yaml:"definitions,omitempty"`
 	SecurityDefinitions map[string]Security `json:"securityDefinitions,omitempty" yaml:"securityDefinitions,omitempty"`
-	Security            map[string][]string `json:"security,omitempty" yaml:"security,omitempty"`
```
astaxie: From the diff this field is deleted?

astaxie: From the spec. Security also useful. http://swagger.io/specification/#fixed-fields-15

```
security	[Security Requirement Object]	A declaration of which security schemes are applied for the API as a whole. The list of values describes alternative security schemes that can be used (that is, there is a logical OR between the security requirements). Individual operations can override this definition
```

franzwilhelm: @astaxie Fixed this now

PR #2656, opened 2017-05-19
Allow injecting dependencies to controllers

BorisBorshevsky:
With this PR I can inject dependencies and write code like this:

```
package routers

import ...

func InitRouters(ds lib.DataStore) {
	beego.AddNamespace(
		beego.NewNamespace("/api",
			beego.NSNamespace("/v1",
				beego.NSInclude(
					&controllers.DriverController{UsersDataStore: ds},
					//&controllers.RiderController{},
				),
			),
		),
	)
}
```

```
package main

import ...
func main() {
	datastore := lib.NewMemoryDatastore(20)
	
	
	routers.InitRouters(datastore)
	beego.Run()
}
```

This allows us to inject dependencies to the controllers, makes the project more modular, easier to disassemble to component and more testable.

WDYT?

### DIFF
--- a/router.go
+++ b/router.go
@@ -116,6 +116,7 @@ type ControllerInfo struct {
 	handler        http.Handler
 	runFunction    FilterFunc
 	routerType     int
+	initialize     func() ControllerInterface
 }
 
 // ControllerRegister containers registered router rules, controller handlers and filters.
@@ -181,6 +182,27 @@ func (p *ControllerRegister) Add(pattern string, c ControllerInterface, mappingM
 	route.methods = methods
 	route.routerType = routerTypeBeego
 	route.controllerType = t
+	route.initialize = func() ControllerInterface {
+		vc := reflect.New(route.controllerType)
+		execController, ok := vc.Interface().(ControllerInterface)
+		if !ok {
+			panic("controller is not ControllerInterface")
+		}
+
+		elemVal := reflect.ValueOf(c).Elem()
+		elemType := reflect.TypeOf(c).Elem()
+		execElem := reflect.ValueOf(execController).Elem()
+
+		numOfFields := elemVal.NumField()
+		for i := 0; i < numOfFields; i++ {
+			fieldVal := elemVal.Field(i)
+			fieldType := elemType.Field(i)
+			execElem.FieldByName(fieldType.Name).Set(fieldVal)
+		}
+
+		return execController
+	}
+
 	if len(methods) == 0 {
 		for _, m := range HTTPMETHOD {
 			p.addToRouter(m, pattern, route)
@@ -760,14 +782,10 @@ func (p *ControllerRegister) ServeHTTP(rw http.ResponseWriter, r *http.Request) 
 	// also defined runRouter & runMethod from filter
 	if !isRunnable {
 		//Invoke the request handler
-		vc := reflect.New(runRouter)
-		execController, ok := vc.Interface().(ControllerInterface)
-		if !ok {
-			panic("controller is not ControllerInterface")
-		}
+		var execController ControllerInterface = routerInfo.initialize()
 
 		//call the controller init function
-		execController.Init(context, runRouter.Name(), runMethod, vc.Interface())
+		execController.Init(context, runRouter.Name(), runMethod, execController)
 
 		//call prepare function
 		execController.Prepare()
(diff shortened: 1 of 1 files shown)


BorisBorshevsky:
@astaxie any thoughts?

router.go
```diff
@@ -187,6 +188,27 @@ func (p *ControllerRegister) addWithMethodParams(pattern string, c ControllerInt
 	route.methods = methods
 	route.routerType = routerTypeBeego
 	route.controllerType = t
+	route.initialize = func() ControllerInterface {
```
astaxie: if the c have no init value, could we just set initialize to nil?

BorisBorshevsky: we can, but I don't think we want.
having initialize set to nil doesn't seem like an expected behavior.
also the only way (I see) to find out if there are any init parameters, is to run the code that is already in the initialize function.

BorisBorshevsky: @astaxie how do we proceed?

astaxie: I am still thinking about this? As if add this initialize, the cost would be high? Could you write a bench test for this?

BorisBorshevsky: Looks like there is already bench test for this.
after a little bit testing, having initialize function as nil doesn't affect the results, the only thing that does affect them is removing reflection, which is hard since beego router is basically based on it... =\

Tested on Mac
``` 
goos: darwin
goarch: amd64
```
### Before
```
go test -bench=BenchmarkController -benchmem
BenchmarkController-4            1000000              1408 ns/op             666 B/op          6 allocs/op
```
### After
```
go test -bench=BenchmarkController -benchmem
BenchmarkController-4            1000000              1697 ns/op             682 B/op          8 allocs/op
```

astaxie: Nice

fushar:
Hey @astaxie, our team needs this cool feature. Could you make a release that includes this? Thank you in advance!

kyurdakok:
this works fine, just checkout develop branch and give it a shot and it worked. but there is another concern here think about you have 10 controllers and 20 "services" like this ```lib.DataStore``` accepting all services as a parameter wont look nice when you write code i think, cant we use a proper dependency injector inside beego ?

scf2k:
I have a controller with unexported field. I don't use any injections which this PR is about. And I have a panic:

```
2018/04/09 13:43:23.619 [C] [asm_amd64.s:573] Handler crashed with error reflect: reflect.Value.Set using value obtained using unexported field
2018/04/09 13:43:23.619 [C] [asm_amd64.s:573] /usr/local/Cellar/go/1.10/libexec/src/runtime/asm_amd64.s:573
2018/04/09 13:43:23.619 [C] [asm_amd64.s:573] /usr/local/Cellar/go/1.10/libexec/src/runtime/panic.go:505
2018/04/09 13:43:23.619 [C] [asm_amd64.s:573] /usr/local/Cellar/go/1.10/libexec/src/reflect/value.go:231
2018/04/09 13:43:23.619 [C] [asm_amd64.s:573] /usr/local/Cellar/go/1.10/libexec/src/reflect/value.go:1367
2018/04/09 13:43:23.619 [C] [asm_amd64.s:573] /Users/scf/golang/src/github.com/astaxie/beego/router.go:207
2018/04/09 13:43:23.619 [C] [asm_amd64.s:573] /Users/scf/golang/src/github.com/astaxie/beego/router.go:796
2018/04/09 13:43:23.619 [C] [asm_amd64.s:573] /usr/local/Cellar/go/1.10/libexec/src/net/http/server.go:2694
2018/04/09 13:43:23.619 [C] [asm_amd64.s:573] /usr/local/Cellar/go/1.10/libexec/src/net/http/server.go:1830
2018/04/09 13:43:23.619 [C] [asm_amd64.s:573] /usr/local/Cellar/go/1.10/libexec/src/runtime/asm_amd64.s:2361
```

If I export the field it works. It also works with unexported field and previous begoo version.

BorisBorshevsky:
@scf2k it was fixed in #2972 #2978 
@astaxie any chance you can tag a new version?, the bug still exists in 1.9.2

scf2k:
Ok, thanks!

hqy:
@astaxie any chance you can tag a new version?, the bug still exists in 1.9.2, please tag new version.

mushoffa:
Hi, how do I get the reference of dependency injection on controller? Thx

PR #2689, opened 2017-06-03
added statusCode in FilterMonitorFunc

jerson:
add statusCode for custom monitor tools like prometheus


jerson:
best solution #2692

Issue #2079, opened 2016-08-03
panic: sync: negative WaitGroup counter 在1.6.1还是出现了(beego/grace)


AdamGuan:
Please answer these questions before submitting your issue. Thanks!

1. What version of Go and beego are you using (`bee version`)?
1.6.1

2. What operating system and processor architecture are you using (`go env`)?
centos7.0, amd64

3. What did you do?
If possible, provide a recipe for reproducing the error.
A complete runnable program is good.
none

4. What did you expect to see?
none

5. What did you see instead?
none

![1](https://cloud.githubusercontent.com/assets/3422160/17353494/b80cb18a-5972-11e6-86db-0d2b25e9f3a1.png)

astaxie:
fix in beego1.7

privationel:
![image](https://cloud.githubusercontent.com/assets/26136195/24002703/61b56238-0a9c-11e7-836b-e4b0aee6e634.png)
![image](https://cloud.githubusercontent.com/assets/26136195/24002721/70d870a2-0a9c-11e7-880b-66bb175b3ee5.png)

PR #2714, opened 2017-06-13
fix panic  sync: negative WaitGroup counter

huwenbo:
when use graceful ,may cause panic: sync: negative WaitGroup counter , because Close() can be called multiple times. so when Close() check if already closed.


taowen:
reference https://github.com/astaxie/beego/issues/2079

Issue #2707, opened 2017-06-10
graceful: negative WaitGroup counter show again


dongjun111111:
Please answer these questions before submitting your issue. Thanks!

1. What version of Go and beego are you using (`bee version`)?

![default](https://user-images.githubusercontent.com/9988625/26999101-7b0cf442-4dc4-11e7-9dc6-3b1ecb3d6b75.png)
![default](https://user-images.githubusercontent.com/9988625/26999106-98b84366-4dc4-11e7-9524-f4abf95bb648.png)

2. What operating system and processor architecture are you using (`go env`)?

3. What did you do?
If possible, provide a recipe for reproducing the error.
A complete runnable program is good.

4. What did you expect to see?

5. What did you see instead?

Issue #2653, opened 2017-05-19
panic: sync: negative WaitGroup counter  (in beego 1.8.0)


gaohf:
panic: sync: negative WaitGroup counter

goroutine 8 [running]:
sync.(*WaitGroup).Add(0xc4200a50a8, 0x1)
	/xxx/go/src/sync/waitgroup.go:75 +0x1c9
github.com/astaxie/beego/grace.(*graceListener).Accept(0xc4202b40f0, 0xa0c888, 0xc4216aa1e0, 0xe66bc0, 0xc4202b4210)
	/xxx/src/github.com/astaxie/beego/grace/listener.go:45 +0x15a
net/http.(*Server).Serve(0xc4200d3080, 0xe65300, 0xc4202b40f0, 0x0, 0x0)
	/xxx/go/src/net/http/server.go:2643 +0x228
github.com/astaxie/beego/grace.(*Server).Serve(0xc4200a5080, 0x2, 0x2)
	/xxx/src/github.com/astaxie/beego/grace/server.go:36 +0x4b
github.com/astaxie/beego/grace.(*Server).ListenAndServe(0xc4200a5080, 0xc, 0xe5fb00)
	/xxx/src/github.com/astaxie/beego/grace/server.go:75 +0x282
github.com/astaxie/beego.(*App).Run.func2(0xc4202ac180, 0xc, 0xc420152250, 0xc4202b0a80)
	/xxx/src/github.com/astaxie/beego/app.go:130 +0xd6
created by github.com/astaxie/beego.(*App).Run
	/xxx/src/github.com/astaxie/beego/app.go:135 +0x713

PR #2717, opened 2017-06-13
fix panic sync: negative WaitGroup counter

huwenbo:
when use graceful ,may cause panic: sync: negative WaitGroup counter , because Close() can be called multiple times. so when Close() check if already closed.

reference #2079 #2707 #2653


huwenbo:
the reason cause this panic https://github.com/golang/go/commit/d1e16d06b4df98930b5c6b0775cdd414dfdebd50

Issue #2713, opened 2017-06-12
Email validation on empty field


osavchenko:
Why it doesn't ignore empty field and trying to validate it?

miraclesu:
Because the `Email` validation just valid a field content is an email address or not. An empty field isn't an email address, isn't it?

I think custom function can meet your expectation.
```
import (
	"github.com/astaxie/beego/validation"
)

type user struct {
	Id    int
	Email string `valid:"MyEmail"`
	Age   int    `valid:"Required;Range(1, 140)"`
}

func MyEmail(v *validation.Validation, obj interface{}, key string) {
	name, ok := obj.(string)
	if !ok {
		// wrong use case?
		// set Error?
		return
	}

	// valid true
	if len(name) == 0 {
		return
	}

	v.Email(obj, key)
}

func main() {
	valid := validation.Validation{}
	if err := validation.AddCustomFunc("MyEmail", MyEmail); err != nil {
		// hadle error
	}
	u := user{Email: "", Age: 40}
	b, err := valid.Valid(u)
	if err != nil {
		// handle error
	}
	if !b {
		// validation does not pass
		// blabla...
	}
}
```

rfyiamcool:
+1

dmitryt:
+1. Quite a strange behaviour. If I want to make my email field required, I can type 
``Email     string `valid:"Required;Email"` ``
Does it mean, that for any optional field with predefined validation(like Email, Phone, etc.) I should implement custom function?

PR #2741, opened 2017-07-03
validation: support required option for some struct tag valids

miraclesu:
Ignore empty field valid with `RequiredFirst` option.

#2713 
#2522 

```
func TestSkipValid(t *testing.T) {
	type User struct {
		ID int

		Email    string `valid:"Email"`
		ReqEmail string `valid:"Required;Email"`

		IP    string `valid:"IP"`
		ReqIP string `valid:"Required;IP"`

		Mobile    string `valid:"Mobile"`
		ReqMobile string `valid:"Required;Mobile"`

		Tel    string `valid:"Tel"`
		ReqTel string `valid:"Required;Tel"`

		Phone    string `valid:"Phone"`
		ReqPhone string `valid:"Required;Phone"`

		ZipCode    string `valid:"ZipCode"`
		ReqZipCode string `valid:"Required;ZipCode"`
	}

	u := User{
		ReqEmail:   "a@a.com",
		ReqIP:      "127.0.0.1",
		ReqMobile:  "18888888888",
		ReqTel:     "02088888888",
		ReqPhone:   "02088888888",
		ReqZipCode: "510000",
	}

	valid := Validation{}
	b, err := valid.Valid(u)
	if err != nil {
		t.Fatal(err)
	}
	if b {
		t.Fatal("validation should not be passed")
	}

	valid = Validation{RequiredFirst: true}
	b, err = valid.Valid(u)
	if err != nil {
		t.Fatal(err)
	}
	if !b {
		t.Fatal("validation should be passed")
	}
}
```

### DIFF
--- a/validation/validators.go
+++ b/validation/validators.go
@@ -23,6 +23,16 @@ import (
 	"unicode/utf8"
 )
 
+// CanSkipFuncs will skip valid if RequiredFirst is true and the struct field's value is empty
+var CanSkipFuncs = map[string]struct{}{
+	"Email":   struct{}{},
+	"IP":      struct{}{},
+	"Mobile":  struct{}{},
+	"Tel":     struct{}{},
+	"Phone":   struct{}{},
+	"ZipCode": struct{}{},
+}
+
 // MessageTmpls store commond validate template
 var MessageTmpls = map[string]string{
 	"Required":     "Can not be empty",
--- a/validation/README.md
+++ b/validation/README.md
@@ -64,6 +64,9 @@ Struct Tag Use:
 
 	func main() {
 		valid := validation.Validation{}
+		// ignore empty field valid
+		// see CanSkipFuncs
+		// valid := validation.Validation{RequiredFirst:true}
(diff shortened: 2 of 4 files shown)


validation/validators.go
```diff
@@ -23,6 +23,16 @@ import (
...
 
+// CanSkipFuncs will skip valid if RequiredFirst is true and the struct field's value is empty
+var CanSkipFuncs = map[string]struct{}{
+	"Email":   struct{}{},
+	"IP":      struct{}{},
+	"Mobile":  struct{}{},
+	"Tel":     struct{}{},
+	"Phone":   struct{}{},
+	"ZipCode": struct{}{},
+}
```
gadelkareem, line 34: @miraclesu why did you limit it to only those templates? I think any validation should be skipped if the field is empty and not required. I am having a use case with `MinSize` which is not handled here.

gadelkareem: never mind, I just added them since it is a public var

PR #2744, opened 2017-07-06
[Proposal] Database Migrations;

gnanakeethan:
Summary: The database migrations now can be created using the methods on
the migration struct. it does not break any existing migration features.
it upgrades the migration struct and adds few more struct types so that
the migrations can be efficiently generated for create, alter, reverse,
drop.

Current Features:
* Supports creation of columns
   * `m.NewCol("name").SetDataType("VARCHAR(10)").SetNullable("true")`
   * **NOTE** `SetNullable` & `SetDefault` methods should not be called on
   same column for consistency
* Supports addition of primary keys
   * `m.PriCol("id").SetDataType("INT(10)").SetNullable("true")`
   * **NOTE** `setAuto(true)` can be only called on Primary keys
* Supports addition of unique keys
   * `m.UniCol("unique_index","column_name").SetDataType("VARCHAR(23)").SetNullable("true")`
   * **NOTE** `UniCol` can be called again with the same index name to
   add column to the index
* Supports rename of columns
   * `m.RenameColumn("from_name","to_name")`
   * Allows standard column methods and methods such that, `SetOldDefault` allows
   reversibility of renames
* Supports Foreign Keys
   * `m.ForeignCol("column","foreign_col","foreign_table")`
   * Allows sequential standard column methods for column definition

### DIFF
--- a/migration/doc.go
+++ b/migration/doc.go
@@ -1,5 +1,3 @@
-package migration
-
 /* Package migration enables you to generate migrations back and forth. It generates both migrations.
 
 
@@ -39,10 +37,5 @@ m.RenameColumn("from","to")...
 //Supports standard column methods, automatic reverse.
 m.ForeignCol("local_col","foreign_col","foreign_table")
 
-
-
-
-
-
-
 */
+package migration
--- a/migration/doc.go
+++ b/migration/doc.go
@@ -1,0 +1,48 @@
+package migration
+
+/* Package migration enables you to generate migrations back and forth. It generates both migrations.
+
+
+//Creates a table
+m.CreateTable("tablename","InnoDB","utf8");
+
+
+
+//Alter a table
+m.AlterTable("tablename")
+
+
+
+//Standard Column Methods
+* SetDataType
+* SetNullable
+* SetDefault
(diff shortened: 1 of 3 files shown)


gnanakeethan:
@franzwilhelm @saturn4er @astaxie  Can you please give some suggestions if any to improve?

astaxie:
@gnanakeethan That's awesome. Could you add a `doc.go` describe how to use ddl

gnanakeethan:
@astaxie  Please have a view on both of PRs. 


cc: @saturn4er @franzwilhelm

migration/doc.go
```diff
@@ -0,0 +1,48 @@
...
+//Supports standard column methods, automatic reverse.
+m.ForeignCol("local_col","foreign_col","foreign_table")
+
+
+
+
+
+
+
+*/
```
astaxie: you need put all these docs on the top. before the package

gnanakeethan: done and pushed

gnanakeethan:
@astaxie I have updated the code. Can we merge this thing out?

PR #2849, opened 2017-09-05
Make parent process exit gracefully.

iclinux:
With beego.BConfig.Listen.Graceful  enabled, when received SIGHUP, we'll fork a child process.
But the parent process still have jobs to finish, So we can't kill the parent process directly.


iclinux:
Please take a look at Why "continuous-integration/travis-ci/pr — The Travis CI build failed" ?

astaxie:
The Travis CI change the MySQL. So the test ORM for mysql all failed

PR #2852, opened 2017-09-06
validation: fix ErrorMap, added AddError(key, message)

Radar8:
AddError func we can use like this:
```
if emailAlreadyExist {
     valid.AddError("email.email", `This email already exist!`)
}
c.Data["Errors"] = valid.ErrorMap()
```
I did'n understand difference between Name and Field, may be, we can use like:  valid.AddError("email",..)

In template:
```
{{ if .Errors.email }}
  <div class="alert alert-danger" role="alert">
    {{ range .Errors.email }}
        {{ . }}<br/>
    {{ end }}
  </div>
{{ end }}
```

### DIFF
--- a/validation/validation.go
+++ b/validation/validation.go
@@ -112,7 +112,7 @@ type Validation struct {
 	RequiredFirst bool
 
 	Errors    []*Error
-	ErrorsMap map[string]*Error
+	ErrorsMap map[string][]*Error
 }
 
 // Clear Clean all ValidationError.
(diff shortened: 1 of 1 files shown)


validation/validation.go
```diff
@@ -112,7 +112,7 @@ type Validation struct {
 	RequiredFirst bool
 
 	Errors    []*Error
-	ErrorsMap map[string]*Error
+	ErrorsMap map[string][]*Error
```
astaxie: Why change to error slice? you want to support multi error for one field?

Radar8: Yes. If field has many errors, we can show it all.

PR #2878, opened 2017-09-26
Add the ability to unregister fixed routes

silviucm:
Certain web application inherit most of the routes from a base application by using the underscore import 
(e.g. import _ "myoldapp.com/routers").
The new application might want to only overwrite certain pages, such as "/about" or "/faq"

The proposed new UnregisterFixedRoute method allows unregistering of the specified paths.
 Usage (replace "GET" with "*" for all methods):
```go
  beego.UnregisterFixedRoute("/yourpreviouspath", "GET")

  beego.Router("/yourpreviouspath", yourControllerAddress, "get:GetNewPage")
```
The children paths are left intact.
For example, /yourpreviouspath/oldchildsubpath should still continue to function in legacy mode.

### DIFF
--- a/app.go
+++ b/app.go
@@ -207,6 +207,72 @@ func Router(rootpath string, c ControllerInterface, mappingMethods ...string) *A
 	return BeeApp
 }
 
+// UnregisterFixedRoute unregisters the route with the specified fixedRoute. It is particularly useful
+// in web applications that inherit most routes from a base webapp via the underscore
+// import, and aim to overwrite only certain paths.
+// The method parameter can be empty or "*" for all HTTP methods, or a particular
+// method type (e.g. "GET" or "POST") for selective removal.
+//
+// Usage (replace "GET" with "*" for all methods):
+//  beego.UnregisterFixedRoute("/yourpreviouspath", "GET")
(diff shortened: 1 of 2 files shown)


silviucm:
Hi @astaxie - a quick ping to see if there is any interest in the route removal functionality ? It is very basic at this point, and only focuses on static, but it can grow into something bigger. 

Let me know, because otherwise I will maintain my own fork and keep the code in a separate file to allow for clean merge.

Cheers

astaxie:
Could you write some test case for this feature?

silviucm:
Hi @astaxie  

I have now provided a new file **unregroute_test.go** which has the following scenario:

Given three routes:
/
/level1
/level1/level2

And two controllers ("before" and "after") which emit content for each of those three roots.

The following scenarios are tested:

Root
- replace the action for "/" ("before" controller) with action from the "after" controller
- ensure the "/level1" and "/level1/level2" still return content based on "before" actions

Level1
- replace the action for "/level1" ("before" controller) with action from the "after" controller
- ensure the "/" and "/level1/level2" still return content based on "before" actions

Level2
- replace the action for "/level1/level2" ("before" controller) with action from the "after" controller
- ensure the "/" and "/level1" still return content based on "before" actions

Cheers

PR #2949, opened 2017-11-29
Change the type of HTTPMETHOD

huhaocc:
Use map[string]bool for  HTTPMETHOD.  It is "golang style" Set.

### DIFF
--- a/app.go
+++ b/app.go
@@ -24,8 +24,8 @@ import (
 	"net/http/fcgi"
 	"os"
 	"path"
-	"time"
 	"strings"
+	"time"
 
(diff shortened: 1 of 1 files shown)


astaxie:
./app.go:258: cannot use m (type bool) as type string in map index
./app.go:261: cannot use m (type bool) as type string in map index
./app.go:262: cannot use m (type bool) as type string in map index
./app.go:265: cannot use m (type bool) as type string in map index
./app.go:265: cannot use m (type bool) as type string in argument to findAndRemoveTree
FAIL	github.com/astaxie/beego [build failed]

PR #2950, opened 2017-11-29
Change the type of HTTPMETHOD

huhaocc:
Use map[string]bool for HTTPMETHOD. It is "golang style" Set.


astaxie:
Why this is Go style? I don't think so.

Issue #2972, opened 2017-12-09
Handler crashed with error reflect: reflect.Value.Set using value obtained using unexported field


zjxpcyc:
Please answer these questions before submitting your issue. Thanks!

1. What version of Go and beego are you using (`bee version`)?
go 1.9.2
beego 1.9.1

2. What operating system and processor architecture are you using (`go env`)?
```bash
GOARCH="amd64"
GOBIN=""
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/yansen/go"
GORACE=""
GOROOT="/usr/local/go1.9"
GOTOOLDIR="/usr/local/go1.9/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build151996307=/tmp/go-build -gno-record-gcc-switches"
CXX="g++"
CGO_ENABLED="1"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
```

3. What did you do?
If possible, provide a recipe for reproducing the error.
A complete runnable program is good.
这个情况在我开发环境是正常的，开发环境的版本跟生产环境不一样。
我有一个 UserController , 里面自定义了一个属性 dao , 用来调用对应的 models 及方法。
dao 是小写，也就是私有的。
然后报错出现在 router.go 的 [206行](https://github.com/astaxie/beego/blob/master/router.go#L206) 

4. What did you expect to see?
正常编译运行

5. What did you see instead?
```bash
2017/12/09 11:46:10.840 [C] [asm_amd64.s:509] the request url is  /api/user
2017/12/09 11:46:10.840 [C] [asm_amd64.s:509] Handler crashed with error reflect: reflect.Value.Set using value obtained using unexported field
2017/12/09 11:46:10.840 [C] [asm_amd64.s:509] /usr/local/go1.9/src/runtime/asm_amd64.s:509
2017/12/09 11:46:10.840 [C] [asm_amd64.s:509] /usr/local/go1.9/src/runtime/panic.go:491
[...]

PR #2978, opened 2017-12-13
fix bug #2972

BorisBorshevsky:
Bug on private fields 
#2972


astaxie:
Thanks, this is a hot fix

PR #2985, opened 2017-12-20
fix orm fields SetRaw function error judge problem

terryding77:
```
func (e *BooleanField) SetRaw(value interface{}) error {
	switch d := value.(type) {
	case bool:
		e.Set(d)
	case string:
		v, err := StrTo(d).Bool()
		if err != nil {
			e.Set(v)
		}
		return err
	default:
		return fmt.Errorf("<BooleanField.SetRaw> unknown value `%s`", value)
	}
	return nil
}
```
there are some fields have `SetRaw` func convert from string.
for convert error, we should set the field when `err == nil`
and return the `err` after judgement.

### DIFF
--- a/orm/models_fields.go
+++ b/orm/models_fields.go
@@ -86,7 +86,7 @@ func (e *BooleanField) SetRaw(value interface{}) error {
 		e.Set(d)
 	case string:
 		v, err := StrTo(d).Bool()
-		if err != nil {
+		if err == nil {
 			e.Set(v)
 		}
 		return err
@@ -191,7 +191,7 @@ func (e *TimeField) SetRaw(value interface{}) error {
 		e.Set(d)
 	case string:
 		v, err := timeParse(d, formatTime)
-		if err != nil {
+		if err == nil {
 			e.Set(v)
 		}
 		return err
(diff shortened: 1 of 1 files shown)


terryding77:
```
=== RUN   TestFormatHeader_0
--- FAIL: TestFormatHeader_0 (0.00s)
	logger_test.go:35: 2017-12-20 07:04:30.570698946 +0000 UTC
=== RUN   TestFormatHeader_1
--- FAIL: TestFormatHeader_1 (0.00s)
	logger_test.go:53: 2017-12-20 07:04:30.570790037 +0000 UTC

=== RUN   TestSmtp
unable to WriteMsg to adapter:smtp,error:535 5.7.8 Username and Password not accepted. Learn more at
5.7.8  https://support.google.com/mail/?p=BadCredentials p68sm2132757itc.26 - gsmtp
--- PASS: TestSmtp (30.09s)
FAIL
FAIL	github.com/astaxie/beego/logs	32.096s
?   	github.com/astaxie/beego/logs/alils	[no test files]
?   	github.com/astaxie/beego/logs/es	[no test files]
?   	github.com/astaxie/beego/migration	[no test files]
```
dose those `test fail` caused by my modification?

Issue #2998, opened 2017-12-23
unable to add column with ALTER TABLE


moririnson:
Please answer these questions before submitting your issue. Thanks!

1. What version of Go and beego are you using (`bee version`)?

```
______
| ___ \
| |_/ /  ___   ___
| ___ \ / _ \ / _ \
| |_/ /|  __/|  __/
\____/  \___| \___| v1.9.1

├── Beego     : 1.9.2
├── GoVersion : go1.9.2
├── GOOS      : windows
├── GOARCH    : 386
├── NumCPU    : 8
├── GOPATH    : D:\Program_Files\go\third_party
├── GOROOT    : D:\Program_Files\go\
├── Compiler  : gc
└── Date      : Saturday, 23 Dec 2017

2. What operating system and processor architecture are you using (`go env`)?

set GOARCH=386
set GOBIN=
set GOEXE=.exe
set GOHOSTARCH=386
set GOHOSTOS=windows
set GOOS=windows
set GOPATH=D:\Program_Files\go\third_party
set GORACE=
set GOROOT=D:\Program_Files\go
set GOTOOLDIR=D:\Program_Files\go\pkg\tool\windows_386
set GCCGO=gccgo
set GO386=sse2
set CC=gcc
set GOGCCFLAGS=-m32 -mthreads -fmessage-length=0
set CXX=g++
set CGO_ENABLED=1
set CGO_CFLAGS=-g -O2
set CGO_CPPFLAGS=
set CGO_CXXFLAGS=-g -O2
set CGO_FFLAGS=-g -O2
set CGO_LDFLAGS=-g -O2
set PKG_CONFIG=pkg-config
```

3. What did you do?

I wanted to add a coulmn to user table. So I tried such code.

```
// Run the migrations
func (m *AddPasswordToUsersTable_20171223_140454) Up() {
	m.AlterTable("users")
	m.NewCol("password").SetNullable(false).SetDataType("VARCHAR(255)")
	m.SQL(m.GetSQL())
}
```

4. What did you expect to see?
"password" column will be added.

5. What did you see instead?
I got a syntax error as below and column was not added.

```
You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '' at line 2
```

moririnson:
I printed SQL generated by the code.

```
2017/12/23 14:16:30 INFO     ▶ 0009 |> 2017/12/23 14:16:28.343
[I] exec sql: ALTER TABLE `users`
[...]

PR #2999, opened 2017-12-23
fix the issue #2998

moririnson:
fix bug #2998


gnanakeethan:
@astaxie This needs to be merged. There was a mistake in the code I pushed earlier.

astaxie:
thanks

PR #3004, opened 2017-12-26
redis cache: make MaxIdle configurable

pcallewaert:
Hi,

One of our applications has a very high req/s, and is hitting the redis cache for every request. In peak moments we were getting socket errors on the server. After investigation we noticed a lot of TIME_WAIT connections to the redis server. Increasing the MaxIdle to a higher value fixed this as connections are not closed as fast as before. In this PR we made the MaxIdle configuration of redigo configurable in the connection parameter of beego cache. The default stays the same.

Kind regards,
Pieter Callewaert

### DIFF
--- a/cache/redis/redis.go
+++ b/cache/redis/redis.go
@@ -171,7 +171,7 @@ func (rc *Cache) StartAndGC(config string) error {
 		cf["password"] = ""
 	}
 	if _, ok := cf["maxIdle"]; !ok {
-		cf["maxIdle"] = ""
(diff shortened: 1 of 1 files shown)


pcallewaert:
No idea why the build failed... Ay ideas?

terryding77:
@pcallewaert emmm, I have met those build failed either. I think it is not your fault. It may be disappeared after next build.(as my pull request)

Hongery:
why ? can't use redis-cluster ?

PR #3020, opened 2018-01-05
add hook function for orm

godblesshugh:
add six hook function:
beforeInsert
afterInsert
beforeUpdate
afterUpdate
beforeDelete
afterDelete


astaxie:
@godblesshugh please fix the conflict

orm/orm.go
```diff
@@ -157,13 +322,23 @@ func (o *orm) ReadOrCreate(md interface{}, col1 string, cols ...string) (bool, i
...
+	}
 	id, err := o.alias.DbBaser.Insert(o.db, mi, ind, o.alias.TZ)
 	if err != nil {
 		return id, err
 	}
-
 	o.setPk(mi, ind, id)
-
+	if hookFunc, ok := o.afterInsert[name]; ok {
+		go func(md interface{}) {
```
astaxie: You can't use goroutine for these functions. As it may cause the data not updated.

orm/orm.go
```diff
@@ -230,20 +405,48 @@ func (o *orm) InsertOrUpdate(md interface{}, colConflitAndArgs ...string) (int64
...
+		if hookErr != nil {
+			return 0, hookErr
+		}
+	}
+	num, err := o.alias.DbBaser.Update(o.db, mi, ind, o.alias.TZ, cols)
+	if err != nil {
+		return num, err
+	}
+	if hookFunc, ok := o.afterUpdate[name]; ok {
+		go func(md interface{}, cols ...string) {
```
astaxie: same as above

orm/orm.go
```diff
@@ -230,20 +405,48 @@ func (o *orm) InsertOrUpdate(md interface{}, colConflitAndArgs ...string) (int64
...
+	}
 	num, err := o.alias.DbBaser.Delete(o.db, mi, ind, o.alias.TZ, cols)
 	if err != nil {
 		return num, err
 	}
 	if num > 0 {
 		o.setPk(mi, ind, 0)
 	}
+	if hookFunc, ok := o.afterDelete[name]; ok {
+		go func(md interface{}, cols ...string) {
```
astaxie: same as above

orm/orm.go
godblesshugh: Using goroutine only if there is 'after' hook function, 'before' hook will not use goroutine.

astaxie: but if after hook failed, what can we do?

godblesshugh: sorry i don't follow, 'after' hook failed can cause the same as other function failed, it'll not cause the whole 'CRUD' function failed.
maybe need a recover() here? thought it's better to let recover() called by owner of hook().

astaxie: I just discuss with the gorm author, in gorm any hooks failed will rollback this action

godblesshugh: sorry for delay.
after study gorm's callback method, i found that gorm using tx(transaction) as param of ['afterInsert', 'afterUpdate', 'afterDelete'] to achieve db rollback, means in those hook functions, we still can't use the 'result' of a CRUD function, that's not an 'after' hook to me, because an 'after' function should let we use an 'updated' object(result after object be updated) right? those 'after' callbacks can all put into current tx, because they using current tx, sounds like an 'append' to me. so maybe they also need a PR :)
My thoughts:
'After' hook should let us produce the 'after-transaction' object.
In this way, 'after' hook still can failed, but no need to add db rollback in it. so maybe no goroutine for 'after' hook, let developers to choose whether goroutine or not?

godblesshugh: Maybe 'after' hook need goroutine after all, avoid 'after' hook failed causing the whole CRUD function failed.

astaxie: I think the after hook just return error, but no panic

astaxie: Here have some ideas https://github.com/ilibs/gosql#hooks

godblesshugh: a bit weird this way, CRUD function return error but with correct influenced lines count. 
and add hook functions into object like gorm maybe more clearly?
`func (s *Product) AfterCreate()`

orm/orm.go
```diff
@@ -335,9 +335,7 @@ func (o *orm) Insert(md interface{}) (int64, error) {
 	}
 	o.setPk(mi, ind, id)
 	if hookFunc, ok := o.afterInsert[name]; ok {
-		go func(md interface{}) {
-			hookFunc(md)
-		}(md)
+		return id, hookFunc(md)
```
godblesshugh: still thought there are no need for 'after' hooks return error or panic. 'after' hook's error can handled by developers, orm just need to trigger the hooks in goroutine mode.
better this way:
```
go func(md interface{}) {
    defer func() {
        if r := recover(); r != nil {
            fmt.Errorf("<Ormer.hook> hook AfterInsert error: %s", r)
        }
    }()
    hookFunc(md)
}(md)
```

orm/orm.go
```diff
@@ -88,9 +89,173 @@ type Params map[string]interface{}
...
+	isTx         bool
+	beforeInsert map[string]HookFunc
+	afterInsert  map[string]HookFunc
+	beforeUpdate map[string]HookFunc // can not add or delete update cols
+	afterUpdate  map[string]HookFunc
+	beforeDelete map[string]HookFunc
+	afterDelete  map[string]HookFunc
+}
+
+type HookFunc func(md interface{}, cols ...string) error
```
gogitops, line 103: warning: exported type HookFunc should have comment or be unexported (golint)

flycash:
We will support those feature in v2.0.0

Issue #1261, opened 2015-07-14
生成文档时提示该错误，具体是什么原因呢


mohong122:
panic: runtime error: index out of range

goroutine 1 [running]:
main.getModel(0xc082206340, 0xb, 0xc082206340, 0x6, 0xc082206347, 0x4, 0xc08230d370, 0x4, 0x0, 0x0, ...)
        E:/wwwroot/Go/src/github.com/beego/bee/g_docs.go:574 +0x1050
main.parserComments(0xc0820ea4a0, 0xc0820e85ad, 0x3, 0xc082217e70, 0xe, 0xc082005761, 0x1a, 0x0, 0x0)
        E:/wwwroot/Go/src/github.com/beego/bee/g_docs.go:406 +0x13e6
main.analisyscontrollerPkg(0xc082005761, 0x1a)
        E:/wwwroot/Go/src/github.com/beego/bee/g_docs.go:308 +0xa03
main.generateDocs(0xc082059530, 0x20)
        E:/wwwroot/Go/src/github.com/beego/bee/g_docs.go:152 +0x54d
main.generateCode(0xb98980, 0xc082003ee0, 0x1, 0x1, 0x0)
        E:/wwwroot/Go/src/github.com/beego/bee/g.go:125 +0xcdf
main.main()
        E:/wwwroot/Go/src/github.com/beego/bee/bee.go:113 +0x3af

---
程序运行没有问题，就是生成文档时报错，不知道是什么原因呢

lei-cao:
@mohong122 Thank you for supporting beego. please update or submit your issue in English for a better international user friendly communication.

JessonChan:
@astaxie 
this issue can be closed.

PR #3046, opened 2018-01-28
Handle pointer validation

aspacca:
I have a struct like the following that I populate from json unmarshaling:
```
type Entity struct {
	ID                  int     `json:"-" db:"rowid"`
	Status            *string `json:"status" db:"status" valid:"Match(/^(active|inactive|deleted)$/)"`
}
```
I rely to the fact that `Status` is set to `nil` when the tag is not present in the json (that's a different behaviour from using `omitempty` since I want to differ between the field not being set at all or sent as an empty string)
beego validation doesn't have any support for pointer in the struct, my PR take care of it (I will add tests)
Please, note that this may change expected behaviour: since beego provides a `Required` validator I supposed that all the other validators allow nil values and `Required` was a kind of a flag that overrides this behaviour, while it seems that `Required` is just a standard validator for multi types "not empty" logic other validators don't accept nil values.
My PR changes this behaviour, so that if the value is nil and the element has no `Required` it will pass validation

PS: I will close https://github.com/astaxie/beego/pull/3043 and continue here instead

### DIFF
--- a/validation/validation.go
+++ b/validation/validation.go
@@ -245,7 +245,17 @@ func (v *Validation) ZipCode(obj interface{}, key string) *Result {
 }
 
 func (v *Validation) apply(chk Validator, obj interface{}) *Result {
-	if chk.IsSatisfied(obj) {
+	if reflect.TypeOf(obj).Kind() == reflect.Ptr {
+		if reflect.ValueOf(obj).IsNil() {
+			if "Required" != chk.GetKey() {
+				return &Result{Ok: true}
+			}
+		} else {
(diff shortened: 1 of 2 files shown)


andreasph:
I don't understand why  #1261.2 and #1261.5 failed, any hint?

aspacca:
[#1261.2](https://travis-ci.org/astaxie/beego/jobs/334417865) and [#1261.5](https://travis-ci.org/astaxie/beego/jobs/334417868) fails don't seem to be related to the changes in the PR

Issue #3063, opened 2018-02-20
"astaxie/beego/config.go" use BEEGO_MODE and BEEGO_RUNMODE.


takeo-lvgs:
Please answer these questions before submitting your issue. Thanks!

1. What version of Go and beego are you using (`bee version`)?

1.9.2

2. What operating system and processor architecture are you using (`go env`)?

GOARCH="amd64"
GOBIN=""
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/h-takeo/go"
GORACE=""
GOROOT="/usr/local/go"
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build784201554=/tmp/go-build -gno-record-gcc-switches"
CXX="g++"
CGO_ENABLED="1"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"

3. What did you do?
If possible, provide a recipe for reproducing the error.
A complete runnable program is good.

Perhaps two words are used in one sense.

4. What did you expect to see?

I want to execute "bee run -runmode=dev".
Now, I execute it. My application can't load "dev.app.conf". My application load "app.conf".

Issue #2916, opened 2017-11-01
Bee runmode option not work


Rustavil:
Please answer these questions before submitting your issue. Thanks!

1. What version of Go and beego are you using (`bee version`)?
├── Beego     : 1.9.0
├── GoVersion : go1.9.1
├── GOOS      : linux
├── GOARCH    : amd64
├── NumCPU    : 4
├── GOPATH    : /home/rustavil/go
├── GOROOT    : /usr/local/go
├── Compiler  : gc
└── Date      : Wednesday, 1 Nov 2017

2. What operating system and processor architecture are you using (`go env`)?

GOARCH="amd64"
GOBIN=""
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/rustavil/go"
GORACE=""
GOROOT="/usr/local/go"
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build269144435=/tmp/go-build -gno-record-gcc-switches"
CXX="g++"
CGO_ENABLED="1"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"

app.conf:
```
appname = server
runmode = dev
Graceful = true
RecoverPanic = true
copyrequestbody = true

[dev]
HTTPAddr = "0.0.0.0"
HTTPPort = 8080

[prod]
HTTPAddr = "0.0.0.0"
HTTPPort = 80
```
I run the application with the command: bee run -runmode=prod <app_path>

I expected the application to work on port 80. But in fact the application uses dev runmode and works on port 8080

osavchenko:
You should use `BEEGO_RUNMODE` environment variable

Rustavil:
Thanks! I tried, it didn't help
in the logs, I see that the **runmode** is changing, but the server runs on the **8080** port

osavchenko:
And what port and why you expect to see?


5 груд. 2017 р. о 22:32 Rustavil <notifications@github.com> пише:

Rustavil:
In the above configuration, the server runs on port **8080**

[...]

PR #3064, opened 2018-02-20
fix the issue #3063

takeo-lvgs:
I fix the issue #3063.

In #2916, 
You had a discussion that "BEEGO_RUNMODE" should be used, so we modified it to "BEEGO_RUNMODE".


takeo-lvgs:
Test is failed. Cause is #3065.

PR #3106, opened 2018-04-04
add a new method GetAll for the beego session Store interface,and implements this method in all of the implementors

Kavin-Cao:
Sometimes, get all key-value in the beego session is important, Now I add a new method GetAll for the beego session Store interface,and implements this method in all of the implementors


beego_test.go
```diff
@@ -0,0 +1,9 @@
+package beego
+
+import (
+	"testing"
+)
+
+func TestBeego(t *testing.T) {
+	Run()
```
astaxie: Why add this testing?

PR #3134, opened 2018-04-25
Replace map+metux with sync.map in log.go

TaoSunkist:
Develop: Replace map+metux with sync.map in log.go


astaxie:
sync.Map only support from Go1.9

PR #3141, opened 2018-04-28
Improve access log

gadelkareem:
- Allow access log regardless of the log level.
- Log errors in access log (status: 404, 500, ...etc).
- Make static request logging optional "EnableStaticLogs" (default no logging).
- Log redirects (301, 302) and abort after redirect

### DIFF
--- a/controller.go
+++ b/controller.go
@@ -272,7 +272,9 @@ func (c *Controller) viewPath() string {
 
 // Redirect sends the redirection response to url with status code.
 func (c *Controller) Redirect(url string, code int) {
+	logAccess(c.Ctx, nil, code)
 	c.Ctx.Redirect(code, url)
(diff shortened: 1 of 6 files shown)


controller.go
```diff
@@ -272,7 +272,9 @@ func (c *Controller) viewPath() string {
 
 // Redirect sends the redirection response to url with status code.
 func (c *Controller) Redirect(url string, code int) {
+	logAccess(c.Ctx, nil, code)
 	c.Ctx.Redirect(code, url)
+	panic(ErrAbort)
```
astaxie: Why add panic here?

gadelkareem: To stop the request and send a redirect header so it works the same as Abort() function. It rarely happens that the user would need to continue with the connection.

PR #3142, opened 2018-04-29
Fix response http code

amitash1109:


amitash1109:
Http status code is converted to 200 upon numerous failures like conversion failure from http json in body to parameter struct

PR #3149, opened 2018-05-04
Support DB.BeginTx in go 1.8

liaoishere:
This patch fixes some bugs and add a new feature for supporting `DB.BeginTx` in go 1.8

bug fixes:

- identifiers in `InsertOrUpdate` must be quoted cause in some databases it's  not case-sensitive for identifier
- skip some test in `InsertOrUpdate` for postgres cause it's not support
- fix `go vet` error in `orm/model_test.go`, `go vet` is called by default when using `go test` in go 1.10

new feature:

- support `DB.BeginTx` in go 1.8

so we can use BeginTx to start a transaction with specified isolation level, e.g.

```
orm.BeginTx(context.Background(), &sql.TxOptions{Isolation: sql.LevelRepeatableRead})
```

or start a transaction with context

```
ctx, cancel := context.WithCancel(context.Background())
o.BeginTx(ctx, nil)
```


orm/orm_go18.go
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// +build go1.8
```
astaxie: I don't think we need to rename the file. For travis, it's already change to go1.9

liaoishere: ok, I've pushed a commit to rename it back

astaxie:
--- FAIL: TestAll (0.00s)
	orm_test.go:169: 	
		Error: <QuerySeter> no row found
		orm_test.go:TestAll:1014: 
		     1005:       throwFailNow(t, AssertIs(users2[0].ID, 0))
		     1006:       throwFailNow(t, AssertIs(users2[1].ID, 0))
		     1007:       throwFailNow(t, AssertIs(users2[2].ID, 0))
		     1008:       throwFailNow(t, AssertIs(users2[0].Profile == nil, false))
		     1009:       throwFailNow(t, AssertIs(users2[1].Profile == nil, false))
		     1010:       throwFailNow(t, AssertIs(users2[2].Profile == nil, true))
		     1011:   
		     1012:       qs = dORM.QueryTable("user")
		     1013:       num, err = qs.Filter("user_name", "nothing").All(&users)
		 >>  1014:       throwFailNow(t, err)

liaoishere:
@astaxie this error in test is caused by a recent merged PR https://github.com/astaxie/beego/pull/3247，which changed the returned error when no row found in `All()` but didn't update the test cases.

I'll try to rebase my commits on top of `develop` branch and fix these issues.

liaoishere:
@astaxie 
Now all test looks fine except test sqlite3 on Go 1.9.2. And I can't reproduce it on my local environment.  
Also, it looks irrelevant to this PR.

astaxie:
@liaoishere thanks for the update. I also fixed in develop. but forget to push. I will fix the conflict.

astaxie:
It looks this PR cause the sqlite3 on Go 1.9.2 failed. I think we need to fix it and then merge it

liaoishere:
@astaxie Bugs in `database/sql` on go 1.9.2 lead to the test failed. The connection won't be released even when the context canceled in a transaction. So I skipped the context transaction test on sqlite3.

liaoishere:
Maybe we can add an FAQ section on README to explain this pitfall or others like this.

Issue #2973, opened 2017-12-11
unexpected EOF  session


Liberxue:
```
2017/12/11 17:55:12 [D] [server.go:2619] |      127.0.0.1| 404 |    581.155µs| nomatch| GET      /static/uploadfile/usercard//68420f9cdae4710cb3a04b25aa5fda8b.png
2017/12/11 17:55:12 [D] [server.go:2619] |      127.0.0.1| 404 |    544.023µs| nomatch| GET      /static/uploadfile/usercard//c7f827033ca043a64da399206b65f85d.png
2017/12/11 17:55:12 [D] [server.go:2619] |      127.0.0.1| 404 |    389.888µs| nomatch| GET      /static/uploadfile/usercard//3e20bd59694691cd072a86900af41cf9.png
2017/12/11 17:55:12 [D] [server.go:2619] |      127.0.0.1| 404 |    387.176µs| nomatch| GET      /static/uploadfile/usercard//2a4504b28595b0e5c8f1df29c2ae57e1.png
2017/12/11 17:55:12 [D] [server.go:2619] |      127.0.0.1| 404 |    513.954µs| nomatch| GET      /static/uploadfile/2017-12/11/40713dd72c6f7491af028ebb27cfd5f9-cropper.jpg
2017/12/11 17:55:12 [D] [server.go:2619] |      127.0.0.1| 404 |     393.69µs| nomatch| GET      /static/uploadfile/usercard//c7f827033ca043a64da399206b65f85d.png
2017/12/11 17:55:12 [E] [server.go:2619] unexpected EOF
2017/12/11 17:55:12 [D] [server.go:2619] |      127.0.0.1| 503 |   2.142235ms| nomatch| GET      /static/uploadfile/usercard//68420f9cdae4710cb3a04b25aa5fda8b.png
2017/12/11 17:55:12 [D] [server.go:2619] |      127.0.0.1| 404 |    624.864µs| nomatch| GET      /static/uploadfile/usercard//3e20bd59694691cd072a86900af41cf9.png
2017/12/11 17:55:12 [D] [server.go:2619] |      127.0.0.1| 404 |    559.759µs| nomatch| GET      /static/uploadfile/usercard//2a4504b28595b0e5c8f1df29c2ae57e1.png
```

``go env``
```
'GOARCH="amd64"
GOBIN="/Users/liber/golang/bin"
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOOS="darwin"
[...]

PR #3152, opened 2018-05-07
[#2973] Fix Unexpected EOF bug in staticfile

ayadav:


JessonChan:
@ayadav this commit solve your problems? why?

ayadav:
> @ayadav this commit solve your problems? why?

@JessonChan Earlier it was caching byte.reader in staticFileMap that caused issues with concurrent reads. This fix caches bytes data and returns new byte reader for each read request.

JessonChan:
@ayadav thanks for your reply. i will try to repeat this bug , i am not sure with the EOF bug and #2973

PR #3227, opened 2018-07-02
Adds the evaluated object to the validator struct

pbarnum:
This addition allows users to create custom validation between multiple data fields (example provided below). If this feature is already available, please let me know!

Example:
```
// MatchPassword will test to see if the confirm password field matches the "password" field.
func MatchPassword(v *validation.Validation, obj interface{}, key string) {
	fieldName := strings.Split(key, ".")[0]
	confirmPassword, ok := obj.(string)
	if !ok {
		v.SetError(fieldName, "Password is not a valid string!")
		return
	}

	valueOfObj := reflect.ValueOf(v.FormObj)
	if valueOfObj.Type().Kind() != reflect.Ptr {
		valueOfObj = reflect.New(reflect.TypeOf(v.FormObj))
	}
	password := valueOfObj.Elem().FieldByName("Password")

	if !password.IsValid() || confirmPassword != password.String() {
		v.SetError(fieldName, "Password confirmation does not match!")
	}
}
```

### DIFF
--- a/validation/validation.go
+++ b/validation/validation.go
@@ -111,6 +111,9 @@ type Validation struct {
 	// it will skip those valid functions, see CanSkipFuncs
 	RequiredFirst bool
 
+	// The object being evaluated (set in Valid())
+	FormObj interface{}
+
 	Errors    []*Error
 	ErrorsMap map[string][]*Error
 }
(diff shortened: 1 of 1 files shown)


pbarnum:
@astaxie Is there a way I can access other form inputs when evaluating another form input? Or would you perhaps like to add this feature, just using a different way than I provided?

PR #3239, opened 2018-07-12
add session redis IdleTimeout config

wilhelmguo:


wilhelmguo:
@astaxie  pushed to develop

PR #3247, opened 2018-07-19
send ErrNoRows if the query returns zero rows ... in method orm_query…

mohan2808:
…set.All()


kamilgregorczyk:
This is ridiculous, this change breaks your tests, makes no sense and you merge it anyway?

peter-kyu:
omg... it breaks  backward compatibility...ㅜㅠ

PR #3269, opened 2018-07-31
hible

514366607:
cache add function
// IncrBy increase counter by num.
IncrBy(key string, num int)
// DecrBy decrease counter by num.
DecrBy(key string, num int)


astaxie:
please send the PR to develop branch. I revert the PR

PR #3278, opened 2018-08-03
Revert "send ErrNoRows if the query returns zero rows ... in method orm_query…"

hurisheng:
Reverts astaxie/beego#3247


hurisheng:
It doesn't make sense to return ErrNoRows at the same time allocate space for the input *[]T type parameter

astaxie:
please fix the test case

hurisheng:
I have fixed the test case, please merge this into master ASAP since it brokes old version code, and the behavior is inconsistent with database/sql package Rows. Rows will not return any error if not data found with Query()

PR #3289, opened 2018-08-12
Remove panic from Redirect()

nezorflame:
This `panic(ErrAbort)` is unnecessary in `Redirect` function and causes problems in the production code.


nezorflame:
Builds should pass, can't trigger restart though.

PR #3304, opened 2018-08-29
Go by default initializes the type `string =""` (empty string).

Haxandmat:
Go by default initializes the type `TplName=""` (empty string).
```
type Controller struct {
  ...
  // template data
  TplName        string // =""
  ...
}
```

These actions also nullify the result of the work:
```
// router.go
beego.Router("/foo", &controllers.BaseController{
	Controller: beego.Controller{
		TplName: "foo.html",
	},
})
```
Therefore, you must delete the rows
```
c.Layout = ""
c.TplName = ""
```


astaxie:
This action is just for some controller reused. So init need to reset the value to empty

PR #3317, opened 2018-09-07
Lock

HSoshiant:
Adding a lock for session read
returning the provider for requesting its function directly such as delete file!

### DIFF
--- a/session/session.go
+++ b/session/session.go
@@ -16,7 +16,7 @@
 //
 // Usage:
 // import(
-//   "github.com/astaxie/beego/session"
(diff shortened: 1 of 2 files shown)


session/session.go
```diff
@@ -16,7 +16,7 @@
 //
 // Usage:
 // import(
-//   "github.com/astaxie/beego/session"
+//   "github.com/hsoshiant/beego/session"
```
astaxie: please fix this

astaxie:
@HSoshiant please fix the comments

PR #3318, opened 2018-09-07
when the written message maxSizeCurSize is greater than MaxSize, the …

tiechui1994:
…file needs to be sharded first


astaxie:
I don't think so. If `size >w.MaxSize` will never write success and entry into endless loop

PR #3326, opened 2018-09-17
172,199开始的号段已经出来

jimmyduang:
172,199开始的号段已经出来


astaxie:
validation_test.go:257: "19800008888" is a valid mobile phone number should be false

PR #3382, opened 2018-11-06
Add custom fs template

lxShaDoWxl:
Added the ability to change the way views are stored.


fs.go
```diff
@@ -0,0 +1,73 @@
...
+func walk(fs http.FileSystem, path string, info os.FileInfo, walkFn filepath.WalkFunc) error {
+	var err error
+	if !info.IsDir() {
+		return walkFn(path, info, nil)
+	}
+
+	dir, err := fs.Open(path)
+	defer dir.Close()
+	if err != nil {
+		err1 := walkFn(path, info, err)
```
amrfaissal: This error can be handled on the same line:
```suggestion
		if err1 := walkFn(path, info, err); err1 != nil {
		       return err1
		}
```

fs.go
```diff
@@ -0,0 +1,73 @@
...
+		// The caller's behavior is controlled by the return value, which is decided
+		// by walkFn. walkFn may ignore err and return nil.
+		// If walkFn returns SkipDir, it will be handled by the caller.
+		// So walk should return whatever walkFn returns.
+		return err1
+	}
+
+	for _, fileInfo := range dirs {
+		filename := filepath.Join(path, fileInfo.Name())
+		err = walk(fs, filename, fileInfo, walkFn)
```
amrfaissal: same as above for error handling

template.go
```diff
@@ -310,20 +320,30 @@ func _getTemplate(t0 *template.Template, root string, subMods [][]string, others
...
-				data, err = ioutil.ReadFile(fileAbsPath)
+				f, err := fs.Open(fileAbsPath)
+				if err != nil {
+					f.Close()
+					continue
+				}
+				data, err = ioutil.ReadAll(f)
+				f.Close()
 				if err != nil {
+					logs.Trace("template parse file err:", err)
```
amrfaissal: Can you please use more descriptif error messages:
```suggestion
					logs.Trace("template file parse error:", err)
```

astaxie:
@lxShaDoWxl Could you provide an example for how to use this feature. I like this feature.

lxShaDoWxl:
@astaxie Yes of course. Interested in precisely with go-bindata ?
there is an example in tests, also in testdata there is a build_view command in Makefile.
Make custom fs  and store files anywhere and however I used it with https://github.com/minio/sio to encrypt a large number of views that were unnecessarily placed in memory (requirements for a closed application). 
Enough to do:
```
type CustomFileSystem struct {
}

func (d CustomFileSystem) Open(name string) (http.File, error) {
	return //Your implementation of reading the file
}
```
and call
```
SetTemplateFSFunc(func() http.FileSystem {
		return CustomFileSystem{}
})
```

PR #3392, opened 2018-11-14
fixed argument parser

land-pack:
If user add params at the end of url with `/demo/api?name=frank`. When we try to parser the url by use `GetString`. and we got `string=frank?name=frank`. so I guess we need to just parser it to `frank`.

### DIFF
--- a/controller.go
+++ b/controller.go
@@ -296,7 +296,7 @@ func (c *Controller) Redirect(url string, code int) {
 	c.Ctx.Redirect(code, url)
 }
 
-// SetData set the data depending on the accepted
+// Set the data depending on the accepted
 func (c *Controller) SetData(data interface{}) {
 	accept := c.Ctx.Input.Header("Accept")
 	switch accept {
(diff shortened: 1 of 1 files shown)


controller.go
```diff
@@ -296,7 +296,7 @@ func (c *Controller) Redirect(url string, code int) {
 	c.Ctx.Redirect(code, url)
 }
 
-// SetData set the data depending on the accepted
+// Set the data depending on the accepted
```
astaxie: please fix this

controller.go
```diff
@@ -370,7 +370,7 @@ func (c *Controller) ServeXML() {
 	c.Ctx.Output.XML(c.Data["xml"], hasIndent)
 }
 
-// ServeYAML sends yaml response.
+// ServeXML sends xml response.
```
astaxie: please fix this

astaxie:
How do you use the `GetString`?

land-pack:
> How do you use the `GetString`?

Like What I said in my commit message.

controller.go
astaxie: ?

astaxie:
test failed

flycash:
it was suspended for a long time. I closed it.

PR #3405, opened 2018-11-19
Add .Elapsed in context.ResponseWriter for monitor purpose

coldnight:
With this commit we can record per requests's elapsed time, so we can easy to monitor that by use a filter.


coldnight:
@astaxie

PR #3408, opened 2018-11-20
add different column name parse strategy

nlimpid:
the default parse column function parse acronym such as "ID" and "URL" to "i_d" and "u_r_l", so add another parse strategy to handle acronym properly.

By the way, the original intention of this PR is that when you using "Id", the go fmt command will warning you to change it to "ID".


JIEXINMING:
学到

PR #3410, opened 2018-11-21
fix QueryRow with oracle

jasonerliu:
convert the field name returned by Oracle to lowercase so that QueryRow can use the structure


flycash:
I don't think we need to do this. Users may be confused about this feature

PR #3425, opened 2018-12-02
Need a func to Register newFileWriter. The name is lower now.

xujunfei:
If I want to write logs to more then two files, how can it work?
The func name is lower now and can not Register the newFileWriter.


astaxie:
please use multifile

PR #3434, opened 2018-12-06
Add support for field of type sql.NullXxx in rawSet.setFieldValue()

### LINKED ISSUE #3426: RawSeter.QueryRows() doesn't support field of type  sql.NullXxx while QuerySeter.All() does
```
______
| ___ \
| |_/ /  ___   ___
| ___ \ / _ \ / _ \
| |_/ /|  __/|  __/
\____/  \___| \___| v1.10.0

├── Beego     : 1.11.1
├── GoVersion : go1.11.2
├── GOOS      : windows
├── GOARCH    : amd64
├── Compiler  : gc
```

3. What did you do?
Query data from database using `QuerySeter.All()` and `RawSeter.QueryRows()`.

Both methods should suport sql.NullXxx type of a struct field.

5. What did you see instead?
When dealing with field of type sql.NullXxx, `QuerySeter.All()` would return correct value wrapped in sql.NullXxx while  `RawSeter.QueryRows()` always return zero value of that type.

Is there some reason that lead to the different behaviors?

zav8:
fix #3426

### DIFF
--- a/orm/models_info_f.go
+++ b/orm/models_info_f.go
@@ -301,7 +301,7 @@ checkType:
 	fi.sf = sf
 	fi.fullName = mi.fullName + mName + "." + sf.Name
 
-	fi.description = sf.Tag.Get("description")
+	fi.description = tags["description"]
 	fi.null = attrs["null"]
 	fi.index = attrs["index"]
 	fi.auto = attrs["auto"]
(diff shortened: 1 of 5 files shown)


astaxie:
Could you add the test case for null?

zav8:
> Could you add the test case for null?

@astaxie 
Added now, hope you'll review soon 😄

PR #3441, opened 2018-12-11
hotfix/error parse

GNURub:
Solved error to parse struct with ServeFormatted method: json: unsupported type: map[interface {}]interface {}

### DIFF
--- a/controller.go
+++ b/controller.go
@@ -350,11 +350,8 @@ func (c *Controller) URLFor(endpoint string, values ...interface{}) string {
 
 // ServeJSON sends a json response with encoding charset.
 func (c *Controller) ServeJSON(encoding ...bool) {
-	var (
-		hasIndent   = BConfig.RunMode != PROD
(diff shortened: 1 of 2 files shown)


controller.go
```diff
@@ -377,9 +374,15 @@ func (c *Controller) ServeYAML() {
 
 // ServeFormatted serve YAML, XML OR JSON, depending on the value of the Accept header
 func (c *Controller) ServeFormatted(encoding ...bool) {
-	hasIndent := BConfig.RunMode != PROD
-	hasEncoding := len(encoding) > 0 && encoding[0]
-	c.Ctx.Output.ServeFormatted(c.Data, hasIndent, hasEncoding)
```
astaxie: Why not to improve `c.Ctx.Output.ServeFormatted`. All these function implement will move to ctx.

GNURub: It is a hotfix because in the current version it is not working correctly. I will work on the improvement.

PR #3486, opened 2019-01-11
utils: fix always-false conditions in mail.go

Quasilyte:
The `len(args) < 1 && len(args) > 2` condition is always false.

Found using gocritic linter (badCond check).

### DIFF
--- a/utils/mail.go
+++ b/utils/mail.go
@@ -162,7 +162,7 @@ func (e *Email) Bytes() ([]byte, error) {
 
 // AttachFile Add attach file to the send mail
 func (e *Email) AttachFile(args ...string) (a *Attachment, err error) {
-	if len(args) < 1 && len(args) > 2 {
+	if len(args) < 1 || len(args) > 2 {
 		err = errors.New("Must specify a file name and number of parameters can not exceed at least two")
 		return
 	}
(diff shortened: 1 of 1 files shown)


utils/mail.go
```diff
@@ -162,7 +162,7 @@ func (e *Email) Bytes() ([]byte, error) {
 
 // AttachFile Add attach file to the send mail
 func (e *Email) AttachFile(args ...string) (a *Attachment, err error) {
-	if len(args) < 1 && len(args) > 2 {
+	if len(args) < 1 || len(args) > 2 {
```
amrfaissal: Since `args` here is expecting exactly two parameters, you could drop the check and explicitly specify the accepted arguments. Example:
```golang
 func (e *Email) AttachFile(name string, id string) (a *Attachment, err error) {
        f, err := os.Open(filename)
	if err != nil {
		return
	}
	ct := mime.TypeByExtension(filepath.Ext(filename))
	basename := path.Base(filename)
	return e.Attach(f, basename, ct, id)
}
```

utils/mail.go
```diff
@@ -183,7 +183,7 @@ func (e *Email) AttachFile(args ...string) (a *Attachment, err error) {
 // Attach is used to attach content from an io.Reader to the email.
 // Parameters include an io.Reader, the desired filename for the attachment, and the Content-Type.
 func (e *Email) Attach(r io.Reader, filename string, args ...string) (a *Attachment, err error) {
-	if len(args) < 1 && len(args) > 2 {
+	if len(args) < 1 || len(args) > 2 {
```
amrfaissal: Same as above!

amrfaissal: Also, can you please change `func NewEMail(config string) *Email{}` to `func NewEmail(config string) *Email{}`.

Quasilyte:
I think those changes will change package public API.
I lack context of this project, so I'll leave those to someone else.
My goal was to show that there was an error, "always false" condition.
Thanks.

PR #3493, opened 2019-01-17
Add DBStats method wrapper to provide sql.DBStats when using ormer

bharat-p:


bharat-p:
Travis CI tests are failing because of `gosimple` being deprecated

PR #3494, opened 2019-01-18
simplify beego grace with http.Shutdown

nuczzz:
#### Attention
Go 1.8 adds support for contexts in more parts of the standard library, including the database/sql and net packages and Server.Shutdown in the net/http package[[reference](https://blog.golang.org/go1.8)].

#### Test
step1: write main.go like this, build and run it  in the first terminal.
```
package main

import (
	"fmt"
	"net/http"
	"os"
	"time"

	"github.com/nuczzz/beego/grace"
)

type Handler struct{}

func (h *Handler) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
	duration, err := time.ParseDuration(req.FormValue("duration"))
	if err != nil {
		http.Error(resp, err.Error(), 400)
		return
	}
	time.Sleep(duration)
	fmt.Fprintf(resp, "server [%d]: hello\n", os.Getpid())
}

func main() {
	grace.ListenAndServe(":8080", &Handler{})
}
```
Console will output something like this:
```
2019/01/19 04:27:37 2394 :8080
```
step2: In a second terminal start a slow HTTP request.
```
curl 'http://localhost:8080/?duration=20s'
```
step3: kill process with -SIGHUP.
```
kill -SIGHUP $pid
```
step4: Trigger another shorter request that finishes before the earlier request.
```
curl 'http://localhost:8080/?duration=0s'
```


nuczzz:
Sorry, TLS SetKeepAlive and SetKeepAlivePeriod was missing in my pr, I'll update it later...

PR #3495, opened 2019-01-20
Changes to allow a model to implement a method to return true|flase d…

asif-fndr:
For most practical cases we run into cases when we want to add some models only to some specific database. For example in a multi tenant scenario, there could be an admin database for administrative APIs and will have domain objects pertinent to administration, those models should not be sync'd to tenant databases.

✅ Allow for implementing a method `isTableForDB` for any model.
✅ isTableForDB will receive the database name and will be able to return whether or not this model should be added to that database

Example
```
type SomeModel struct {
  ---- some fields
}

func (this *SomeModel) IsTableForDB(dbname string) bool {
	if dbname == "AdminDB" {
		return true
	}
	return false
}

```

### DIFF
--- a/orm/cmd.go
+++ b/orm/cmd.go
@@ -135,6 +135,10 @@ func (d *commandSyncDb) Run() error {
 	}
 
 	for i, mi := range modelCache.allOrdered() {
+                if !isTableForDB(mi.addrField, d.al.Name) {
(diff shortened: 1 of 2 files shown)


asif-fndr:
@astaxie pls review.

asif-fndr:
@astaxie can we merge this PR?

flycash:
Thank you, I moved those changes to #4211 and add unit tests.

PR #3496, opened 2019-01-20
Foreign key enforcement added on database level

asif-fndr:
Before this fix: there was no database level foreign key enforcement happening even if we create a reference using `orm:"rel(fk);null"`
After this fix: the `orm:"rel(fk);null"` adds the foreign key enforcement on database level.

### DIFF
--- a/orm/utils.go
+++ b/orm/utils.go
@@ -257,6 +257,13 @@ func SetNameStrategy(s string) {
 	nameStrategy = s
 }
 
+// converst *pkg1.pkg2.StructName to StructName
+func stripPrefixes(in string) string  {
+	in = strings.Trim(in, "*")
+	toks := strings.Split(in, ".")
+	return toks[len(toks)-1]
+}
+
 // camel string, xx_yy to XxYy
 func camelString(s string) string {
 	data := make([]byte, 0, len(s))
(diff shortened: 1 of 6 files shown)


asif-fndr:
@astaxie pls review

astaxie:
Could you add test case for this change

astaxie:
@asif-fndr Could you add the test case for it?

astaxie:
@asif-fndr Could you add the test for it? I will release a new version. If you can add the test case before that release would be great.

asif-fndr:
@astaxie added unit test . :56f97866c0c996dc6697f00b73cd11c60aebf273
thought i noticed that there are issues in general in orm_test.go; not clear how we ensure that all models are registered before running UTs and thats why it fails to create tables.

orm/utils.go
```diff
@@ -257,6 +257,13 @@ func SetNameStrategy(s string) {
 	nameStrategy = s
 }
 
+// converst *pkg1.pkg2.StructName to StructName
```
gogitops, line 260: warning: "converst" is a misspelling of "converts" (misspell)

flycash:
It's inactive for a long time. I closed it. If necessary, we could move it to develop-2.0.

PR #3503, opened 2019-01-24
fix: when parse post form it didnt parse fields correctly

wtospit:
When multiple fields which have same name is sent from browser but the first index is empty,
When parse form, Currently It only checked whether the first index is empty, 
Unfortunately the first index is empty but the rest is not (["", "second", "third"])
so this PR fixed this issue by adding logic for a slice checking.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,9 +2,9 @@ module github.com/astaxie/beego
 
 require (
 	github.com/Knetic/govaluate v3.0.0+incompatible // indirect
+	github.com/OwnLocal/goes v1.0.0
 	github.com/beego/goyaml2 v0.0.0-20130207012346-5545475820dd
 	github.com/beego/x2j v0.0.0-20131220205130-a0352aadc542
-	github.com/OwnLocal/goes v1.0.0	
 	github.com/bradfitz/gomemcache v0.0.0-20180710155616-bc664df96737
 	github.com/casbin/casbin v1.7.0
 	github.com/cloudflare/golz4 v0.0.0-20150217214814-ef862a3cdc58
--- a/go.sum
+++ b/go.sum
@@ -1,5 +1,6 @@
 github.com/Knetic/govaluate v3.0.0+incompatible h1:7o6+MAPhYTCF0+fdvoz1xDedhRb4f6s9Tn1Tt7/WTEg=
 github.com/Knetic/govaluate v3.0.0+incompatible/go.mod h1:r7JcOSlj0wfOMncg0iLm8Leh48TZaKVeNIfJntJ2wa0=
+github.com/OwnLocal/goes v1.0.0/go.mod h1:8rIFjBGTue3lCU0wplczcUgt9Gxgrkkrw7etMIcn8TM=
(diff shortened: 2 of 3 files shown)


JessonChan:
new TEST file will be ok without the templatefunc.go's change,other words your code may do not work.

wtospit:
> new TEST file will be ok without the templatefunc.go's change,other words your code may do not work.

Will check it, thanks

wtospit:
Sorry, I've missed to add the case the cause an issue, new commit has been pushed

wtospit:
> @wtospit like this,and compack your commits,thanks

Is it OK to parse the slice with all element has an empty value ?
my commit is aware about this, but if you say OK, I will change the code to you suggestion :)

JessonChan:
yes ,it works as well and change only a little code.

wtospit:
> yes ,it works as well and change only a little code.

Updated !!.  Thanks. 😁

templatefunc.go
```diff
@@ -85,24 +85,24 @@ func DateFormat(t time.Time, layout string) (datestring string) {
...
-	"y", "06",   //A two digit representation of a year   Examples: 99 or 03
+	"y", "06", //A two digit representation of a year   Examples: 99 or 03
 
 	// month
-	"m", "01",      // Numeric representation of a month, with leading zeros 01 through 12
-	"n", "1",       // Numeric representation of a month, without leading zeros   1 through 12
-	"M", "Jan",     // A short textual representation of a month, three letters Jan through Dec
+	"m", "01", // Numeric representation of a month, with leading zeros 01 through 12
+	"n", "1", // Numeric representation of a month, without leading zeros   1 through 12
+	"M", "Jan", // A short textual representation of a month, three letters Jan through Dec
```
astaxie: Please format the comments. The diff shows you doesn't do it.

go.mod
```diff
@@ -2,9 +2,9 @@ module github.com/astaxie/beego
 
 require (
 	github.com/Knetic/govaluate v3.0.0+incompatible // indirect
+	github.com/OwnLocal/goes v1.0.0
 	github.com/beego/goyaml2 v0.0.0-20130207012346-5545475820dd
 	github.com/beego/x2j v0.0.0-20131220205130-a0352aadc542
-	github.com/OwnLocal/goes v1.0.0
```
astaxie: why have this change?

go.sum
```diff
@@ -1,5 +1,6 @@
 github.com/Knetic/govaluate v3.0.0+incompatible h1:7o6+MAPhYTCF0+fdvoz1xDedhRb4f6s9Tn1Tt7/WTEg=
 github.com/Knetic/govaluate v3.0.0+incompatible/go.mod h1:r7JcOSlj0wfOMncg0iLm8Leh48TZaKVeNIfJntJ2wa0=
+github.com/OwnLocal/goes v1.0.0/go.mod h1:8rIFjBGTue3lCU0wplczcUgt9Gxgrkkrw7etMIcn8TM=
```
astaxie: ?

wtospit:
Im not sure about the go mod related but I will format the code. thx

wtospit:
@astaxie when I run go fmt, it removes the space between the code and the comment, could you please suggest how can I do format that ?

wtospit:
@astaxie I tried both in command line and vscode

PR #3529, opened 2019-02-12
add ignore auto_now_add field when update

GeorgeXc:
When I update database with ORM module to update a table which has a field with tag `auto_now_add`, I found that `o.Update(u)` will update the `create_time` field which has a tag `"auto_now_add;type(datetime)"`.
I think this is not correct.
So I create this pull request to fix it.
type` UserAbc struct {
 Id int
 Name string
 Age int 
UpdateTime time.Time `orm:"auto_now;type(datetime)"`
CreateTime time.Time`orm:"auto_now_add;type(datetime)"` 
}

### DIFF
--- a/orm/db.go
+++ b/orm/db.go
@@ -631,10 +631,10 @@ func (d *dbBase) Update(q dbQuerier, mi *modelInfo, ind reflect.Value, tz *time.
 	}
 
 	if find {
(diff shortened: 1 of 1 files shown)


GeorgeXc:
close

PR #3530, opened 2019-02-12
[orm] add ignore auto_now_add field when update 

GeorgeXc:
@astaxie 
1. My code is below:
package main

import (
	"fmt"
	"time"

	_ "github.com/go-sql-driver/mysql"

	"github.com/astaxie/beego/orm"
)

type UserAbc struct {
	Id         int
	Name       string
	Age        int
	UpdateTime time.Time `orm:"auto_now;type(datetime)"`
	CreateTime time.Time `orm:"auto_now_add;type(datetime)"`
}

func main() {
	orm.RegisterDriver("mysql", orm.DRMySQL)
	orm.RegisterDataBase("default", "mysql", "test:123456@tcp(127.0.0.1:3306)/testdb?charset=utf8")

	orm.RegisterModel(new(UserAbc))
	orm.RunSyncdb("default", true, true)
	orm.Debug = true
	o := orm.NewOrm()
	u := &UserAbc{
		Name: "aname",
		Age:  11,
	}
	id, err := o.Insert(u)
	if err != nil {
		fmt.Println(err.Error())
		return
	}
	fmt.Println(time.Now())
	time.Sleep(time.Second * 3)
	u.Id = int(id)
	u.Name = "updatename"
	o.Update(u) //which should only update field without tag auto_now_add
}

2. What did I expected ?
I expected the values of  fields name, update_time, age  will be updated
3. What did I get?
The values of fields name, update_time, age and create_time be updated

4. Why I create this pull request?

I add some code to skip the first field with tag "auto_now_add" in the ORM  Update operation.

### DIFF
--- a/orm/db.go
+++ b/orm/db.go
@@ -632,10 +632,12 @@ func (d *dbBase) Update(q dbQuerier, mi *modelInfo, ind reflect.Value, tz *time.
 
 	if find {
 		newSetNames := make([]string, 0, len(setNames)-1)
-		newSetNames = append(setNames[0:index], setNames[index+1:]...)
+		newSetNames = append(newSetNames, setNames[0:index]...)
+		newSetNames = append(newSetNames, setNames[index+1:]...)
 		setNames = newSetNames
 		newSetValues := make([]interface{}, 0, len(setNames)-1)
-		newSetValues = append(setValues[0:index], setValues[index+1:]...)
+		newSetValues = append(newSetValues, setValues[0:index]...)
+		newSetValues = append(newSetValues, setValues[index+1:]...)
 		setValues = newSetValues
 
 	}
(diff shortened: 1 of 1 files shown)


GeorgeXc:
@astaxie

orm/db.go
```diff
@@ -621,6 +621,27 @@ func (d *dbBase) Update(q dbQuerier, mi *modelInfo, ind reflect.Value, tz *time.
...
+			index = i
+			find = true
+		}
+	}
+
+	if find {
+		newSetNames := make([]string, 0, len(setNames)-1)
+		newSetNames = append(newSetNames, setNames[0:index]...)
+		newSetNames = append(newSetNames, setNames[index+1:]...)
+		setNames = newSetNames
```
astaxie: If you want to delete an element from a slice, you just can use `a = append(a[:i], a[i+1:]...)`

astaxie:
@GeorgeXc

orm/db.go
GeorgeXc: thanks ,I have updated.

PR #3552, opened 2019-02-28
avoid a handler crash situation

zav8:
When automatic-parameter-routing is triggered, handler would crush if a required param is not present.

But according to the [docs](https://beego.me/docs/mvc/controller/params.md#automatic-parameter-routing), it should return `400 Bad Request`.

update: when run in prod mode, the client would receive `400 Bad Request` as expected.

### DIFF
--- a/config.go
+++ b/config.go
@@ -31,8 +31,8 @@ import (
 
 // Config is the main struct for BConfig
 type Config struct {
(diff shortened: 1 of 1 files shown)


zav8:
When run in prod mode, the client would receive `400 Bad Request` as expected.

PR #3606, opened 2019-04-11
fix recover panic status code

ganiszulfa:
`ctx.ResponseWriter.WriteHeader()` will have no effect if it's after `showErr()`, the solution is simply to change the order.


flycash:
I think we don't need to do that because showing error is what we want to do.

PR #3611, opened 2019-04-16
[Fix] Fix Read with SQLite not supporting SELECT FOR UPDATE syntax

maxshine:
Hi core maintainers

I hit an issue while using Beego with SQLite as backend data store. I would like contribute my fix to this problem. Please kindly review it.

## Problem Description

As SQLite is a native serialized database, it does not support the syntax of ` SELECT FOR UPDATE`. If the ORM call `Read(.... true)` of **dbBaser** is made, the ORM with SQLite backend would throws error :
```
near "FOR": syntax error
```

## Expected Behavior

Beego ORM could be compatible with SQLite backend with uniform interface of both normal read and read for update

## Fix Details

As stated in SQLite document and answers excerpted from Stackoverflow, SQLite is native a serialized database. The locking is redundant and it drop the support for `SELECT FOR UPDATE`
[SQLite Isolation Document](https://sqlite.org/isolation.html)
[Stackoverflow answer](https://stackoverflow.com/questions/5800133/how-to-enforce-sqlite-select-for-update-transaction-behavior-in-sqlalchemy)

Due to above supporting facts, the fix is simple that **SQLite** dbBaser instance could just override the `Read` from base. The override should ignores the `isForUpdate` param and constantly use `false` to invoke base call instead.

@astaxie would you please review this fix ? 

Thank YOU !


maxshine:
Hi @astaxie 

Would you please kindly review this fix ?

Thank You !

maxshine:
Hi @astaxie 

would you take care of this PR ?

Thanks.

maxshine:
Hi @astaxie @amrfaissal 

could you please review this PR ?

Thanks !

PR #3612, opened 2019-04-16
feature/kubernetes_configMaps

GNURub:
Add support to ConfigMaps


flycash:
It was suspended for a long time. Supporting kubernetes is a good feature but I don't have time to resolve those conficts and validate this PR. I closed it and if necessary, I want to support kubernetes in the future.

PR #3613, opened 2019-04-16
feature/protobuf

GNURub:
Add support to protobuf API


flycash:
We may think about support it in the future, but now it's not very necessary. Thank you.

PR #3642, opened 2019-05-08
Fix BatchUpdate not update the auto_now field

GeorgeXc:
when we do batch update by the code below, it will not update the field with tag `auto_now`.
`
num, err := o.QueryTable("user").Filter("name", "slene").Update(orm.Params{
    "name": "astaxie",
})
fmt.Printf("Affected Num: %s, %s", num, err)
// SET name = "astaixe" WHERE name = "slene"
`
I create this pull request to fix this.


GeorgeXc:
@astaxie

PR #3650, opened 2019-05-13
Returning an error for invalid sid chars

HSoshiant:
Returning an error if there are any invalid chars in sid.


HSoshiant:
- replacing filepder as local variable by FileSessionStore property

PR #3708, opened 2019-07-05
Fixed logs.BeeLogger.SetLogger: open logs/info.log: permission denied

alimy:
* make missed directory in correct defalt permision(0755) when init logger that adaper is file.This patch can fix  ` logs.BeeLogger.SetLogger: open logs/info.log: permission denied`  problem


flycash:
Now the perm is configurable.

PR #3455, opened 2018-12-27
go module path compatible when application outside GOPATH

qxhy123:
when application outside GOPATH and set GO111MODULE to "ON" or "AUTO", Annotation router still could generate

PR #3718, opened 2019-07-12
Fix no commentsRouter generated outside GOPATH.

liran:
#3162 #3455

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -89,8 +89,9 @@ func defaultGOPATH() string {
 }
 
 func GetModule() bool {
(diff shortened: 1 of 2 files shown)


RanQiwu:
Please help to find the strange mistake.

utils/utils.go
```diff
@@ -87,3 +87,27 @@ func defaultGOPATH() string {
 	}
 	return ""
 }
+
+func GetModule() bool {
```
gogitops, line 91: warning: exported function GetModule should have comment or be unexported (golint)

flycash:
We had resolved this problem. Thank you!

PR #3726, opened 2019-07-19
open session set cookiename err :http: named cookie not present

haleyrom:
open session set cookiename err :http: named cookie not present; getSid return nil;

### DIFF
--- a/session/session.go
+++ b/session/session.go
@@ -196,7 +196,7 @@ func (manager *Manager) getSid(r *http.Request) (string, error) {
 			}
 		}
 
-		return sid, nil
+		return sid, errs
 	}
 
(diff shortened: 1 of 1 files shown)


session/session.go
```diff
@@ -196,7 +196,7 @@ func (manager *Manager) getSid(r *http.Request) (string, error) {
 			}
 		}
 
-		return sid, nil
+		return sid, errs
```
flycash, line 199: We can not return errs here since the invocation `sid = r.FormValue(manager.config.CookieName)` maybe successful.

haleyrom: Success is nil. Failure is err

PR #3758, opened 2019-08-12
Change background yellow color on Windows

kotori2:
due to https://github.com/shiena/ansicolor/issues/14


flycash:
Coud you please provide screen shot? I don't know the difference since I don't use Windows.

PR #3771, opened 2019-08-21
Added orm.DefaultDeletePolicy, which can be one of values we can use …

kispi:
There is quite a chance that the row having foreign key can be unexpectedly deleted in current code.

For example,

type Job struct {
    ...
}

type Profile struct {
    Job *Job `orm:"column(job_id);rel(fk);"`
}

If we don't have relationship config in the `profiles` table, Beego will automatically delete the `profile` if the `job` is deleted, which doesn't make sense for above case.
(Should deleting `job` cause deleting the `profile` of user...?)

And nowadays many devs don't put real relation config for foreign keys in the table actually, due to performance issue.

So I suggest below:
1. Let's have
orm.DefaultDeletePolicy: ["cascade", "set_null", "set_default", "do_nothing"]
2. Let's set it's default as "do_nothing" instead of "cascade"

### DIFF
--- a/orm/models_info_f.go
+++ b/orm/models_info_f.go
@@ -349,7 +349,7 @@ checkType:
 			}
 		default:
 			if onDelete == "" {
-				onDelete = DefaultDeletePolicy
+				onDelete = DefaultOdPolicy
 			} else {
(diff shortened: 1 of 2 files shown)


flycash:
we moved orm to pkg/orm. This is not a mandatory change,  so I close it.

PR #3784, opened 2019-08-30
Optimize rawSet.QueryRows to avoid many unnecessary calls to parseStr…

GazeboXu:
Optimize rawSet.QueryRows to avoid many unnecessary calls to parseStructTag, performance and memory usage can be optimized significantly.


wysnxzm:
是一只猫嘛?

flycash:
I moved those changes to #4210, it will be released in v2.0. Thank you!

PR #3798, opened 2019-09-05
Support *time.Time in orm field

cd1989:
Signed-off-by: cd1989 <chende@caicloud.io>

ORM field doesn't support `*time.Time`, if some one defines such type field, timezone won't be properly processed.


flycash:
Actually, time.Time is a complicated topic. We should think twice and add more tests.

PR #3805, opened 2019-09-11
add SetMaxLifetimeConns and rewrite SetMaxOpenConns method without call reflect

LampScript:
beego orm don't provide the SetMaxLifetimeConns functions but provide SetMaxIdleConns and SetMaxOpenConns.While I get bad connections error, I found the other people use GetDB() then set maxLifeConn, so I think we could full wrapped sql.DB and people don't have to care about sql.DB when they use a orm.

### DIFF
--- a/orm/db_alias.go
+++ b/orm/db_alias.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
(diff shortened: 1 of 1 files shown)


flycash:
we resolved this problem. #3985

PR #3863, opened 2019-11-10
Update app.go

tiniyo:
In case of graceful shutdown , beego server never exit. graceful shutdown return nil and endRunning will never become true.


flycash:
similar with #4005

PR #3867, opened 2019-11-19
Abort with the pre-defined status code when handling XSRF error

ywk253100:
As the status codes(422 and 417) are set in the error map, abort with them directly to active the pre-defined error handlers

Signed-off-by: Wenkai Yin <yinw@vmware.com>


ywk253100:
The pre-defined error handlers are here: https://github.com/astaxie/beego/blob/develop/hooks.go#L35

Issue #3871, opened 2019-11-21
使用go module包管理，运行时beego没有生成路由文件的问题


junzizhao8:
Please answer these questions before submitting your issue. Thanks!

1. What version of Go and beego are you using (`bee version`)?
 go version: 1.12.4
beego version : 1.12.0

2. What operating system and processor architecture are you using (`go env`)?
system: windows
architecture: amd64

根目录的router.go文件
Include(cList ...ControllerInterface)函数写死了路径
for _, wg := range wgopath {   
　　wg, _ = filepath.EvalSymlinks(filepath.Join(wg, "src", t.PkgPath()))   
　　if utils.FileExists(wg) {   
　　　　pkgpath = wg   
　　　　break   
　　}   
}
希望支持go module配置

Hareis:
Hello, I ran into a similar problem today. All of a sudden, the normal API interface suddenly went wrong, console output. ```failed to  generate  the doc```

Hareis:
> Hello, I ran into a similar problem today. All of a sudden, the normal API interface suddenly went wrong, console output. `failed to generate the doc`

I solved this problem because I changed the root directory file name of the project so that the module could not be found

avtion:
如果目录命名中包含“_”这个字符好像会出问题

Jackmrzhou:
Same here, I try every possible ways I could found but the router file just can't be generated, which is really annoying. I used gin before, and I really love the way beego managing routers by adding annotations in comments. Hoping support go module soon.

Jackmrzhou:
#3882 Please check my pull request for solving router generation problems.

danvinhe:
1. 注解路由必须设置runmode = dev开发模式
2. 注解路由的项目必须在gopath目录下

PR #3882, opened 2019-12-05
fix: generating router file in go module environment and filename gen…

Jackmrzhou:
fix: generating router file in go module environment and filename generated incorrctly in some situation.
I've test on my project, and it works on Windows10 with go version 1.13.  In general, what I did are two things:
1. Detect whether the the go mod flag is set and treat differently.
2. Using runtime path maybe improper when the application running inside the IDE like goland, because it doesn't put the target in the project directory. Using the package name will avoid this problem.

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -77,9 +77,7 @@ func init() {
 }
 
 func parserPkg(pkgRealpath, pkgpath string) error {
-	rep := strings.NewReplacer("\\", "_", "/", "_", ".", "_")
-	commentFilename, _ = filepath.Rel(AppPath, pkgRealpath)
-	commentFilename = commentPrefix + rep.Replace(commentFilename) + ".go"
+	commentFilename = commentPrefix + strings.Join(strings.Split(pkgpath, "/")[1:], "_") + ".go"
 	if !compareFile(pkgRealpath) {
 		logs.Info(pkgRealpath + " no changed")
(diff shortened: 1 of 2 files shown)


Jackmrzhou:
#3871

parser.go
```diff
@@ -77,9 +77,7 @@ func init() {
 }
 
 func parserPkg(pkgRealpath, pkgpath string) error {
-	rep := strings.NewReplacer("\\", "_", "/", "_", ".", "_")
-	commentFilename, _ = filepath.Rel(AppPath, pkgRealpath)
-	commentFilename = commentPrefix + rep.Replace(commentFilename) + ".go"
+	commentFilename = commentPrefix + strings.Join(strings.Split(pkgpath, "/")[1:], "_") + ".go"
```
astaxie: you need to consider windows, Mac and Linux version

Jackmrzhou: The `commentFilename` does not include application path, and `pkgpath` only includes slashes. So there is no need to concern the cross-platform problem.

litterGuy:
In v1.12.1, the problem still exists. When generating a routing file, there was only one "commentsRouter_.go" file at the end because the routing file name was the same. When will this bug be fixed?

flycash:
#4107

PR #3890, opened 2019-12-12
Leverage bytes pool to reuse bytes

panjf2000:
Signed-off-by: Andy Pan <panjf2000@gmail.com>


panjf2000:
@astaxie 
Please take a look~

panjf2000:
It seems something wrong when getting `github.com/couchbase/go-couchbase` on travisCI,  root cause could be the recent code changes in go-couchbase, so I ran the `go test` with my code changes on my own computer:
WX20191214-232156@2x
which shows a successful pass, please note.

PR #3894, opened 2019-12-21
fixed automatically creat table bug

arrowfeng:
creating table using uppercase prefix, orm.RegisterModelWithPrefix("T_", new(User)) , it is a little problem when I use func o.QueryTable("T_user") , it tell me this table doesn't exist. actually it is in database.


flycash:
Could you provide some codes to reproduces this issue? I can not understand this PR

PR #3929, opened 2020-02-27
add staticfile size limit

hyperq:
Beego will cache static files to staticFileMap, but this will cause the program to occupy a lot of memory and eventually cause the program to crash. 
I increased the value of MaxStaticFileSize to limit the size of the files that can be cached.


flycash:
similar with #3984

PR #3938, opened 2020-02-29
cache_memory_lock_fix_20200229

yingongzi:
when delete item from map, the map should be locked


flycash:
I think we don't need this lock because method `clearItems` already has lock. Even though there is time window between `expiredKeys` invocation and `clearItems` window, it won't be a problem.

PR #3950, opened 2020-03-21
Revert "Add stmt"

GeorgeXc:
Reverts astaxie/beego#3689
It create stmt without closing, if the program is restarted frequently, the mysql server will keep much more unclosing stmt until mysql server report error:  MySQL error: 1461 "Can't create more than max_prepared_stmt_count statements (current value: 16382)"

### DIFF
--- a/orm/db_alias.go
+++ b/orm/db_alias.go
@@ -15,7 +15,6 @@
 package orm
 
 import (
-	"context"
 	"database/sql"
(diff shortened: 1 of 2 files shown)


flycash:
I think we should think about close the stmt rather than reverse #3689

flycash:
#4025

PR #3992, opened 2020-05-15
[Fix-Issue-3991] Fix Read with SQLite not supporting SELECT FOR UPDAT…

### LINKED ISSUE #3991: SQLite not supporting SELECT FOR UPDATE syntax
Hi core maintainers

I hit an issue while using Beego with SQLite as backend data store. I would like contribute my fix to this problem. Please kindly review it.

## Problem Description

As SQLite is a native serialized database, it does not support the syntax of ` SELECT FOR UPDATE`. If the ORM call `Read(.... true)` of **dbBaser** is made, the ORM with SQLite backend would throws error :
```
near "FOR": syntax error
```

## Expected Behavior

Beego ORM could be compatible with SQLite backend with uniform interface of both normal read and read for update

## Suggest Fix

As stated in SQLite document and answers excerpted from Stackoverflow, SQLite is native a serialized database. The locking is redundant and it drop the support for `SELECT FOR UPDATE`
[SQLite Isolation Document](https://sqlite.org/isolation.html)
[Stackoverflow answer](https://stackoverflow.com/questions/5800133/how-to-enforce-sqlite-select-for-update-transaction-behavior-in-sqlalchemy)

maxshine:
Hi

This is to fix #3991 . As stated in this issue, SQLite does not support `SELECT FOR UPDATE`

The fix is simple that **SQLite** dbBaser instance could just override the `Read` from base. The override should ignores the `isForUpdate` param and constantly use `false` to invoke base call instead.

Please kindly review this MR. 

Thanks.

### DIFF
--- a/orm/db_sqlite.go
+++ b/orm/db_sqlite.go
@@ -17,6 +17,8 @@ package orm
 import (
 	"database/sql"
 	"fmt"
+	"reflect"
+	"time"
 )
 
 // sqlite operators.
@@ -65,6 +67,11 @@ type dbBaseSqlite struct {
 }
 
 var _ dbBaser = new(dbBaseSqlite)
+
+// override base db read for update behavior as SQlite does not support syntax
+func (d *dbBaseSqlite) Read(q dbQuerier, mi *modelInfo, ind reflect.Value, tz *time.Location, cols []string, isForUpdate bool) error {
+	return d.dbBase.Read(q, mi, ind, tz, cols, false)
+}
 
 // get sqlite operator.
 func (d *dbBaseSqlite) OperatorSQL(operator string) string {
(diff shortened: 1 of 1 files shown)


orm/db_sqlite.go
```diff
@@ -66,6 +68,11 @@ type dbBaseSqlite struct {
 
 var _ dbBaser = new(dbBaseSqlite)
 
+// override base db read for update behavior as SQlite does not support syntax
+func (d *dbBaseSqlite) Read(q dbQuerier, mi *modelInfo, ind reflect.Value, tz *time.Location, cols []string, isForUpdate bool) error {
+	return d.dbBase.Read(q, mi, ind, tz, cols, false)
```
flycash, line 73: Please log some warning information if the `isForUpdate` is `true`

maxshine: Hi @flycash . I have added the warn message to note user about isForUpdate param on SQLite is ignored and always set as false in calls. Please review. Thanks.

PR #3993, opened 2020-05-16
fix httplib PostFile method

skanger:
PostFile method don't send Content-Length in the head.

It may cause some error  on  other  server.

Use Transfer-Encoding":"chunked" to fixed it ,  or specific the content length.


flycash:
其实使用content-type: chunked有时候会出一些很奇诡的问题。这个我们还需要慎重考虑一下。

flycash:
We had got strange error when we set the "chunked", but don't know why. So we should be more careful about thsi PR

PR #3996, opened 2020-05-18
Fixes #3995 Use handlers with middleware when starting Graceful server

### LINKED ISSUE #3995: net/http Middleware set via RunWithMiddleware or App.Run(middleware) doesn't work when "BConfig.Listen.Graceful" is set to true

> Go version: 1.13
> Beego 1.12.1

GOHOSTARCH="amd64"
GOHOSTOS="darwin"

3. What did you do?
Add a new middleware and start sample app with `beego.RunWithMiddleWares`
```
package main

import (
	"fmt"
	"net/http"

	"github.com/astaxie/beego"
)
type MainController struct {
	beego.Controller
}

func (m *MainController) Get() {
	m.Ctx.WriteString(fmt.Sprintf("hello world"))
}

func main() {
	
	// This is the cause of middleware not working
	beego.BConfig.Listen.Graceful = true
	
	beego.Router("/", &MainController{})
	beego.RunWithMiddleWares(":8084", func(next http.Handler) http.Handler {
		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
			fmt.Printf("log: in  the middleware from: %s\n", r.RequestURI)
			next.ServeHTTP(w, r)
		})
	})
}
```

Expected to see log "in the middleware"

5. What did you see instead?
Log was not printed

Middleware works fine if I remove:
`	beego.BConfig.Listen.Graceful = true`

bharrat:
Fix issue where Middleware set via `RunWithMiddleware` or with `app.Run` were not working if 
`beego.BConfig.Listen.Graceful` is set to `true`

### DIFF
--- a/app.go
+++ b/app.go
@@ -123,7 +123,7 @@ func (app *App) Run(mws ...MiddleWare) {
 					httpsAddr = fmt.Sprintf("%s:%d", BConfig.Listen.HTTPSAddr, BConfig.Listen.HTTPSPort)
 					app.Server.Addr = httpsAddr
 				}
-				server := grace.NewServer(httpsAddr, app.Handlers)
+				server := grace.NewServer(httpsAddr, app.Server.Handler)
 				server.Server.ReadTimeout = app.Server.ReadTimeout
 				server.Server.WriteTimeout = app.Server.WriteTimeout
 				if BConfig.Listen.EnableMutualHTTPS {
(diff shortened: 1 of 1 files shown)


Julio-Guerra:
I was also debugging this issue and I would rather move this line https://github.com/astaxie/beego/pull/3996/files#diff-a66cc59eb6bfe62b7b3b69cd52b84747R104 after the for loop that applies the middlewares so that `app.Server.Handler` and `app.Handlers` are correctly the same value.

flycash:
@bharrat I think @Julio-Guerra 's solution would be better because the `app.Handlers` still point to the wrong handler. Someone may use the wrong `app.Handlers` too.

PR #3998, opened 2020-05-22
for go modules, generate route by `GO111MODULE=on`

guhan121:
when `RunMode == DEV`  use go modules.  

it will goe a panic.
```
panic("you are in dev mode. So please set gopath")
```


guhan121:
@flycash  看看这个PR呗

flycash:
我只能说肉眼我也看不出来

PR #4006, opened 2020-06-06
fix CI and UT

flycash:


logs/es/es.go
```diff
@@ -38,10 +41,13 @@ func (el *esLogger) Init(jsonconfig string) error {
 		return err
 	} else if u.Path == "" {
 		return errors.New("missing prefix")
-	} else if host, port, err := net.SplitHostPort(u.Host); err != nil {
-		return err
 	} else {
-		conn := goes.NewClient(host, port)
+		conn, err := elasticsearch.NewClient(elasticsearch.Config{
+			Addresses: []string{u.Host},
```
guhan121, line 46: must be `el.DSN` => Addresses: []string{el.DSN},
Otherwise, you will got
```
unable to WriteMsg to adapter:es,error:unsupported protocol scheme "localhost"
```

guhan121:
@flycash @askuy 

if only call 
```
err := logs.SetLogger(logs.AdapterEs, `{"dsn":"http://localhost:9200/","level":7}`)
	if err != nil {
		fmt.Println(`err:`, err)
	}
```
you will got a error
```
err: logs: unknown adaptername "es" (forgotten Register?)
```
because
```
func init() {
	logs.Register(logs.AdapterEs, NewES)
}

```
in package `es` not in `logs`

PR #4014, opened 2020-06-16
over size file do not run openFile

jianzhiyao:
@flycash


flycash:
Did you test it end to end?

jianzhiyao:
> Did you test it end to end?

I already did some tests for it
image

flycash:
Could you join this wechat group? And then please add me（大明），I want to introduce you to our maintainer group.
image

PR #4016, opened 2020-06-19
Update redis.go

huija:
use `scan` instead of `keys` to avoid blocking.


flycash:
Please rebase the develop branch and add UT for Scan function

huija:
UT for Scan function has been added

flycash:
#4024

Issue #3873, opened 2019-11-25
The cache behavior of `stmts` for `func getStmt(query string)` may case `memory leak` unexpectedly.


szyhf:
https://github.com/astaxie/beego/blob/1923b8c7677ad80e47f93fe1a0d2cd4118fa82c3/orm/db_alias.go#L121-L137

In most cases, I agree that cache the query make a better performance,  but in our case, the query is very long and build dynamically (even with placeholder its different), the memory keep increase unexpectedly and `go pprof` only can told me the variable of query string is leaking. I spent a lot of  time to find out the final reason and fix it with `RawPrepare`.

Cache is useful to save time for repeatedly query but not every query is repeatedly.

Please considering add a comment on document to warning user like me to use `RawPrepare` instead or add some params to orm to close the cache.

flycash:
#4022 we resolved this problem

PR #3827, opened 2019-10-13
fix bug caused by without close stmt

GeorgeXc:
This commit add a function to fix the bug reported in the pull request https://github.com/astaxie/beego/pull/3825/commits/330e4ef970c2bae454bdb17823c5d7485a461e17.

To make this fix working .
we need add a line before our program exit
`orm.DeregisterAllDatabases()`

Issue #3737, opened 2019-07-24
Error 1461: Can't create more than max_prepared_stmt_count statements


ducnt114:
Please answer these questions before submitting your issue. Thanks!

1. What version of Go and beego are you using (`bee version`)?
go 1.11.5
beego v1.12.0

2. What operating system and processor architecture are you using (`go env`)?
GOARCH="amd64"
GOOS="darwin"

3. What did you do?
Update to beego version v1.12.0 

4. What did you expect to see?

5. What did you see instead?
Get error: `Error 1461: Can't create more than max_prepared_stmt_count statements`
The prepare statement count from database always increase
Check with query
```sql
SELECT sp.thread_id, t.processlist_user user, t.processlist_host host,
sp.count_star - sd.count_star open_com_query_ps,
cp.count_star - cc.count_star open_com_prepare_ps
FROM
( SELECT COUNT_STAR,
THREAD_ID
FROM events_statements_summary_by_thread_by_event_name
WHERE event_name = 'statement/sql/prepare_sql' ) sp
JOIN
( SELECT COUNT_STAR,
THREAD_ID
FROM events_statements_summary_by_thread_by_event_name
WHERE event_name = 'statement/com/Prepare' ) cp
ON (cp.THREAD_ID = sp.THREAD_ID)
JOIN
( SELECT COUNT_STAR,
THREAD_ID
FROM events_statements_summary_by_thread_by_event_name
WHERE event_name = 'statement/sql/dealloc_sql' ) sd
ON (sd.THREAD_ID = sp.THREAD_ID)
JOIN
( SELECT COUNT_STAR,
THREAD_ID
FROM events_statements_summary_by_thread_by_event_name
WHERE event_name = 'statement/com/Close stmt' ) cc
ON (cc.THREAD_ID = sp.THREAD_ID)
JOIN threads t ON (t.thread_id = sp.thread_id)
ORDER BY GREATEST(open_com_query_ps, open_com_prepare_ps) DESC;
```
database: `performance_schema`

Quons:
here is the bug code，the prepare stmt is opened but not off

func (d *DB) getStmt(query string) (*sql.Stmt, error) {
	d.RLock()
	if stmt, ok := d.stmts[query]; ok {
		d.RUnlock()
		return stmt, nil
	}
	d.RUnlock()

	stmt, err := d.Prepare(query)
	if err != nil {
		return nil, err
	}
	d.Lock()
	d.stmts[query] = stmt
[...]

PR #4022, opened 2020-06-21
stmt: fix bug of non concurrent security & add double check

jianzhiyao:


flycash:
#3791
#3737
#3827

szyhf:
Hi, the main reason of `memory leak` I descripted in #3873 is because the `d.stmts[query]` is set by `query`, but my `query` has really large number of version without params, in different stmt template by '?' as SQL Placeholder, and the `d.stmts` cached them all and never release.

flycash:
> Hi, the main reason of `memory leak` I descripted in #3873 is because the `d.stmts[query]` is set by `query`, but my `query` has really large number of version without params, in different stmt template by '?' as SQL Placeholder, and the `d.stmts` cached them all and never release.
> The lock doesn't case the problem.

#4024

flycash:
we will disable cache in v2.0 by default. #4098

Issue #3201, opened 2018-06-13
Cache: redis cache 会复用已经断开的连接


lyfunny:
Please answer these questions before submitting your issue. Thanks!

1. What version of Go and beego are you using (`bee version`)?

├── Beego     : 1.9.2
├── GoVersion : go1.10.2
├── GOOS      : linux
├── GOARCH    : amd64
├── NumCPU    : 4
├── GOPATH    : /mnt/e/soulsocial/gowork
├── GOROOT    : /usr/local/go
├── Compiler  : gc
└── Date      : Wednesday, 13 Jun 2018

2. What operating system and processor architecture are you using (`go env`)?
GOARCH="amd64"
GOBIN="//usr/local/go/bin"
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/user/soulsocial/gowork"
GORACE=""
GOROOT="/usr/local/go"
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build381897661=/tmp/go-build -gno-record-gcc-switches"
CXX="g++"
CGO_ENABLED="1"
PKG_CONFIG="pkg-config"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"

3. What did you do?
If possible, provide a recipe for reproducing the error.
A complete runnable program is good.

使用redis cache,redis-server 已经把不活动连接关了(redis-server 会自动关闭空闲连接)，但是beego这边感知不到，下次使用的时候会取到这条已经关闭的连接，然后发生err=EOF的错误，这次请求就失败了，下个请求才会重新开新连接
4. What did you expect to see?
希望beego能主动关闭以断开的连接，自动处理这种需要重连的情况，不要使业务失败。
```go
rc.p = &redis.Pool{
		MaxIdle:     3,
		IdleTimeout: 180 * time.Second,
		Dial:        dialFunc,
		TestOnBorrow: func(c redis.Conn, t time.Time) error {
			_, err := c.Do("PING")
			return err
		},
	}
```
[...]

PR #4033, opened 2020-06-27
make redis client idle timeout configurable

jianzhiyao:
@flycash 
Close https://github.com/astaxie/beego/issues/3201

### DIFF
--- a/cache/redis/redis.go
+++ b/cache/redis/redis.go
@@ -55,6 +55,9 @@ type Cache struct {
 	key      string
 	password string
 	maxIdle  int
+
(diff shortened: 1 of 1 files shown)


cache/redis/redis.go
```diff
@@ -55,6 +55,9 @@ type Cache struct {
 	key      string
 	password string
 	maxIdle  int
+
+	//the timeout to a value less than the redis server's timeout.
+	timeout  int
```
flycash, line 60: make timeout as time.Duration. and the default value is 180 * time.Second

PR #4047, opened 2020-07-01
WIP:Add transaction clorsure

jianzhiyao:


orm/orm.go
```diff
@@ -478,16 +495,79 @@ func (o *orm) BeginTx(ctx context.Context, opts *sql.TxOptions) error {
...
+		o.db = &dbQuerierInTransaction{tx: tx}
 	}
 	return nil
 }
 
+func (o *orm) Transaction(fc func(orm Ormer) error) error {
+	return o.TransactionWithContext(context.Background(), fc, nil)
+}
+
+func (o *orm) TransactionWithContext(ctx context.Context,fc func(orm Ormer) error, opts *sql.TxOptions) (err error) {
```
flycash, line 507: I think it could be simply:
```go

tx := xxx
err := fc(xxx)
if err == nil {
   tx.commit
} else {
tx.rollback
```
The scope of tx is not out of this function, so it's thread safe.
And we don't need to define `dbQuerierInTransaction`

PR #4051, opened 2020-07-03
Add sleep on reconnect

### LINKED ISSUE #3972: Sleep for a short time when httplib try to re-connect

3. What did you do?

        // retries default value is 0, it will run once.
	// retries equal to -1, it will run forever until success
	// retries is setted, it will retries fixed times.
	for i := 0; b.setting.Retries == -1 || i <= b.setting.Retries; i++ {
		resp, err = client.Do(b.req)
		if err == nil {
			break
		}
                 time.sleep(b.setting.RetriesTimes)
	}

5. What did you see instead?

gmelodie:
Sleep for 400ms every time `DoRequest` fails.

Fixes #3972.

**Obs**: We had doubts about how to implement the testing for the sleep feature. So we shortened a link to https://beego.me/ using [goolnk](https://goolnk.com/) because we know that issuing a request to the shortened link triggers a redirect, which is known to call the [`CheckRedirect`](https://golang.org/pkg/net/http/#Client) function. What we did was to make `CheckRedirect` always return an unknown error which, as per the [documentation of the `client.Do`](https://golang.org/pkg/net/http/#Client.Do) function, gets echoed out. This makes sure the `sleep` will be triggered during the testing. We tried using [httpbin](https://httpbin.org/) to force the redirects but it turns out [there's an issue with that at the moment](https://github.com/postmanlabs/httpbin/issues/617).

Me, @IamCathal and @tayoogunbiyi worked together on this.

### DIFF
--- a/httplib/httplib.go
+++ b/httplib/httplib.go
@@ -144,6 +144,7 @@ type BeegoHTTPSettings struct {
 	Gzip             bool
 	DumpBody         bool
 	Retries          int // if set to -1 means will retry forever
+	RetryDelay       time.Duration
 }
 
 // BeegoHTTPRequest provides more useful methods for requesting one url than http.Request.
@@ -199,6 +200,11 @@ func (b *BeegoHTTPRequest) Debug(isdebug bool) *BeegoHTTPRequest {
 // others means retried times.
 func (b *BeegoHTTPRequest) Retries(times int) *BeegoHTTPRequest {
 	b.setting.Retries = times
+	return b
+}
+
+func (b *BeegoHTTPRequest) RetryDelay(delay time.Duration) *BeegoHTTPRequest {
+	b.setting.RetryDelay = delay
 	return b
 }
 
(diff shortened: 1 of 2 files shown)


httplib/httplib.go
```diff
@@ -512,11 +512,13 @@ func (b *BeegoHTTPRequest) DoRequest() (resp *http.Response, err error) {
 	// retries default value is 0, it will run once.
 	// retries equal to -1, it will run forever until success
 	// retries is setted, it will retries fixed times.
+	// Sleeps for a 400ms inbetween calls to reduce spam
 	for i := 0; b.setting.Retries == -1 || i <= b.setting.Retries; i++ {
 		resp, err = client.Do(b.req)
 		if err == nil {
 			break
 		}
+		time.Sleep(400 * time.Millisecond)
```
flycash, line 521: I think making this configurable would be better.

IamCathal: Sure, do you think a fixed value set by the user would be better or a multiplier i.e 2x the last wait time like the TCP slow start algorithm

flycash: If you can define an interfaces that allow users to implement their strategy, it will be more scalable. But now, I think a fixed value looks good. What do you think?

IamCathal: Yeah a fixed value would do the job. It would only have to trigger this on edge cases where its internet connection so it shouldn't be too much of an important change.

httplib/httplib.go
```diff
@@ -518,7 +524,7 @@ func (b *BeegoHTTPRequest) DoRequest() (resp *http.Response, err error) {
 		if err == nil {
 			break
 		}
-		time.Sleep(400 * time.Millisecond)
+		time.Sleep(time.Duration(b.setting.RetryDelay) * time.Millisecond)
```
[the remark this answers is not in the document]
IamCathal: Yeah I'd agree with that. Should I comment the field in the struct to note that it's in ms?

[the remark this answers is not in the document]
IamCathal: Changed it to `time.Duration` and rebased into one commit

PR #4053, opened 2020-07-04
Add docker-compose to support running unit test locally

flycash:
Now it's hard to start all components, like mysql, to run unit tests locally. 

I write a docker-compose file and test.sh script to reduce the difficulties for developers to run unit tests locally


flycash:
There are som problems in orm_test. We will fix it later.

Issue #3974, opened 2020-04-22
beego healthcheck for kubernates


publicocean0:
the healthcheck is a html page , it is impossible to see a json format

tayoogunbiyi:
Hey @flycash.
I'd like to work on this if that's ok ?

flycash:
Thank you! It's very welcome!

PR #4055, opened 2020-07-06
Allow Healthcheck endpoint return JSON for Kubernetes

tayoogunbiyi:
This pull request allows the `/healthcheck` endpoint to be able to return JSON if a JSON flag is specified as a query string `/healthcheck?json=true`. It also includes unit tests for the changes.

The tests which check the regular healthcheck endpoint without the `json` query string could definitely be improved by more powerful introspection of the template which is generated. Perhaps checking against the classes generated in the HTML template?

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -326,8 +326,8 @@ func healthcheck(rw http.ResponseWriter, r *http.Request) {
 	shouldReturnJSON, _ := strconv.ParseBool(jsonFlag)
 
 	if shouldReturnJSON {
-		responseMap := buildHealthCheckResponseMap(resultList)
-		jsonResponse, err := json.Marshal(responseMap)
+		response := buildHealthCheckResponseList(resultList)
+		jsonResponse, err := json.Marshal(response)
 
 		if err != nil {
 			http.Error(rw, err.Error(), http.StatusInternalServerError)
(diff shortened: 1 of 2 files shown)


tayoogunbiyi:
@flycash This PR is aimed at closing issue https://github.com/astaxie/beego/issues/3974

admin_test.go
```diff
@@ -75,3 +96,81 @@ func oldMap() M {
...
+
+	handler := http.HandlerFunc(healthcheck)
+
+	handler.ServeHTTP(w, req)
+	if status := w.Code; status != http.StatusOK {
+		t.Errorf("handler returned wrong status code: got %v want %v",
+			status, http.StatusOK)
+	}
+
+	expectedResponseBody := `[{"message":"database","name":"success","status":"OK"},{"message":"cache","name":"error","status":"no cache detected"}]`
```
gmelodie, line 171: Nit: Maybe break into multiple lines [like is done here](https://yourbasic.org/golang/json-example/)

admin.go
```diff
@@ -322,12 +320,44 @@ func healthcheck(rw http.ResponseWriter, _ *http.Request) {
...
+	}
+
 	content["Data"] = resultList
 	data["Content"] = content
 	data["Title"] = "Health Check"
+
 	execTpl(rw, data, healthCheckTpl, defaultScriptsTpl)
 }
 
+func execJSON(rw http.ResponseWriter, jsonData []byte) {
```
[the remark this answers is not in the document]
tayoogunbiyi: Main reason for naming it `execJSON` was to follow similar convention (check out `execTpl` which I presume renders a template as a response)

admin.go
```diff
@@ -322,12 +320,44 @@ func healthcheck(rw http.ResponseWriter, _ *http.Request) {
...
+		}
+
+		JSONResponse, err := json.Marshal(response)
+		if err != nil {
+			http.Error(rw, err.Error(), http.StatusInternalServerError)
+		} else {
+			execJSON(rw, JSONResponse)
+		}
+
+		return
```
flycash, line 346: I think extracting this block as a new method looks better. And the `JSONResponse` which is start with upper case character need to be renamed.

admin.go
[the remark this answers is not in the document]
flycash: How about rename both? `writeJson` and `writeTpl`?

admin_test.go
flycash: Formating the json string is more readable.

tayoogunbiyi:
@flycash Some tests seem to fail but they've nothing to do with the current changes

tayoogunbiyi:
> @flycash Some tests seem to fail but they've nothing to do with the current changes

I've also implemented the requested changes. Let know what you think :)

Issue #4040, opened 2020-06-29
support user-define ES index name 


flycash:
Please answer these questions before submitting your issue. Thanks!

#2662 

1. What version of Go and beego are you using (`bee version`)?
Any version

2. What operating system and processor architecture are you using (`go env`)?
Any

3. What did you do?
If possible, provide a recipe for reproducing the error.
A complete runnable program is good.

When we use ES as the logger, the index name looks like "2018-01-12",

but I think we should allow users to custom their own index name.

4. What did you expect to see?

User can define their index name

5. What did you see instead?

Index name looks like "2019-01-23"

741369:
I will fix this issue later and  [issue4008](https://github.com/astaxie/beego/issues/4008).

Issue #4008, opened 2020-06-06
Using official elasticsearch client


flycash:
Beego use [goes](https://github.com/OwnLocal/goes/tree/v1.0.0) as ES client library since there is not any official library.

But now, the official library was provided: [go-elasticsearch](https://github.com/elastic/go-elasticsearch)

We are considering replacing goes with go-elasticsearch. 

For now, the ES is only used in logging, so the goals are:
1. replacing goes with go-elasticsearch;
2. Support ES v5.x, v6.x, v7.x;
3. Beego users can choose the ES version by providing some config

To fix CI problem, I had removed goes library and use ES v6.x official client. But it's a temporary solution. see #4006

guhan121:
@flycash  PR has a bug

in `es.go` `LogDocument` must be blow.
```
type LogDocument struct {
	Timestamp string `json:"timestamp"`
	Msg       string `json:"msg"`
}

```

Otherwise, 
```
body, err := json.Marshal(idx)
	if err != nil {
		return err
	}
fmt.Println(`body:`,string(body))   // {}
```
nothing put to es-server

guhan121:
![image](https://user-images.githubusercontent.com/5517369/85404472-8c7b5600-b591-11ea-9913-debe8483c97d.png)
感觉index自己设置比较好，不然的话岂不是每天都在变 @flycash

flycash:
@lg1024

flycash:
@741369

PR #4057, opened 2020-07-07
Fix user define es index name in logs/es/es.go

741369:
I'm pushing one commits. User can define index name and index prefix name and read beego config appName.

Fixes [#4040](https://github.com/astaxie/beego/issues/4040) and [#4008](https://github.com/astaxie/beego/issues/4008)

### DIFF
--- a/logs/es/es.go
+++ b/logs/es/es.go
@@ -9,6 +9,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/astaxie/beego"
+
 	"github.com/elastic/go-elasticsearch/v6"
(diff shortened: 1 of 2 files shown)


741369:
> This is not scalable. I think we should define an interface looks like:
> 
> ```go
> type IndexNaming interface {
>   GetIndexName(when, msg, level) string 
> }
> ```
> 
> And the default implementation based on prefix will be provided.
> But we allow users to use their own implementation.
> For example, users possibly want to put all error msg into one index.

I think

flycash:
#4233

PR #4058, opened 2020-07-08
Fix response payload too large

### LINKED ISSUE #4054: MaxMemory behavior is different from other frameworks

 1.12.2

 1.14.2

3. What did you do?
Post with a body size larger then `MaxMemory`.

Response http code 413.

5. What did you see instead?
A truncated body is passed.

6. More detail
`MaxMemory` behavior is different from other the framework that I've used, and it doesn't seem to satisfy the [rfc7231#section-6.5.11](https://tools.ietf.org/html/rfc7231#section-6.5.11)
I think this code is problematic [router.go#L743-L748](https://github.com/astaxie/beego/blob/develop/router.go#L743-L748)
```
	if r.Method != http.MethodGet && r.Method != http.MethodHead {
		if BConfig.CopyRequestBody && !context.Input.IsUpload() {
			context.Input.CopyBody(BConfig.MaxMemory)
		}
		context.Input.ParseFormOrMulitForm(BConfig.MaxMemory)
	}
```
We should response error code rather than pass a truncated body. Silent errors make it difficult to troubleshoot errors.
Here is my test
```
// copy to router_test.go and run this test
func TestRouterPostLargeBody(t *testing.T) {
	maxMemory := BConfig.MaxMemory
	BConfig.CopyRequestBody = true
	BConfig.MaxMemory = 20

	type reqBody struct {
		Foo string `json:"foo"`
	}

	originBody := &reqBody{
		Foo: "barbarbarbarbar",
	}

	s, _ := json.Marshal(originBody)
	b := bytes.NewBuffer(s)
	r, _ := http.NewRequest("POST", "/user/123", b)
	w := httptest.NewRecorder()

	handler := NewControllerRegister()
	handler.Post("/user/:id", func(ctx *context.Context) {
		var body reqBody
		if err := json.Unmarshal(ctx.Input.RequestBody, &body); err != nil {
                       // TestRouterPostLargeBody unmarshal body size, recive {"foo":"barbarbarbar
			t.Errorf("TestRouterPostLargeBody unmarshal body size, recive %s", string(ctx.Input.RequestBody))
		}

		if originBody.Foo != body.Foo {
			t.Errorf("TestRouterPostLargeBody request and recive body not equal")
		}
		BConfig.MaxMemory = maxMemory

		var res []byte
		ctx.Output.Body(res)
	})
	handler.ServeHTTP(w, r)
}

```

a631807682:
Close #4054

### DIFF
--- a/error.go
+++ b/error.go
@@ -363,12 +363,13 @@ func gatewayTimeout(rw http.ResponseWriter, r *http.Request) {
 func payloadTooLarge(rw http.ResponseWriter, r *http.Request) {
 	responseError(rw, r,
 		413,
-		"<br>The page you have requested is unavailable."+
-			"<br>Perhaps you are here because:"+
-			"<br><br><ul>"+
-			"<br>The request entity is larger than limits defined by server"+
-			"<br>Please change the request entity and try again."+
-			"</ul>",
+		`<br>The page you have requested is unavailable.
+		 <br>Perhaps you are here because:<br><br>
+		 <ul>
+			<br>The request entity is larger than limits defined by server.
+			<br>Please change the request entity and try again.
+		 </ul>
+		`,
 	)
 }
 
--- a/error.go
+++ b/error.go
@@ -28,7 +28,7 @@ import (
 )
 
 const (
-	errorTypeHandler    = iota
+	errorTypeHandler = iota
 	errorTypeController
 )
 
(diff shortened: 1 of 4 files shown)


router.go
```diff
@@ -707,6 +707,10 @@ func (p *ControllerRegister) ServeHTTP(rw http.ResponseWriter, r *http.Request)
 
 	if r.Method != http.MethodGet && r.Method != http.MethodHead {
 		if BConfig.CopyRequestBody && !context.Input.IsUpload() {
+			if r.ContentLength > BConfig.MaxMemory {
```
flycash, line 710: Could you please add some comments here to tell users this break change? And then log some error information.

error.go
```diff
@@ -359,6 +359,19 @@ func gatewayTimeout(rw http.ResponseWriter, r *http.Request) {
 	)
 }
 
+// show 413 Payload Too Large
+func payloadTooLarge(rw http.ResponseWriter, r *http.Request) {
+	responseError(rw, r,
+		413,
+		"<br>The page you have requested is unavailable."+
```
flycash, line 366: Could you use ` operator to quote this template, like:
```go
`this 
is 
a 
str
`
```
I found that other functions don't use this operator, but we need to fix them in the future.

Issue #3769, opened 2019-08-17
JSON_FORMAT access log should only output json without any prefix


bharat-p:
Please answer these questions before submitting your issue. Thanks!

1. What version of Go and beego are you using (`bee version`)?
Go  = 1.12
Beego = 1.12.0

4. What did you do?
* Set up a simple api using beego
* enable access log by setting: 
`beego.BConfig.Log.AccessLogs = false`
* set access log format to JSON_FORMAT
`beego.BConfig.Log.AccessLogsFormat = "JSON_FORMAT"`

4. What did you expect to see?
I expected beego to output access log as a valid json string
```
{"remote_addr":"::1","request_time":"2019-08-17T14:19:40.952582-07:00","request_method":"GET","request":"GET /test HTTP/1.1","server_protocol":"HTTP/1.1","host":"localhost:8082","status":200,"body_bytes_sent":0,"elapsed_time":1967,"http_referrer":"","http_user_agent":"PostmanRuntime/7.15.2","remote_user":""}
```

5. What did you see instead?

```
2019/08/17 14:19:41.102  {"remote_addr":"::1","request_time":"2019-08-17T14:19:40.952582-07:00","request_method":"GET","request":"GET /test HTTP/1.1","server_protocol":"HTTP/1.1","host":"localhost:8082","status":200,"body_bytes_sent":0,"elapsed_time":1967,"http_referrer":"","http_user_agent":"PostmanRuntime/7.15.2","remote_user":""}
```
Access log is prefixed with string representing log time, log time should also be put inside the json being outputted

IamCathal:
Working on this at the moment

PR #4063, opened 2020-07-08
Fix JSON log formatting

IamCathal:
Fixed https://github.com/astaxie/beego/issues/3769

Now when the `AccessLogsFormat` is set to `JSON_FORMAT` is prints the logs in proper JSON format without time and line number before inplain text.

A full test for this would involve spinning up a new api so I'm not sure it would be the best idea. To test you can run `bee api testapi` and the `main.go` would be 
```go
package main

import (
	_ "testapi/routers"

	"github.com/astaxie/beego"
)

func main() {
	if beego.BConfig.RunMode == "dev" {
		beego.BConfig.WebConfig.DirectoryIndex = true
		beego.BConfig.WebConfig.StaticDir["/swagger"] = "swagger"
	}
	beego.BConfig.Log.AccessLogsFormat = "JSON_FORMAT"
	beego.SetLogger("file", `{"filename":"test.json"}`)

	beego.Run()
}

```

which gives the desired output for requests (this is formatted just to look nice): 
```
{
    "remote_addr": "::1",
    "request_time": "2020-07-08T18:14:03.305436912+01:00",
    "request_method": "GET",
    "request": "GET / HTTP/1.1",
    "server_protocol": "HTTP/1.1",
    "host": "localhost:8080",
    "status": 404,
    "body_bytes_sent": 0,
    "elapsed_time": 243366,
    "http_referrer": "",
    "http_user_agent": "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/83.0.4103.116 Safari/537.36",
    "remote_user": "",
    "line_reference": "server.go:2807"
}
```

I initially tried to add a `logFormat` parameter to the functions to differentiate between the two formats but it wasn't going to work due to complications. Instead what it does it it tries to decode the message into a JSON struct. If it can then it means it is in `JSON_FORMAT` and it goes on as it should. If it cannot decode into a JSON struct then it is in `APACHE_FORMAT`

```go
tempStruct := &AccessLogRecord{}
	jsonDecodeError := json.Unmarshal([]byte(msg), tempStruct)
	if jsonDecodeError == nil {
		logFormat = "JSON_FORMAT"
	}
```

Also to distinguish the two it sets the `when` parameter for `JSON_FORMAT` to an empty time struct (`time.Time{}`). Later before printing to the console it's checked. If [when.isZero()](https://golang.org/pkg/time/#Time.IsZero) it knows it's in `JSON_FORMAT` and it therefore does not print it before.

### DIFF
--- a/logs/file.go
+++ b/logs/file.go
@@ -148,30 +148,42 @@ func (w *fileLogWriter) WriteMsg(when time.Time, msg string, level int) error {
 	if level > w.Level {
 		return nil
 	}
-	hd, d, h := formatTimeHeader(when)
-	msg = string(hd) + msg + "\n"
-	if w.Rotate {
-		w.RLock()
-		if w.needRotateHourly(len(msg), h) {
-			w.RUnlock()
-			w.Lock()
+
+	logFormat := "APACHE_FORMAT"
+	tempStruct := &AccessLogRecord{}
+	jsonDecodeError := json.Unmarshal([]byte(msg), tempStruct)
+	if jsonDecodeError == nil {
+		// If successfully unmarshelled the text, it was a JSON
+		// format, else it was an apache format
+		logFormat = "JSON_FORMAT"
+	}
+
+	if logFormat == "APACHE_FORMAT" {
+		hd, d, h := formatTimeHeader(when)
+		msg = string(hd) + msg + "\n"
+		if w.Rotate {
+			w.RLock()
 			if w.needRotateHourly(len(msg), h) {
-				if err := w.doRotate(when); err != nil {
-					fmt.Fprintf(os.Stderr, "FileLogWriter(%q): %s\n", w.Filename, err)
+				w.RUnlock()
+				w.Lock()
+				if w.needRotateHourly(len(msg), h) {
+					if err := w.doRotate(when); err != nil {
+						fmt.Fprintf(os.Stderr, "FileLogWriter(%q): %s\n", w.Filename, err)
+					}
 				}
-			}
-			w.Unlock()
-		} else if w.needRotateDaily(len(msg), d) {
-			w.RUnlock()
-			w.Lock()
-			if w.needRotateDaily(len(msg), d) {
-				if err := w.doRotate(when); err != nil {
-					fmt.Fprintf(os.Stderr, "FileLogWriter(%q): %s\n", w.Filename, err)
+				w.Unlock()
+			} else if w.needRotateDaily(len(msg), d) {
+				w.RUnlock()
+				w.Lock()
+				if w.needRotateDaily(len(msg), d) {
+					if err := w.doRotate(when); err != nil {
+						fmt.Fprintf(os.Stderr, "FileLogWriter(%q): %s\n", w.Filename, err)
+					}
 				}
+				w.Unlock()
+			} else {
+				w.RUnlock()
 			}
-			w.Unlock()
-		} else {
-			w.RUnlock()
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


flycash:
I will review it this weekend

logs/file.go
```diff
@@ -148,30 +148,42 @@ func (w *fileLogWriter) WriteMsg(when time.Time, msg string, level int) error {
...
+	if jsonDecodeError == nil {
+		// If successfully unmarshelled the text, it was a JSON
+		// format, else it was an apache format
+		logFormat = "JSON_FORMAT"
+	}
+
+	if logFormat == "APACHE_FORMAT" {
+		hd, d, h := formatTimeHeader(when)
+		msg = string(hd) + msg + "\n"
+		if w.Rotate {
```
flycash, line 164: it looks like that we don't do rotate if the format is not "APACHE_FORMAT"?

logs/file.go
```diff
@@ -148,30 +148,42 @@ func (w *fileLogWriter) WriteMsg(when time.Time, msg string, level int) error {
...
 	}
-	hd, d, h := formatTimeHeader(when)
-	msg = string(hd) + msg + "\n"
-	if w.Rotate {
-		w.RLock()
-		if w.needRotateHourly(len(msg), h) {
-			w.RUnlock()
-			w.Lock()
+
+	logFormat := "APACHE_FORMAT"
```
flycash, line 152: If we define a new interface like
```
type LoggerFormatter interface {
    Format(when, msg, level)
}
```
It will be more scalable, but more complicated. What do you think?
Here is a similar issue #4002 
If you don't understand Chinese, just ignore the Chinese comment, and we can discuss here.

logs/file.go
IamCathal: I somehow thought rotating was only needed for `APACHE_FORMAT`. I'll fix that

logs/file.go
IamCathal: I agree. My fix was working but not the cleanest. I'll get working on an implementation that uses a new interface for this as it will benefit anyone looking to implement their own logging format in the future

flycash: #4073 It's a similar issue. So allowing user custom their log info format looks like very necessary.

IamCathal: Alright I think I've made a bit of progress on the issue but I'm no expert so I'd like an opinion on it before I continue to make sure I'm going in the right direction.

So the user would define some sort of function themselves possibly that would be able to manipulate the fields in `AccessLogRecord`. I think these fields would be sufficient for a user to choose from.
https://github.com/astaxie/beego/blob/7a48fbb69846515d9dc3627096590f2b7b2a1bca/logs/accesslog.go#L31-L45

From this function the output would be a single string with the formattted message in whatever way the user wants it to be. Beego could access this function within the [LogAccess](https://github.com/astaxie/beego/blob/7a48fbb69846515d9dc3627096590f2b7b2a1bca/router.go#L1019) func and then pass on the formatted message to  `beeLogger.writeMsg` ([AccessLog](https://github.com/astaxie/beego/blob/7a48fbb69846515d9dc3627096590f2b7b2a1bca/logs/accesslog.go#L65) wouldn't really be needed anymore or we could move my change from `LogAccess` into it)

```go
// LogAccess logging info HTTP Access
func LogAccess(ctx *beecontext.Context, startTime *time.Time, statusCode int) {
	...
	record := &logs.AccessLogRecord{
		...
	}
	logMessage := Format(record)
	logs.AccessLog(logMessage)
}
```

This doesn't include an interface and that's mainly because I've not used them much before but I've done a good bit of reading into them today and I do think they should be used in the actual implementation

flycash: I think what we need to do is defining an API looks like 

`WriteMsg(when time.Time, filePath string, msg string, level int) error`

The key point is `writeMsg` and `writeToLoggers` method of `BeeLogger`.

we need to change the implementation like:

```go
func (bl *BeeLogger) writeMsg(logLevel int, msg string, v ...interface{}) error {
 // ... so somthing 
file :=""

if bl.enableFuncCallDepth {
  // read file full name
}

bl.writeToLoggers(when, msg, logLevel, file)

}


func (bl *BeeLogger) writeToLoggers(when time.Time, msg string, level int, file string) {
	for _, l := range bl.outputs {
		err := l.WriteMsg(when, msg, level, file)
		if err != nil {
			fmt.Fprintf(os.Stderr, "unable to WriteMsg to adapter:%v,error:%v\n", l.name, err)
		}
	}
}
```
And then we change the interface:
```go
type Logger interface {
	Init(config string) error
	WriteMsg(when time.Time, msg string, level int, file string) error
	Destroy()
        SetFormatter(fmt Formatter)
	Flush()
}

type Formatter interface {
    Format(when time.Time, msg string, level int, file string) string
}
```
The implementation of this `Logger#WriteMsg` looks like:
```go
WriteMsg(...) {
    msg = fmtter.Format(...)
    doWrite(msg)
}
```

And we need to add new method:
```go
func SetLogger(adapter string, formatter, config ...string) error {
...
}
```

In this case, JSON log, users may invoke `SeLogger(xxx, jsonFormat, config...)`
```go
type JsonFormatter struct {
}

func (jf JsonFormatter) Format(...) string {
    if isJson(...) {
   } else {
   ...
}
}
```

flycash: we could schdule a zoom meeting to have further discussion if you want.

flycash: It's a big change, we could release this feature in v2.0.1 (next is v2.0.0) since we decide to refactor move log to pkg/log and then refactor logs module. If you want to be responsible for refactoring log module, we can put this in v2.0.0. 

Now we are focusing on refactoring ORM module :)

IamCathal: I think a call would definitely be beneficial. It's quite a large change and I'm not too sure how long it'd take to implement fully or how much I could do myself but I'd be up for giving it my best effort. 

I'll write down some notes as to some of the questions I have and ideas to talk about during a call. I'll email you to schedule that.

flycash:
Here is the meeting minutes:
```
1. define new structure to represents log, file path and linumber, level, time, msg
2. change the WriteMsg or add one method;
3. (bl *BeeLogger) writeToLoggers , (bl *BeeLogger) writeMsg(logLevel int, msg string, v ...interface{})
4. Formetter interface
5. default implementation keeping current behavior, JsonFormtter
6. SetFortter and SetGlobalFormatter

base branch develop-2.0 under pkg dir
```

PR #4065, opened 2020-07-09
update README.md 

Acmefocus:


README.md
```diff
@@ -12,6 +12,15 @@ It is inspired by Tornado, Sinatra and Flask. beego has some Go-specific feature
 
     go get github.com/astaxie/beego
 
+#### Create `hello` directory, cd `hello` directory
```
flycash, line 15: Actually we recommend using `bee create hello`。
Here is the doc https://beego.me/docs/install/bee.md. You can change the language to English if you need it.

And here, the correct order is creating hello  directory, cd , and then go mod init. and then go get beego...

You can try it.

Acmefocus: I didn't find the BEE Create option, just the BEE New option,This command creates only the directory, but does not do module initialization.  There is no `go mod init` here, no   `go mod init`  in bee new too.

flycash: Sorry, it's `bee new`. 
You are right, it doesn't create go mod. so we have two ways to create a new project. Using `bee new` and then `go mod init`. In this way, `bee` will create a standard project structure. Or users create project manually.

PR #4068, opened 2020-07-13
fix `index out of range` when sid len = 1

jianzhiyao:
Close https://github.com/astaxie/beego/issues/3898

### DIFF
--- a/session/sess_file.go
+++ b/session/sess_file.go
@@ -15,11 +15,11 @@
 package session
 
 import (
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
 	"os"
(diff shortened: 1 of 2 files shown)


jianzhiyao:
@flycash

session/sess_file.go
```diff
@@ -180,6 +180,10 @@ func (fp *FileProvider) SessionExist(sid string) bool {
 	filepder.lock.Lock()
 	defer filepder.lock.Unlock()
 
+	if len(sid) < 2 {
+		return false
```
flycash, line 184: Could you please add some log? I think warning looks good.

PR #4072, opened 2020-07-16
refactor orm

jianzhiyao:


pkg/orm/types.go
```diff
@@ -74,63 +82,93 @@ type Ormer interface {
...
+}
+
+type Ormer interface {
+	DQL
+	DML
+	DriverGetter
+	TxBeginner
+}
+
+type TxOrmer interface {
```
flycash, line 165: I think we should allow users to start a new transaction when they use TxOrmer. So that they can execute the transaction inside another transaction.

pkg/orm/types.go
```diff
@@ -35,35 +35,43 @@ type Fielder interface {
...
-	ReadOrCreate(md interface{}, col1 string, cols ...string) (bool, int64, error)
+type TxBeginner interface {
+	//self control transaction
+	Begin() (TxOrmer, error)
+	BeginWithCtx(ctx context.Context) (TxOrmer, error)
+	BeginWithOpts(opts *sql.TxOptions) (TxOrmer, error)
+	BeginWithCtxAndOpts(ctx context.Context, opts *sql.TxOptions) (TxOrmer, error)
+
+	//closure control transaction
+	DoTransaction(task func(txOrm TxOrmer) error) error
```
flycash, line 46: We use `Tx` to represent `Transaction`.

pkg/orm/plugin/cluster/read_write_decorator.go
```diff
@@ -0,0 +1,175 @@
...
+	"database/sql"
+	"github.com/astaxie/beego/pkg/orm"
+)
+
+type SimpleReadWriteOrm struct {
+	master   orm.Ormer
+	follower orm.Ormer
+}
+
+var _ orm.Ormer = new(SimpleReadWriteOrm)
```
flycash, line 28: Those codes are demo. You should not merge those codes to develop-2.0

pkg/orm/orm.go
```diff
@@ -442,105 +478,131 @@ func (o *orm) QueryTable(ptrStructOrTableName interface{}) (qs QuerySeter) {
...
+func (o *orm) DoTransactionWithCtx(ctx context.Context, task func(txOrm TxOrmer) error) error {
+	return o.DoTransactionWithCtxAndOpts(ctx, nil, task)
+}
+
+func (o *orm) DoTransactionWithOpts(opts *sql.TxOptions, task func(txOrm TxOrmer) error) error {
+	return o.DoTransactionWithCtxAndOpts(context.Background(), opts, task)
+}
+
+func (o *orm) DoTransactionWithCtxAndOpts(ctx context.Context, opts *sql.TxOptions, task func(txOrm TxOrmer) error) error {
+	tx, err := o.db.(txer).BeginTx(ctx, opts)
```
flycash, line 549: `tx, err = o.Begin(ctx, opts)`...

pkg/orm/types.go
jianzhiyao: sql.Tx not provide method **Begin**. So i think sholud not provide transaction in TxOrmer

PR #4078, opened 2020-07-19
refactor orm

jianzhiyao:


pkg/orm/orm.go
```diff
@@ -442,105 +479,147 @@ func (o *orm) QueryTable(ptrStructOrTableName interface{}) (qs QuerySeter) {
...
+	panicked = false
 	return err
 }
 
-// return a raw query seter for raw sql string.
-func (o *orm) Raw(query string, args ...interface{}) RawSeter {
-	return newRawSet(o, query, args)
+type txOrm struct {
+	ormBase
+	isClosed   bool
```
flycash, line 576: If you want to check `isClosed`, it means all invocation you should check, including `Read`, `ReadForUpdate`... If you don't check those invocations, you don't need to check `Commit` and `Rollback` here.

flycash: Actually I think that users should be responsible for keeping the txOrm is thread-safe. We don't need to do that.

PR #4082, opened 2020-07-20
Refactor RegisterDatabase

flycash:
In old API,  the `RegisterDatabase` API is:
```go
func RegisterDataBase(aliasName, driverName, dataSource string, params ...int) error {}
```
The last parameter `params` only accept `int`. But in general, we have many optional configurations, their types are different, and the `params...int` declaration is not convenient for users.

The new API is :
``` go
func RegisterDataBase(aliasName, driverName, dataSource string, params ...common.KV) error {}
```

### DIFF
--- a/orm/db_alias.go
+++ b/orm/db_alias.go
@@ -12,16 +12,21 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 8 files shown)


flycash:
@jianzhiyao please help to check

jianzhiyao:
checked

PR #4096, opened 2020-07-24
fix memory leak of request context

jianzhiyao:
close https://github.com/astaxie/beego/issues/4094

### DIFF
--- a/context/input.go
+++ b/context/input.go
@@ -323,28 +323,20 @@ func (input *BeegoInput) SetParam(key, val string) {
 // This function is used to clear parameters so they may be reset between filter
 // passes.
 func (input *BeegoInput) ResetParams() {
-	if len(input.pnames) > 0 {
-		input.pnames = input.pnames[:0]
-	}
-	if len(input.pvalues) > 0 {
-		input.pvalues = input.pvalues[:0]
-	}
+	input.pnames = input.pnames[:0]
+	input.pvalues = input.pvalues[:0]
 }
 
 // ResetData: reset data
 func (input *BeegoInput) ResetData() {
 	input.dataLock.Lock()
-	if input.data != nil {
-		input.data = nil
-	}
+	input.data = nil
 	input.dataLock.Unlock()
 }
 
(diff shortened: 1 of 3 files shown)


context/input.go
```diff
@@ -49,6 +49,7 @@ type BeegoInput struct {
 	CruSession    session.Store
 	pnames        []string
 	pvalues       []string
+	pLock         sync.RWMutex
```
flycash, line 52: I think we don't need the lock since this struct only be used in one goroutine.

router.go
```diff
@@ -319,6 +319,8 @@ func (p *ControllerRegister) GetContext() *beecontext.Context {
 
 // GiveBackContext put the ctx into pool so that it could be reuse
 func (p *ControllerRegister) GiveBackContext(ctx *beecontext.Context) {
+	// clear input cached data
+	ctx.Input.Clear()
```
flycash, line 323: Actually, not only Input but also Output need to be clear. 

If fact, we need to clear whole context.

context/input.go
```diff
@@ -323,8 +323,36 @@ func (input *BeegoInput) SetParam(key, val string) {
...
+	}
+	if len(input.pvalues) > 0 {
+		input.pvalues = input.pvalues[:0]
+	}
+}
+
+// ResetData: reset data
+func (input *BeegoInput) ResetData() {
+	input.dataLock.Lock()
+	if input.data != nil {
```
flycash, line 337: you don't need the `if-stmt`, `input.data=nil` is fine.

context/input.go
```diff
@@ -323,8 +323,36 @@ func (input *BeegoInput) SetParam(key, val string) {
...
+		input.data = nil
+	}
+	input.dataLock.Unlock()
+}
+
+// ResetBody: reset body
+func (input *BeegoInput) ResetBody() {
+	if len(input.RequestBody) > 0 {
```
flycash, line 345: you don't need this `if-stmt` either.

PR #4103, opened 2020-07-27
Enhance: NewOrmUsingDB & remove useless methods

flycash:
We need to add a new method that can create Orm instance by using specific alias because we remove `Using` method.

Last time I had said that `Using` is not thread-safe and it's dangerous. 

`QueryRelated` and `QueryRelatedWithCtx` are public methods but defined in private structure. It can't be used outside the package. So we remove them.


pkg/orm/orm.go
```diff
@@ -591,20 +577,22 @@ func (t *txOrm) Rollback() error {
 // NewOrm create new orm
 func NewOrm() Ormer {
 	BootStrap() // execute only once
+	return NewOrmUsingDB(`default`)
+}
 
+// NewOrm create new orm with the name
```
jianzhiyao, line 583: wrong comment method name

Issue #4093, opened 2020-07-23
Proposal: Generate routers based on annotations


flycash:
Generating routers like
```go
// CMS API
type CMSController struct {
    beego.Controller
}

func (c *CMSController) URLMapping() {
    c.Mapping("StaticBlock", c.StaticBlock)
    c.Mapping("AllBlock", c.AllBlock)
}

// @router /staticblock/:key [get]
func (this *CMSController) StaticBlock() {

}

// @router /all/:key [get]
func (this *CMSController) AllBlock() {
}
```
is very useful. 

But this feature has many problems. #3624 #3162 #4046 #3908 #3946 #3871 #3505 #3058 (most of them were reported by Chinese users, so those issues are written by Chinese).

This feature is affected by:
1. module name. The short name "module1" works but "xx.xxxx.xxx.module" don't;
2. Project is under GOPATH;
3. Using go mod

In fact, the root cause is that Beego looks up the controller path incorrectly.

Inspired by @AllenX2018 in #4046, we think we can add some configuration, for example, named "WebConfig.ControllerPath = path/to/controller", we will generate the routers in this path.

The default value is "./controller".

I think it's very simple and easy to implement.

AllenX2018:
I think I can try to create a pr to solve this issue : )

flycash:
@AllenX2018 Hi, now you own this. :)

flycash:
@AllenX2018 Please checkout a new branch from develop-2.0

AllenX2018:
@flycash OK

PR #4107, opened 2020-07-28
add comment router path configuration

AllenX2018:
Add an option to config the comment router directory to solve the probelm with comment router generation when using go module. See the discussion in #4093

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -32,6 +32,7 @@ require (
 	github.com/wendal/errors v0.0.0-20130201093226-f66c77a7882b // indirect
 	golang.org/x/crypto v0.0.0-20191011191535-87dc89f01550
 	golang.org/x/net v0.0.0-20190620200207-3b0461eec859 // indirect
+	golang.org/x/tools v0.0.0-20200117065230-39095c1d176c
 	google.golang.org/grpc v1.31.0 // indirect
 	gopkg.in/yaml.v2 v2.2.8
 )
--- a/pkg/beego.go
+++ b/pkg/beego.go
@@ -97,6 +97,7 @@ func initBeforeHTTPRun() {
 		registerTemplate,
 		registerAdmin,
 		registerGzip,
+		registerCommentRouter,
 	)
 
 	for _, hk := range hooks {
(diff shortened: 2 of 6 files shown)


AllenX2018:
> Does it works for sub directory? For example, if my controller directory looks like:

I'm sorry that it won't work for sub directory. I will add new commits in this pr to support it.

> And if we use go.mod, but it's out of GOPATH, does it work?

It works when we use go.mod. I've tested in my environment.

AllenX2018:
I add some commits:

- I introduce the `golang.org/x/tools/go/packages` package which can be used to scan all of the packages and build the AST of a directory. With this package, we don't need to call `parserPkg` every time when we call `Include` to add comment router, so I add a new function called `registerCommentRouter` to hook list and `parserPkg` will be called in it only once(develop mode)

- Since `parserPkg` will be called only once now, the `getpathTime` should scan the comment router directory recursively to find the last update time

- I've tested for both go.mod and GOPATH scenario

AllenX2018:
@flycash

flycash:
Thank you! I will test this PR next days. It will take some times, please wait for my feedback!

flycash:
Could you move these changes to the pkg since we copied all files to that directory? We want to adapt to the new project layout which all APIs are under pkg directory. We will remove old files in the future

AllenX2018:
OK, I will move them to the files under pkg directory.

AllenX2018:
> Could you move these changes to the pkg since we copied all files to that directory?

Done

AllenX2018:
I clone your project to my local environment and test. It failed because `pkg.Syntax` is nil too. I debug and find that the reason is that the version of the `golang.org/x/tools` package which the go.mod depend on is too old. It need to be updated. I add the below line to the go.mod file of beego and the problem will be solved:

```
golang.org/x/tools v0.0.0-20200806234136-990129eca547
```

Pls try again and tell if you have any other question!

AllenX2018:
I've add a commit to update the verion of `golang.org/x/tools` for go.mod to depend on.

flycash:
@AllenX2018 Thank you! It works well. And could you please merge develop-2.0? We had fixed some unit tests, CI is more stable.

Issue #4088, opened 2020-07-22
Proposal: Field serializer and json implementation


flycash:
Many users store the json in one column. Though I think it's not a good practice, but it's very convinient. In such cases, users want a feature that we can sirialize the value to json and then store into DB. When desialize record from DB, we can convert the json value to object.

Here is an example:

```go
type User struct {
    Name string
    Attributes []Attribute `serializer: json`
}
```
The `[]Attribute` will be converted to json string before storing it into DB. 

The `serializer` represents one implementation of interface:
```go
type FieldSerializer interface {
    Serialize(obj interface{}) string
    Deserialize(str interface) interface{}
}
```

We can allow user to register their own implementation:
```go
orm.RegisterFieldSerializer("my-s", sri)
```

They can use it in their structure:

```go
type User struct {
   SomeField XXX `serializer: "my-s"`
}
```

tayoogunbiyi:
I'd work on this. You can assign it to me.

flycash:
Please checkout branch from develop-2.0 and put your code in pkg/orm

tayoogunbiyi:
I've been working on this but there are some blockers involving how to restore the JSON stored in the DB back to the original model's field type.
I'm going to link my PR so far here (in case someone wants to continue here), and come back to it some other time.)

https://github.com/astaxie/beego/pull/4118

flycash:
> I've been working on this but there are some blockers involving how to restore the JSON stored in the DB back to the original model's field type.

My idea is that you can add more parameter in `deserialize` method, like `field reflect.Field` or `fi fieldInfo`

jianzhiyao:
Hey guys!
@flycash @tayoogunbiyi 
### @flycash 's proposal
I think that proposal is not a good idea!
That could limit data format in many using case.
```
type User struct {
    Name string
    Attributes []Attribute `serializer: json`
}
```

### My proposal
Implementing ``encoding.BinaryMarshaler`` and ``encoding.BinaryUnmarshaler`` for model struct.
And return error when struct not implements then.
```
type User struct {
	Name           string
	JsonAttributes []JsonAttribute
	XmlAttributes  []XmlAttribute
}

type JsonAttribute struct {
	Key   string
	Value string
}

var _ encoding.BinaryMarshaler = new(JsonAttribute)
var _ encoding.BinaryUnmarshaler = new(JsonAttribute)

func (a *JsonAttribute) UnmarshalBinary(data []byte) error {
[...]

PR #4118, opened 2020-07-30
Serialization/Deserialization

tayoogunbiyi:
Aims to close https://github.com/astaxie/beego/issues/4088

### DIFF
--- a/pkg/orm/models_info_f.go
+++ b/pkg/orm/models_info_f.go
@@ -241,7 +241,24 @@ checkType:
 			}
 		}
 
-		fieldType, err = getFieldType(addrField)
+		serializerKey := sf.Tag.Get(serializerStructTagName)
+
+		if serializerKey != "" {
+			fieldType = TypeVarCharField
+			err = nil
+
+			if fi.auto || fi.pk {
+				err = fmt.Errorf("rel/reverse:many field must be slice")
+				goto end
+			} else if _, ok := serializerMap[serializerKey]; !ok {
+				err = fmt.Errorf("serializer %s is not registered", serializerKey)
+				goto end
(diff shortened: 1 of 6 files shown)


pkg/orm/models_info_f.go
```diff
@@ -241,7 +242,27 @@ checkType:
...
 
-		fieldType, err = getFieldType(addrField)
+		serializerKey := sf.Tag.Get(serializerStructTagName)
+
+		if serializerKey != "" {
+			tags["type"] = "text"
+			fieldType = TypeTextField
+			err = nil
+
+			if fi.auto || fi.pk {
```
flycash, line 252: I think we don't need to check those. Users should be responsible for making sure that the serializer works for this field.

pkg/orm/serializers.go
```diff
@@ -0,0 +1,40 @@
...
+	"encoding/json"
+)
+
+// SerializerMap maps from the serializer's string representation to the Serializer
+var serializerMap = make(map[string]FieldSerializer)
+
+// FieldSerializer is the interface which all serializers must implement
+type FieldSerializer interface {
+	Serialize(obj interface{}) string
+	Deserialize(fieldContent string) interface{}
```
flycash, line 13: If you want to deserialize, you may need the field's metadata.

tayoogunbiyi:
Next step is to handle nil fields appropriately, as well as add more tests for the fields in the comment mentioned above.

flycash:
are you still working on this? It's hard to verify without unit tests.

tayoogunbiyi:
I've not been able to work on this for a while. Please feel free to re-assign

PR #4119, opened 2020-07-30
Add custom logging format functionality

IamCathal:
A decent implementation of user defined log formats https://github.com/astaxie/beego/pull/4063. Users can use the existing Apache or JSON formatters or they can define their own for a single log adapter or a global setting for all adapters. To see the results use this script:

```go
package main

import (
	"encoding/json"
	"fmt"
	"time"

	"github.com/astaxie/beego"
)

// AccessLogRecord struct for holding access log data.
type AccessLogRecord struct {
	RemoteAddr     string        `json:"remote_addr"`
	RequestTime    time.Time     `json:"request_time"`
	RequestMethod  string        `json:"request_method"`
	Request        string        `json:"request"`
	ServerProtocol string        `json:"server_protocol"`
	Host           string        `json:"host"`
	Status         int           `json:"status"`
	BodyBytesSent  int64         `json:"body_bytes_sent"`
	ElapsedTime    time.Duration `json:"elapsed_time"`
	HTTPReferrer   string        `json:"http_referrer"`
	HTTPUserAgent  string        `json:"http_user_agent"`
	RemoteUser     string        `json:"remote_user"`
	FilePath       string        `json:"file_path"`
	LoggerLevel    string        `json:"logger_level"`
}

func Formatter(req string) string {
	var r AccessLogRecord
	_ = json.Unmarshal([]byte(req), &r)
	// Unmarshal JSON into struct, then return whatever format the user wants from the fields given
	return fmt.Sprintf("%s %d %s [%db]", r.Request, r.Status, r.LoggerLevel, r.BodyBytesSent)
}

func main() {

	ctrl := &MainController{}
	beego.Router("/yuppa", ctrl, "get:Yuppa")

	beego.BeeLogger.FullFilePath = true
	beego.BeeLogger.EnableFuncCallDepth(true)

	// Setting global formatter overrides any other formatters set for
	// individual logs e.g file, console etc
	// beego.BeeLogger.SetGlobalFormatter(Formatter)

	// Set our custom Formatter func to format our logs for our file output
	beego.SetLogger("file", Formatter, `{"filename":"test.json"}`)
	// Set the default JSON formatter for our console logs
	beego.SetLogger("console", beego.BeeLogger.JSONFormatter, "")

	beego.Run()
}

type MainController struct {
	beego.Controller
}

func (ctrl *MainController) Yuppa() {
	output := ctrl.Ctx.Output
	output.Header("Content-Type", "application/json")
	ctrl.Ctx.WriteString("Golang is a great language\n")
}

```

I came into one main problem with this implementation. To define your own logging function you see the function takes in a string and returns a string. 
```go
func Formatter(inputString string) string {}
```
This `inputString` is actually a JSON formatted string representation of the `AccessLogRecord` struct. I think using this struct as an argument would be much better and clear compared to having to import the struct and then unmarshal it to manipulate the logging format. The fields in the `AccessLogRecord` struct are all that are needed as it holds all information about the request that might need to be used by the user. When I tried to change the argument from `string` to `AccessLogRecord` I couldn't do this as it gave me some errors about mismatched types which I couldn't understand in the end. I could only make the `Formatter` func accept types such as string and int, structs didn't work.

Overall I'm happy with this. Once all the issues are sorted I can update and write some more documentation on how to use this new feature since the [current](https://beego.me/docs/module/logs.md) docs will be outdated. There might be some bugs and edge cases but this is a good start and I'm happy to implement any fix  that you can come up with. If you have any questions I'd be free to hop on a call with you to explain my thoughts that went into the code.

*Conflicts are just from deletions of those files that can be solved easily. I'm not quite sure how accessing log functions changed because of it*


pkg/log.go
```diff
@@ -51,8 +51,8 @@ func SetLogFuncCall(b bool) {
 
 // SetLogger sets a new logger.
 // Deprecated: use github.com/astaxie/beego/logs instead.
-func SetLogger(adaptername string, config string) error {
-	return logs.SetLogger(adaptername, config)
+func SetLogger(adaptername string, fmtFunc logs.FormatterFunc, config string) error {
```
flycash, line 54: I think we should add a new method named 
```go
SetLoggerWithOpts(adaptername string,  config string, opts... common.KV)
```
We could use KV defined in pkg/common or defining similar structure in log module.

In most cases, users don't want to custom anything, so they can use old method `SetLoggger`

pkg/logs/log.go
```diff
@@ -91,6 +94,36 @@ type Logger interface {
 	Flush()
 }
 
+type FormatterFunc func(string) string
+
+type formatter interface {
```
flycash, line 99: `FormatterFunc` and `formatter` we only need one of them. This is the most important part. The key method should looks like:
```go
func format(when, level, filepath, linenumber, msg, args...interface{})
```
Thining about this case:
1. Users want to output logs look like "INFO|2020.7.8|/path/to/my/business.go:176|Hello, here is my log"
2. Some users want to output logs look like "2020.7.8 business.go:176 [INFO] Hello, here is my log"

Taking scalability into consideration, we may define a structure:
```go
type LogRecord structure {
    when time.Time,
    level int,
    filepath string, 
    linenumber int, 
    msg string, 
    args...interface{}
}
```
The key method is changed to:
```go
func format(record LogRecord) string {}
```

So the `WriteMsgV2` is 
```go
func WriteMsgV2(record LogRecord)...
```

pkg/log.go
IamCathal: I agree that most users would not want to set a formatter and `SetLoggerWithOpts` would be a better idea.

pkg/logs/log.go
IamCathal: Sure, different arguments for the format function would be great. I just couldn't get it to work on my own for some reason. I'm free tomorrow or on Sunday anytime from 11:00 (GMT+1) to talk about this new feature on a call and see what needs to be reworked to get a working solution

flycash: Sorry, I need some time to prepare a demo. And because air tickets and hotels are extreme cheap due to the virus, I go vocation every weekend. I think next Monday would be better. 😄

IamCathal: Sure no worries, so Monday the 7th?

flycash: The 3rd of August. And I am free from 10:00 to 23:00(UTC+8) every weekday, so you can choose any time you like.

PR #4120, opened 2020-07-31
Update redis session provider

phiphi282:
This pull request changes the redis session provider to use go-redis instead of redigo.

This change allows to add better connection timeout checking and retrying for more stable connections.

Additionally the interface for providers was changed slightly to return an error in the sessionExist function.

Our specific usecase for this is using harbor with a high available redis behind it. Before it often failed because of a timed out session. This was specifically hard to debug because the error for the sessionExist function got swallowed.

Few questions:
- Is it okay to change the provider interface? This change breaks existing custom providers.
- With go-redis v8 a mandatory context is introduced. Should I add this to the provider interface as well?
- There are new variables added to the redis Client. Should these be configurable and used for sentinel and cluster as well?

The change of the redis session provider itself does not come with any breaking changes.


flycash:
> This pull request changes the redis session provider to use go-redis instead of redigo.
> 

I am not sure whether we need to change redis client. It's a big change. I noticed that users prefer to use redigo in their application because we choose the redigo. So they will be impacted.

We could think about creating a PR for redigo to support timeout feature.

> Few questions:
> 
> * Is it okay to change the provider interface? This change breaks existing custom providers.

We are doing refactoring, and session module will be refactored in the future. At that time, we will redesign those API. And add more parameters. I think swallowing error is not acceptable and I wish to change it.

> * With go-redis v8 a mandatory context is introduced. Should I add this to the provider interface as well?

We probably introduce context in the API because we want to introduce metric and tracing feature to session module.

> * There are new variables added to the redis Client. Should these be configurable and used for sentinel and cluster as well?

This could be optional and I personally want to support sentinel and cluster.


If you want to do those enhancements, it will be very appreciated.  But we don't accept any new feature PR merging to branch `develop`. Only critical bug fixed is acceptable. 

You could checkout branch from `develop-2.0`. And you can change the API.

phiphi282:
I split my pr in 2 parts and changed the api for develop-2.0 now.
#4137 and #4124

PR #4121, opened 2020-07-31
Refactor DeleteBatch

### LINKED ISSUE #4010: got 1516473 parameters but PostgreSQL only supports 65535 parameters when invokes Delete()

```
(ENV) 🍺 /Users/xsky/go/src/xsky-demon ☞ git:(test) ✗ bee version
______
| ___ \
| |_/ /  ___   ___
| ___ \ / _ \ / _ \
| |_/ /|  __/|  __/
\____/  \___| \___| v1.10.0

├── Beego     : 1.10.1
├── GoVersion : go1.14.3
├── GOOS      : darwin
├── GOARCH    : amd64
├── NumCPU    : 4
├── GOPATH    : /Users/xsky/go
├── GOROOT    : /usr/local/go
├── Compiler  : gc
└── Date      : Monday, 8 Jun 2020
```

```
(ENV) 🍺 /Users/xsky/go/src/xsky-demon ☞ git:(test) ✗ go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/xsky/Library/Caches/go-build"
GOENV="/Users/xsky/Library/Application Support/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/xsky/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/Cellar/go/1.14.3"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/Cellar/go/1.14.3/pkg/tool/darwin_amd64"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/pb/rc1txhc12_x0jt1tmld1bn300000gn/T/go-build782939608=/tmp/go-build -gno-record-gcc-switches -fno-common"
```

3. What did you do?
The program invokes `Delete()` in the file of `github.com/astaxie/beego/orm/types.go`, then calls  `Delete()` in file of `github.com/astaxie/beego/orm/orm_queryset.go`, the actual invocation is `DeleteBatch()`in file of `github.com/astaxie/beego/orm/db.go`
```
178 // execute delete
179 func (o *querySet) Delete() (int64, error) {
180     return o.orm.alias.DbBaser.DeleteBatch(o.orm.db, o, o.mi, o.cond, o.orm.alias.TZ)
181 }
```
`DeleteBatch` first queries all the records, then justify the count of records and delete specific records using `IN (...)` statement, but it has limitation of no larger than 65535.
https://github.com/astaxie/beego/blob/3e30f37172671150e8903828e955c3ee5de1793e/orm/db.go#L885
https://github.com/lib/pq/blob/d408f9c9485706b60bb2b9b9718f3c54f077b506/conn.go#L1342
 
Succeed to delete records.

5. What did you see instead?
time=2020-06-08T11:18:03+08:00 level=error module=controllers/base.go:477 msg="pq: got 1516473 parameters but PostgreSQL only supports 65535 parameters

gmelodie:
`DeleteBatch` previously selected the desired entries on the database and, one by one, deleted them. Now it should only call SQL's `DELETE` statement directly.

Fixes #4010.

**Obs**: It seems the testing database assumes the `DeleteBatch` function will delete related entries on other tables, regardless of what types of relations (e.g. foreign keys) they actually have on the SQL/database level. The way I see it this is not desired. We only want the ORM to wrap around the database, not do its job. 

That being said, I believe these failing tests should be fixed not by changing how `DeleteBatch` works, but changing the way the testing models are actually implemented on SQL.

### DIFF
--- a/pkg/orm/db.go
+++ b/pkg/orm/db.go
@@ -811,7 +811,7 @@ func (d *dbBase) UpdateBatch(q dbQuerier, qs *querySet, mi *modelInfo, cond *Con
 }
 
 // delete related records.
-// do UpdateBanch or DeleteBanch by condition of tables' relationship.
+// do UpdateBatch or DeleteBatch by condition of tables' relationship.
 func (d *dbBase) deleteRels(q dbQuerier, mi *modelInfo, args []interface{}, tz *time.Location) error {
 	for _, fi := range mi.fields.fieldsReverse {
 		fi = fi.reverseFieldInfo
@@ -860,11 +860,9 @@ func (d *dbBase) DeleteBatch(q dbQuerier, qs *querySet, mi *modelInfo, cond *Con
 
 	// SELECT t0.column FROM 'table' T0 JOIN ... WHERE ...
 	selectQuery := fmt.Sprintf("SELECT T0.%s FROM %s%s%s T0 %s%s", cols, Q, mi.table, Q, join, where)
-	fmt.Println("selectQuery --------------> ", selectQuery)
 
 	// DELETE FROM 'table' WHERE column IN (SELECT t0.column FROM 'table' T0 JOIN ... WHERE ...)
 	deleteQuery := fmt.Sprintf("DELETE FROM %s%s%s WHERE %s IN (%s)", Q, mi.table, Q, cols, selectQuery)
-	fmt.Println("deleteQuery --------------> ", deleteQuery)
 	d.ins.ReplaceMarks(&deleteQuery)
 
 	// Execute delete query
@@ -880,37 +878,8 @@ func (d *dbBase) DeleteBatch(q dbQuerier, qs *querySet, mi *modelInfo, cond *Con
 		return 0, err
 	}
 
-	nDeleted, err := result.RowsAffected()
-	if err != nil {
-		return 0, err
-	}
-	if nDeleted > 0 {
-		err := d.deleteRels(q, mi, args, tz)
-		if err != nil {
-			return nDeleted, err
-		}
-	}
-	return nDeleted, nil
-
-	// result, err := q.Exec(deleteQuery, args...)
-	// if err != nil {
-	// 	return 0, err
-	// }
-
 	// FIXME: RowsAffected is not supported by all DB drivers
-
-	// Count deleted entries
-	// TODO: enhance count using sql COUNT
-
-	// setValues = append(setValues, pkValue)
-
-	// sep := fmt.Sprintf("%s = ?, %s", Q, Q)
-	// setColumns := strings.Join(setNames, sep)
-
-	// query := fmt.Sprintf("UPDATE %s%s%s SET %s%s%s = ? WHERE %s%s%s = ?", Q, mi.table, Q, Q, setColumns, Q, Q, pkName, Q)
-
-	// d.ins.ReplaceMarks(&query)
-
+	return result.RowsAffected()
 }
 
 // read related records.
--- a/pkg/orm/db.go
+++ b/pkg/orm/db.go
@@ -848,74 +848,69 @@ func (d *dbBase) DeleteBatch(q dbQuerier, qs *querySet, mi *modelInfo, cond *Con
 	}
 
 	if cond == nil || cond.IsEmpty() {
-		panic(fmt.Errorf("delete operation cannot execute without condition"))
+		return 0, errors.New("Delete operation cannot execute without condition")
 	}
 
+	// Build delete query
 	Q := d.ins.TableQuote()
 
 	where, args := tables.getCondSQL(cond, false, tz)
 	join := tables.getJoinSQL()
+	cols := fmt.Sprintf("%s%s%s", Q, mi.fields.pk.column, Q)
 
-	cols := fmt.Sprintf("T0.%s%s%s", Q, mi.fields.pk.column, Q)
-	query := fmt.Sprintf("SELECT %s FROM %s%s%s T0 %s%s", cols, Q, mi.table, Q, join, where)
+	// SELECT t0.column FROM 'table' T0 JOIN ... WHERE ...
+	selectQuery := fmt.Sprintf("SELECT T0.%s FROM %s%s%s T0 %s%s", cols, Q, mi.table, Q, join, where)
+	fmt.Println("selectQuery --------------> ", selectQuery)
 
-	d.ins.ReplaceMarks(&query)
+	// DELETE FROM 'table' WHERE column IN (SELECT t0.column FROM 'table' T0 JOIN ... WHERE ...)
+	deleteQuery := fmt.Sprintf("DELETE FROM %s%s%s WHERE %s IN (%s)", Q, mi.table, Q, cols, selectQuery)
+	fmt.Println("deleteQuery --------------> ", deleteQuery)
+	d.ins.ReplaceMarks(&deleteQuery)
 
-	var rs *sql.Rows
-	r, err := q.Query(query, args...)
+	// Execute delete query
+	var err error
+	var result sql.Result
+
+	if qs != nil && qs.forContext {
+		result, err = q.ExecContext(qs.ctx, deleteQuery, args...)
+	} else {
+		result, err = q.Exec(deleteQuery, args...)
+	}
 	if err != nil {
 		return 0, err
 	}
-	rs = r
-	defer rs.Close()
 
-	var ref interface{}
-	args = make([]interface{}, 0)
-	cnt := 0
-	for rs.Next() {
-		if err := rs.Scan(&ref); err != nil {
-			return 0, err
-		}
-		pkValue, err := d.convertValueFromDB(mi.fields.pk, reflect.ValueOf(ref).Interface(), tz)
+	nDeleted, err := result.RowsAffected()
+	if err != nil {
+		return 0, err
+	}
+	if nDeleted > 0 {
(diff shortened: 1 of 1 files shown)


flycash:
From the codes, I think your refactoring work should be compatible with old unit tests. I will check those failed tests when I am free. 😄

flycash:
> **Obs**: It seems the testing database assumes the `DeleteBatch` function will delete related entries on other tables, regardless of what types of relations (e.g. foreign keys) they actually have on the SQL/database level. 

I am not the ORM module author but I guess that there are some foreign key indexes in the table. According to the SQL standard, DB deletes related entries automatically.

gmelodie:
I've used [SQLite Browser](https://sqlitebrowser.org/) to check out the testing DB and I didn't see any foreign keys being created. But that was my first time trying it, I'll probably give this another try on Monday ;)

flycash:
> I've used [SQLite Browser](https://sqlitebrowser.org/) to check out the testing DB and I didn't see any foreign keys being created. But that was my first time trying it, I'll probably give this another try on Monday ;)

I read the original codes, I didn't find any codes used to remove related entries. In the beginning, I believed that we do query before delete entries because we need to get all removed ids and then use them to removed related entries. But I can not find that logic in original codes. 

So I guess that we use some foreign key indexes. 

I just notice that why we don't use "delete from xxx where xxx and xxx " but use "delete from xxx where id in (select xxx where xxx)"?

flycash:
I just test this PR and found that the SQL can't work. I run ORM test and the orm_test.go#TestCURD failed. And then I rewrite a simple test:
```go
	profile := NewProfile()
	profile.Age = 30
	profile.Money = 1234.12
	_, err := dORM.Insert(profile)

	assert.Nil(t, err)

	user := NewUser()
	user.UserName = "sleneXXX"
	user.Email = "vslene@gmail.com"
	user.Password = "pass"
	user.Status = 3
	user.IsStaff = true
	user.IsActive = true
	user.Profile = profile

	_, err = dORM.Insert(user)
	assert.Nil(t, err)


	num, err := dORM.Delete(user)
	assert.Nil(t, err)
	assert.Equal(t, int64(1), num)
```

I got 
```shell
[DELETE FROM `post` WHERE `id` IN (SELECT T0.`id` FROM `post` T0 WHERE T0.`user_id` IN (?) )] - `239` - Error 1093: You can't specify target table 'post' for update in FROM clause
```

gmelodie:
So [according to this](https://stackoverflow.com/a/45498/6032037) it seems that the problem is in the way the query is constructed. If we did something like `DELETE FROM post WHERE id IN (?)` it should work, but I'm not sure we would be able to use `JOIN` clauses to select data then.

flycash:
You can set the `Debug` flag to true. I think using `join` looks good. 
image

in file `models_test.go`

gmelodie:
> I just test this PR and found that the SQL can't work. I run ORM test and the orm_test.go#TestCURD failed.
> I got
> Error 1093: You can't specify target table 'post' for update in FROM clause

According to [this](https://stackoverflow.com/a/45498/6032037) it seems to be specific for the MySQL database (I'm running my tests on SQLite3 and am not getting this error on the same query). I don't know how to proceed about this. Do we try a general solution or a workaround for MySQL?

Also, there is another error I'm getting: Deleting a profile on the `user_profile` table does not override the related `profile_id` on the `user` table to `NULL`. In other words: if a user has a profile and this profile gets deleted, then the `profile_id` of the user (on the `user` table should be changed to `NULL`, which doesn't happen for some reason).

```
[Queries/default] - [  OK /     db.Exec /     5.0ms] - [DELETE FROM `user_profile` WHERE `id` IN (SELECT T0.`id` FROM `user_profile` T0 LEFT OUTER JOIN `user` T1 ON T1.`profile_id` = T0.`id` WHERE T1.`user_name` = ? )] - `slene`
[Queries/default] - [  OK / db.QueryRow /     0.1ms] - [SELECT COUNT(*) FROM `user` T0 LEFT OUTER JOIN `user_profile` T1 ON T1.`id` = T0.`profile_id` WHERE T0.`user_name` = ? AND T0.`profile_id` IS NULL ] - `slene`
    TestDelete: orm_test.go:157: 	
        Error: expected: `1`, get `0`
        orm_test.go:TestDelete:1997: 
             1988:   func TestDelete(t *testing.T) {
             1989:       qs := dORM.QueryTable("user_profile")
             1990:       num, err := qs.Filter("user__user_name", "slene").Delete()
             1991:       throwFail(t, err)
             1992:       throwFail(t, AssertIs(num, 1))
             1993:   
             1994:       qs = dORM.QueryTable("user")
             1995:       num, err = qs.Filter("user_name", "slene").Filter("profile__isnull", true).Count()
             1996:       throwFail(t, err)
         >>  1997:       throwFail(t, AssertIs(num, 1))

```

Should we take care of this SQL errors and then tackle the MySQL-specific error or should I focus on the MySQL one first?

flycash:
we need to test this over all supported DB. I think we should rewrite this DELETE statement. Using inner join would be better. 

And the second question, do you mean that original codes forget to set the `profile_id` to NULL?

gmelodie:
> I think we should rewrite this DELETE statement. Using inner join would be better.

I'll look into this

> And the second question, do you mean that original codes forget to set the profile_id to NULL?

I *think* so. I think the original ORM goes around nullifying related entries manually instead of letting the database handle it. If I'm right, we would need to make some changes in the tests. 

The biggest problem I'm facing is understanding the code, the variables are not very explanatory (`i`, `mi`, etc.) but I'll give this another shot  :)

flycash:
> I _think_ so. I think the original ORM goes around nullifying related entries manually instead of letting the database handle it. If I'm right, we would need to make some changes in the tests.
> 

If original codes don't set the ref_id to NULL, we could keep this behavior. But I think the better way is setting it as null. You can make this decision by yourself.


> The biggest problem I'm facing is understanding the code, the variables are not very explanatory (`i`, `mi`, etc.) but I'll give this another shot :)

Yes, these are so-called "c-styled" name. 
I put this enhancement to v2.0.1 since we will release v2.0.0-beta this weekend. After that, we don't merge any new enhancement or feature. 

So you have a lot of time to do this. Don't worry. 😄

gmelodie:
Yeah so I'm thinking there's quite some room for improvement here.

The problem is that the ORM is not using native Foreign Keys (meaning SQL Foreign Keys), it creates an abstraction of FKs instead of relying on the database. This is bad because, as we've seen, it creates the need for a lot of wrapping and manual `SELECT`ing and `DELETE` ing of related entries, which ideally the database should be able to do on its own. 

I'm having a hard time testing it. I wanted to create my own small testing code using the changes I do in the `/pkg/orm/db.go` file. Do you have some tips on how to do that?

Thanks again for all the patience and understanding, really appreciate it :smiley:

flycash:
> The problem is that the ORM is not using native Foreign Keys (meaning SQL Foreign Keys), it creates an abstraction of FKs instead of relying on the database. This is bad because, as we've seen, it creates the need for a lot of wrapping and manual `SELECT`ing and `DELETE` ing of related entries, which ideally the database should be able to do on its own.

In practice, we don't use native foreign keys because we may need partition, sharding in the future. Native FK is a big problem. Another concern is that, usually DB is the performance bottleneck, so we try best to move calculation from DB to application codes. And taking DB lock into consideration, using native DB FK makes lock condition worst.


> I'm having a hard time testing this though. I wanted to create my own small testing code using the changes I do in the `/pkg/orm/db.go` file. Do you have some tips on how to do that?

Those tests in DB are dependent on each other. If you want to run tests locally or from IDE, you can check the scripts/test.sh. It takes three steps:
```sh
docker-compose -f "$(pwd)/scripts/test_docker_compose.yaml" up -d
```
we use docker to start the DB and other dependencies.

```sh
export ORM_DRIVER=mysql
export TZ=UTC
export ORM_SOURCE="beego:test@tcp(localhost:13306)/orm_test?charset=utf8"
```
you can change `ORM_DRIVER`, `ORM_SOURCE`. 

And then you can run tests.

Another simple way is to change the test codes which you need to roll back when you finish your tests.
```go
func init() {
	// Debug, _ = StrTo(DBARGS.Debug).Bool()
	//
	// if DBARGS.Driver == "" || DBARGS.Source == "" {
	// 	fmt.Println(helpinfo)
	// 	os.Exit(2)
	// }

	Debug = true
	DBARGS.Driver = "mysql"
	DBARGS.Source="you connection info"

	RegisterDataBase("default", DBARGS.Driver, DBARGS.Source, 20)

	alias := getDbAlias("default")
	if alias.Driver == DRMySQL {
		alias.Engine = "INNODB"
	}

}
```
This method is in file models_test.

PR #4124, opened 2020-08-03
Update session provider interface to return errors on SessionExist

phiphi282:
Update the interface and all default providers to return an error for the sessionExist function.

This change makes debugging issues a lot easier because possible errors don't get swallowed and shown to the user instead.

Additionally this pr changes the behavior of the sessionStart function because it now fails when sessionExists returns an error instead of assuming no session for the given sid exists.


session/README.md
```diff
@@ -101,7 +101,7 @@ Maybe you will find the **memory** provider is a good example.
 	type Provider interface {
 		SessionInit(gclifetime int64, config string) error
 		SessionRead(sid string) (SessionStore, error)
-		SessionExist(sid string) bool
+		SessionExist(sid string) (bool, error)
```
flycash, line 104: why we need to return `error`. I think returning false when got exception is reasonable too.

So, in what scenario, we want to check whether return error?

phiphi282: For example if connecting to a redis client fails, the session start method would just assume that there is no session available and is creating a new session.

We had this problem while trying to create a harbor instance with high available redis and since the error message got swallowed it was fairly hard to debug.

phiphi282: Created this pr because of #4120 
There should be some additional information as well

flycash: We had copied session module to pkg/session and old session directory will be removed in v2.1.x.

Please put those codes into pkg/session. :)

flycash: We don't want to change the old session's API

phiphi282: Thanks for telling me :) 
I have moved my changes to pkg/session
The session tests have run successfully for me as well

PR #4128, opened 2020-08-04
fix CI fail for connection log test

AllenX2018:
I notice that the CI fail sometimes because of log module's `TestReconnect` test case. This pr fix it.

### DIFF
--- a/logs/conn_test.go
+++ b/logs/conn_test.go
@@ -18,6 +18,7 @@ import (
 	"net"
 	"os"
 	"testing"
+	"time"
 )
 
 // ConnTCPListener takes a TCP listener and accepts n TCP connections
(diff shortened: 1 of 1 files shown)


flycash:
Thank you!

flycash:
It makes some other problems. Do you test it?

AllenX2018:
Yes, I've tested it. And I check the fail log on travis ci and found theses output information:
> ......
> FAIL	github.com/astaxie/beego	0.055s
> ......
> FAIL	github.com/astaxie/beego/toolbox	600.006s
> ......

In this pr I just modify the `TestReconnect` function. The test case in **github.com/astaxie/beego** and **github.com/astaxie/beego/toolbox** should not be affected. It's strange...

flycash:
Yes, I can't understand this behavior. Now I moved this test to pkg/conn_test, it works well.

PR #4136, opened 2020-08-06
Feature: implement the time precison for time.Time type

### LINKED ISSUE #3875: Model中datetime类型的字段可以指定精度吗？
在model中对time.time类型进行标注的时候是：
type m struct{
...
Timestamp `orm:"type(datetime)"`
}
实际建表时，timestamp字段是datetime(0)，如何调整时间的精度，比如设置成datetime(6)；试了一下把注解改为 `orm:"type(datetime(6))"`没有效果

AllenX2018:
Fixes #3875


AllenX2018:
I'm adding the unit test. But I'm confused why the precision of the time is cut when the time was read from the database?

https://github.com/astaxie/beego/blob/2fce8f9d1bba8c93128f876f4dde712ca97be9ae/pkg/orm/db.go#L1329-L1332

flycash:
@jianzhiyao could you help to check this PR? I think you are very familiar with ORM module.

pkg/orm/db_mysql.go
```diff
@@ -42,24 +42,25 @@ var mysqlOperators = map[string]string{
...
-	"uint64":          "bigint unsigned",
-	"float64":         "double precision",
-	"float64-decimal": "numeric(%d, %d)",
+	"auto":                "AUTO_INCREMENT NOT NULL PRIMARY KEY",
+	"pk":                  "NOT NULL PRIMARY KEY",
+	"bool":                "bool",
+	"string":              "varchar(%d)",
+	"string-char":         "char(%d)",
+	"string-text":         "longtext",
+	"time.Time-date":      "date",
```
flycash, line 51: Is it possible that users need "time.Time-date-precision"?

AllenX2018: The date saves the precision to the day, the format is: YYYY-MM-DD

AllenX2018:
> Please provide unit tests.
> You must test this feature overall DBs.

The unit test has not been submitted, because the query result of the test framework will cut the time precision, so it cannot pass the test. This was mentioned in the discussion above.

flycash:
> I'm adding the unit test. But I'm confused why the precision of the time is cut when the time was read from the database?
> 
> https://github.com/astaxie/beego/blob/2fce8f9d1bba8c93128f876f4dde712ca97be9ae/pkg/orm/db.go#L1329-L1332

Let me check this later

flycash:
> I'm adding the unit test. But I'm confused why the precision of the time is cut when the time was read from the database?
> 
> https://github.com/astaxie/beego/blob/2fce8f9d1bba8c93128f876f4dde712ca97be9ae/pkg/orm/db.go#L1329-L1332

I check this file, and I think we need to change those codes to support the more precise date time. But it will be a hard job because we need to test it overall DB. 

I put this to V2.0.1 and I plan to release v2.0.0-beta this weekend.

PR #4137, opened 2020-08-06
Add additional options to redis session prov

phiphi282:
Adding option for frequency of checking timed out connections as well as
an option to specify retries.

These changes make redis provider more stable since connection problems
are becoming fewer.

Since redigo does not have this options and since redis_sentinel and
redis_cluster are using go-redis as a client, this commit changes from
redigo to go-redis for redis session provider.

Added tests for redis session provider as well.

There were some concerns for changing the redis client.
I can change the client back to redigo when the retry/idle-connection are implemented there as well, unfortunately I don't have the capacity to do it myself currently.


flycash:
I am not sure whether we should change the redis dependency from redigo to go-redis. So, @AllenX2018, @IamCathal, @tayoogunbiyi , @jianzhiyao , what do you think?

IamCathal:
I haven't used either libraries before or redis itself, but from a general standpoint with a major release (like the upcoming 2.0 release) I think making some bigger changes wouldn't be the worst idea as a big release is meant to bring in many new features and change a lot of things that can't be changed in a minor release like 1.8 -> 1.9.

Most of the PR is adding tests and changing dependencies with a smaller part being actual changes to the codebase. So atleast from just reading this PR the difference from a user's point of view doesn't seem to bad (provided the tests still pass with go-redis). I don't know enough about the two libraries to make a definitive decision as one library might support x feature but the other doesn't which might end up with users in the future wanting to switch back for the same reason you want to switch now.

So overall if someone has a deeper knowledge of these two libraries and the existing implementation of redis in beego I'd say they should make the decision. Effecting userspace would be the biggest concern if we were to switch but again I'm not sure if that would be a big issue.

phiphi282:
@IamCathal you have two other similar session providers called redis-sentinel and redis-cluster that are already using go-redis.
Only the default redis session provider is still using redigo.
Though outside of the session package I think redigo is being used for caching as well.

phiphi282:
I have resolved the conflicts now.

No problem with taking your time. It's always better to be safe than sorry. 😃

pkg/session/redis/sess_redis.go
```diff
@@ -157,45 +156,40 @@ func (rp *Provider) SessionInit(maxlifetime int64, savePath string) error {
...
+		}
 	}
 
-	rp.poollist.IdleTimeout = idleTimeout
+	rp.poollist = redis.NewClient(&redis.Options{
+		Addr:               rp.savePath,
+		Password:           rp.password,
+		PoolSize:           rp.poolsize,
+		DB:                 rp.dbNum,
+		IdleTimeout:        idleTimeout,
```
flycash, line 179: Why not declare those config items as `rp`'s field? like `rp.idleTimeout`...

pkg/session/redis/sess_redis.go
```diff
@@ -212,38 +206,34 @@ func (rp *Provider) SessionRead(sid string) (session.Store, error) {
...
 		// oldsid doesn't exists, set the new sid directly
 		// ignore error here, since if it return error
 		// the existed value will be 0
-		c.Do("SET", sid, "", "EX", rp.maxlifetime)
+		c.Do(c.Context(), "SET", sid, "", "EX", rp.maxlifetime)
 	} else {
-		c.Do("RENAME", oldsid, sid)
-		c.Do("EXPIRE", sid, rp.maxlifetime)
+		c.Rename(oldsid, sid)
+		c.Expire(sid, time.Duration(rp.maxlifetime))
```
flycash, line 227: It seems that `Expire key1 10` means that the key1 will be expired after 10 **SECOND**? so if we use `time.Duration(rp.maxlifetime)`, it will be a problem.

phiphi282: MaxLifetime is set to 3600 by default.
The time was being provided in seconds before as well so I think this should be fine.

I am not really sure why the sid gets expired in the first place though.

pkg/session/redis/sess_redis.go
phiphi282: Good catch 👍

pkg/session/redis/sess_redis.go
phiphi282: But there is a `* time.Second` missing there

phiphi282: Okay understood the Expiry. Because redis does not has a garbage collection defined, all entries get an expiry time.

PR #4147, opened 2020-08-10
specify index

jianzhiyao:


pkg/orm/types.go
```diff
@@ -527,24 +543,27 @@ type txEnder interface {
...
 	DeleteBatch(dbQuerier, *querySet, *modelInfo, *Condition, *time.Location) (int64, error)
-	Count(dbQuerier, *querySet, *modelInfo, *Condition, *time.Location) (int64, error)
+
+	SupportUpdateJoin() bool
 	OperatorSQL(string) string
 	GenerateOperatorSQL(*modelInfo, *fieldInfo, string, []interface{}, *time.Location) (string, []interface{})
 	GenerateOperatorLeftCol(*fieldInfo, string, *string)
 	PrepareInsert(dbQuerier, *modelInfo) (stmtQuerier, string, error)
-	ReadValues(dbQuerier, *querySet, *modelInfo, *Condition, []string, interface{}, *time.Location) (int64, error)
-	RowsTo(dbQuerier, *querySet, *modelInfo, *Condition, interface{}, string, string, *time.Location) (int64, error)
```
flycash, line 547: So, this method is useless?

pkg/common/kv.go
```diff
@@ -36,14 +36,25 @@ func (s *SimpleKV) GetValue() interface{} {
...
 
-// KVs will store SimpleKV collection as map
-type KVs struct {
+// KVs interface
+type KVs interface {
+	GetValueOr(key interface{}, defValue interface{}) interface{}
+	Contains(key interface{}) bool
+	IfContains(key interface{}, action func(value interface{})) KVs
+	Put(key interface{}, value interface{}) KVs
+	Clone() KVs
```
flycash, line 45: Why do we need this? It looks like we don't use this method.

pkg/common/kv.go
```diff
@@ -36,14 +36,25 @@ func (s *SimpleKV) GetValue() interface{} {
...
 }
 
-// KVs will store SimpleKV collection as map
-type KVs struct {
+// KVs interface
+type KVs interface {
+	GetValueOr(key interface{}, defValue interface{}) interface{}
+	Contains(key interface{}) bool
+	IfContains(key interface{}, action func(value interface{})) KVs
+	Put(key interface{}, value interface{}) KVs
```
flycash, line 44: I think this KVs should be immutable, so I don't want to expose any method that could modify Kvs

Issue #3830, opened 2019-10-16
数据库类型postgres，主键类型string，insert操作返回 0，LastInsertId is not supported by this driver


YungRun:
Please answer these questions before submitting your issue. Thanks!

1. What version of Go and beego are you using (`bee version`)?
```
go version go1.12.7 linux/amd64
github.com/astaxie/beego v1.12.0
github.com/lib/pq v1.2.0
```

2. What operating system and processor architecture are you using (`go env`)?

```
GOARCH="amd64"
GOBIN="/usr/local/go/bin:/home/summer/go/bin"
GOCACHE="/home/summer/.cache/go-build"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/summer/go"
GOPROXY="https://goproxy.io"
GORACE=""
GOROOT="/usr/local/go"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/home/summer/workspace/gomod/pgteset/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build837144964=/tmp/go-build -gno-record-gcc-switches"

```
3. What did you do?
```golang
package main

import (
	"fmt"
	"github.com/astaxie/beego/orm"
	_ "github.com/lib/pq"
)

type User struct {
	Id      string   `orm:"column(id);size(64);pk"`
	Name    string
	Profile *Profile `orm:"rel(one)"`      // OneToOne relation
	Post    []*Post  `orm:"reverse(many)"` // 设置一对多的反向关系
}

type Profile struct {
	Id   string `orm:"column(id);size(64);pk"`
	Age  int16
	User *User `orm:"reverse(one)"` // 设置一对一反向关系(可选)
}

type Post struct {
	Id    string  `orm:"column(id);size(64);pk"`
	Title string
	User  *User  `orm:"rel(fk)"` //设置一对多关系
[...]

PR #4158, opened 2020-08-14
fix:return error when calling ``InsertOrUpdate`` is successful with string primary key

jianzhiyao:


jianzhiyao:
https://github.com/astaxie/beego/issues/3830

pkg/orm/orm.go
```diff
@@ -85,6 +85,8 @@ var (
 	ErrStmtClosed    = errors.New("<QuerySeter> stmt already closed")
 	ErrArgs          = errors.New("<Ormer> args error may be empty")
 	ErrNotImplement  = errors.New("have not implement")
+
+	ErrLastInsertIdUnavailable = errors.New("<Ormer> last insert id is unavailable")
```
flycash, line 89: If we define this error, why not just return this error so that Users can check this error. It looks like:
```golang
 err := InsertOrUpdate(xxx)
if err == ErrLastInsertIdUnavailable {
    // do nothing
} else {
   // really failed.
}
```

PR #4160, opened 2020-08-14
fix issue #3776

### LINKED ISSUE #3776: orm.RawSeter does not support orm.Fielder

1.12.0

1.12.1

3. What did you do?
```golang
type SharedStorage struct {
        ...
	StorageAttribute StorageAttr `orm:"column(storage_attribute);type(text)"`
}
type StorageAttr struct {
	FileSysId    string            `json:"fileSystemId"`
	StorageSize  string            `json:"storageSize"`
	MountDomain  string            `json:"mountDomain"`
	VolumeLabels map[string]string `json:"volumeLabels"`
}

var _ orm.Fielder = new(StorageAttr)

func (e *StorageAttr) String() string {
	var data []byte
	if e == nil {
		data, _ = json.Marshal(struct{}{})
	} else {
		data, _ = json.Marshal(e)
	}
	return string(data)
}

func (e *StorageAttr) FieldType() int {
	return orm.TypeTextField
}

func (e *StorageAttr) SetRaw(value interface{}) error {
	switch d := value.(type) {
	case string:
		return json.Unmarshal([]byte(d), e)
	case []byte:
		return json.Unmarshal(d, e)
	default:
		return fmt.Errorf("<JsonField.SetRaw> unknown value `%v`", value)
	}
}

func (e *StorageAttr) RawValue() interface{} {
	return e.String()
}
```

orm.NewOrm().Raw("select * from dual").QueryRows(&rows)
this line of code should deserialize the StorageAttribute field with values.

5. What did you see instead?
however what I get is just an empty struct.

AllenX2018:
fixes #3776

fix the bug that `Fielder`'s `SetRaw` is not called when calling `orm.Raw()` to query from database

### DIFF
--- a/pkg/orm/orm_raw.go
+++ b/pkg/orm/orm_raw.go
@@ -17,6 +17,7 @@ package orm
 import (
 	"database/sql"
 	"fmt"
+	"github.com/pkg/errors"
 	"reflect"
 	"time"
 )
@@ -369,7 +370,15 @@ func (o *rawSet) QueryRow(containers ...interface{}) error {
 							field.Set(mf)
 							field = mf.Elem().FieldByIndex(fi.relModelInfo.fields.pk.fieldIndex)
 						}
-						o.setFieldValue(field, value)
+						if fi.isFielder {
+							fd := field.Addr().Interface().(Fielder)
+							err := fd.SetRaw(value)
+							if err != nil {
+								return errors.Errorf("set raw error:%s", err)
+							}
+						} else {
+							o.setFieldValue(field, value)
+						}
 					}
 				}
 			} else {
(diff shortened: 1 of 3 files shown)


AllenX2018:
Seems that the unit test will fail with postgresql. Pls wait and I will fix it.

pkg/orm/orm_raw.go
```diff
@@ -369,7 +369,15 @@ func (o *rawSet) QueryRow(containers ...interface{}) error {
 							field.Set(mf)
 							field = mf.Elem().FieldByIndex(fi.relModelInfo.fields.pk.fieldIndex)
 						}
-						o.setFieldValue(field, value)
+						if fi.isFielder {
+							fd := field.Addr().Interface().(Fielder)
+							err := fd.SetRaw(value)
+							if err != nil {
+								return fmt.Errorf("set raw error:%s", err)
```
flycash, line 376: What about using `github.com/pkg/errors`, we can change it in the future. I will check this PR this weekend.

AllenX2018: Sure. But I wonder what's the difference between `errors.New()` and `fmt.Errorf()` ?

PR #4173, opened 2020-08-19
Fix issue 3866

AllenX2018:
#3866 
Fix the problem that the nested structure of `queryRow()` cannot assign values.


AllenX2018:
Seems like the ut fail with sqlite3/postgresql as orm source. I will check it.

AllenX2018:
I've fixed the ci failure. Please review it. @flycash

flycash (approved):

PR #4177, opened 2020-08-20
Refactor SessionInit method

AllenX2018:
I change the input config string to json and declare provider config struct for every session provider. I think it's enough to just support json as the config string for now. Please review it and if these changes are approved, I will add UT for every provider. @flycash

### DIFF
--- a/pkg/session/couchbase/sess_couchbase.go
+++ b/pkg/session/couchbase/sess_couchbase.go
@@ -33,8 +33,8 @@
 package couchbase
 
 import (
+	"encoding/json"
 	"net/http"
-	"strings"
 	"sync"
 
 	couchbase "github.com/couchbase/go-couchbase"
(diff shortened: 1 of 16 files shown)


flycash (commented):
It looks good. 
But I think we should keep compatible with "savePath",  which means that we check the input. If input is in JSON format, we unmarshal it or we go back to old logic. In old logic, we log `Error` (or Warn) to tell users that we only support json format in v2.1.x.

AllenX2018:
> Yes. I think the logs could be the basic module so that it could be used in other modules.

OK. I will add some commit to keep compatible  with the old logic.

> We want to reorganize all modules. So there are many conflicts, don't worry about it. We just move files and do not change the content. It's easy to resolve these conflicts.

No Problem. I will update my fork repo and resolve the conflict myself.

flycash:
> It looks good.
> But I think we should keep compatible with "savePath", which means that we check the input. If input is in JSON format, we unmarshal it or we go back to old logic. In old logic, we log `Error` (or Warn) to tell users that we only support json format in v2.1.x.

@AllenX2018 Just ignore this since we had developed an adapter layer to keep compatible with v1.x. You just need to resolve those conflicts.

AllenX2018:
@flycash ok

PR #4188, opened 2020-08-24
Custom logging format PR#3

IamCathal:
## Features

Final PR for the custom log formatter

* Implements `SetLoggerWithOpt` func to allow users to define custom logging formats. Users only need to provide a function with the following signature
```go
func (lm *logs.LogMsg) string
```
* Implements `SetGlobalFormatter` to allow users to set a global formatter if they so choose. Users again use the same signature 
```go
func (lm *logs.LogMsg) string
```

The hierarchy goes like this:
### adapter specific formatter > global formatter > default formatter
i.e adapter specific can override global formatters and global formatters can override default set formatters

## Example code
```go
package main

import (
	"fmt"

	beego "github.com/astaxie/beego/pkg"
	"github.com/astaxie/beego/pkg/logs"
)

type MainController struct {
	beego.Controller
}

func customFormatter(lm *logs.LogMsg) string {
	return fmt.Sprintf("[CUSTOM CONSOLE LOGGING] %s", lm.Msg)
}

func customFormatter2(lm *logs.LogMsg) string {
	return fmt.Sprintf("[CUSTOM FILE LOGGING] %s", lm.Msg)
}

func GlobalFormatter(lm *logs.LogMsg) string {
	return fmt.Sprintf("[GLOBAL] %s", lm.Msg)
}

func main() {

	// GlobalFormatter only overrides default log adapters. Hierarchy is like this:
	// adapter specific formatter > global formatter > default formatter
	logs.SetGlobalFormatter(GlobalFormatter)

	// logs.SetLogger("console", "")
	// logs.SetLogger("file", `{"filename":"test.json"}`)

	logs.SetLoggerWithOpts("console", []string{""}, customFormatter)
	logs.SetLoggerWithOpts("file", []string{`{"filename":"test.json"}`}, customFormatter2)

	logs.EnableFullFilePath(false)
	logs.EnableFuncCallDepth(true)
	beego.Run()
}
```

### DIFF
--- a/pkg/logs/console.go
+++ b/pkg/logs/console.go
@@ -81,7 +81,6 @@ func NewConsole() Logger {
 
 // Init initianlizes the console logger.
 // jsonConfig must be in the format '{"level":LevelTrace}'
-// func (c *consoleWriter) Init(jsonConfig string, LogFormatter ...func(*LogMsg) string) error {
 func (c *consoleWriter) Init(jsonConfig string, opts ...common.SimpleKV) error {
 
 	for _, elem := range opts {
--- a/pkg/logs/console.go
+++ b/pkg/logs/console.go
@@ -19,6 +19,8 @@ import (
 	"os"
 	"strings"
 
+	"github.com/astaxie/beego/pkg/common"
+
 	"github.com/shiena/ansicolor"
 )
 
@@ -47,11 +49,10 @@ var colors = []brush{
 
 // consoleWriter implements LoggerInterface and writes messages to terminal.
 type consoleWriter struct {
-	lg                 *logWriter
-	UseCustomFormatter bool
-	CustomFormatter    func(*LogMsg) string
-	Level              int  `json:"level"`
-	Colorful           bool `json:"color"` //this filed is useful only when system's terminal supports color
+	lg              *logWriter
+	customFormatter func(*LogMsg) string
+	Level           int  `json:"level"`
+	Colorful        bool `json:"color"` //this filed is useful only when system's terminal supports color
 }
 
 func (c *consoleWriter) Format(lm *LogMsg) string {
@@ -80,11 +81,16 @@ func NewConsole() Logger {
 
 // Init initianlizes the console logger.
 // jsonConfig must be in the format '{"level":LevelTrace}'
-func (c *consoleWriter) Init(jsonConfig string, LogFormatter ...func(*LogMsg) string) error {
-	for _, elem := range LogFormatter {
-		if elem != nil {
-			c.UseCustomFormatter = true
-			c.CustomFormatter = elem
+// func (c *consoleWriter) Init(jsonConfig string, LogFormatter ...func(*LogMsg) string) error {
+func (c *consoleWriter) Init(jsonConfig string, opts ...common.SimpleKV) error {
+
+	for _, elem := range opts {
+		if elem.Key == "formatter" {
+			formatter, err := GetFormatter(elem)
+			if err != nil {
+				return err
+			}
+			c.customFormatter = formatter
 		}
 	}
 
@@ -107,10 +113,11 @@ func (c *consoleWriter) WriteMsg(lm *LogMsg) error {
 
 	msg := ""
 
-	if c.UseCustomFormatter {
-		msg = c.CustomFormatter(lm)
+	if c.customFormatter != nil {
+		msg = c.customFormatter(lm)
 	} else {
 		msg = c.Format(lm)
+
 	}
 
 	c.lg.writeln(msg)
--- a/pkg/logs/console.go
+++ b/pkg/logs/console.go
@@ -47,9 +47,11 @@ var colors = []brush{
 
 // consoleWriter implements LoggerInterface and writes messages to terminal.
 type consoleWriter struct {
-	lg       *logWriter
-	Level    int  `json:"level"`
-	Colorful bool `json:"color"` //this filed is useful only when system's terminal supports color
+	lg                 *logWriter
+	UseCustomFormatter bool
+	CustomFormatter    func(*LogMsg) string
+	Level              int  `json:"level"`
+	Colorful           bool `json:"color"` //this filed is useful only when system's terminal supports color
 }
 
 func (c *consoleWriter) Format(lm *LogMsg) string {
@@ -62,7 +64,7 @@ func (c *consoleWriter) Format(lm *LogMsg) string {
 	h, _, _ := formatTimeHeader(lm.When)
 	bytes := append(append(h, msg...), '\n')
 
-	return "eee" + string(bytes)
+	return string(bytes)
 
 }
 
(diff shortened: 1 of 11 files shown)


pkg/logs/console.go
```diff
@@ -47,9 +47,11 @@ var colors = []brush{
 
 // consoleWriter implements LoggerInterface and writes messages to terminal.
 type consoleWriter struct {
-	lg       *logWriter
-	Level    int  `json:"level"`
-	Colorful bool `json:"color"` //this filed is useful only when system's terminal supports color
+	lg                 *logWriter
+	UseCustomFormatter bool
```
flycash, line 51: we don't need to use those two fields. We could do this like:

```go
type consoleWriter struct {
     lg *logWriter
     formatter LogFormatter // use the interface defined in your last PR
    Formatter string `json:"formatter"`
    Level int
    Colorful bool
}

func (c *consoleWriter) Init(jsonConfig string) error {
	if len(jsonConfig) == 0 {
		return nil
	}
        res := json.Unmarshal([]byte(jsonConfig), c)
        if res.Formatter != "" {
                // we need two more methods, RegisterLogFormatter(name string, formatter LogFormatter) and 
                // GetFormatter(name string)
                res.formatter = GetFormatter(res.Formatter)
        } else {
                // In your last PR, the consoleWriter already implement LogFormatter interface.
                // if user do not use custom formatter, we use consoleWriter default formatter(itself)
                res.formatter = res
        }
	return res
}

// WriteMsg writes message in console.
func (c *consoleWriter) WriteMsg(lm *LogMsg) error {
	msg := c.formatter.Format(lm)
	c.lg.writeln(msg)
	return nil
}
```
Other implementations are similar to this.

IamCathal: One problem that I'm thinking about that if every adapter has a `LogFormatter` field it means it must have a `Format` method. This isn't an issue when you don't want to use custom formatting. In my last PR I hardcoded in a format method for each adapter like so:
```go
func (c *consoleWriter) Format(lm *LogMsg) string {
	msg := lm.Msg

	if c.Colorful {
		msg = strings.Replace(lm.Msg, levelPrefix[lm.Level], colors[lm.Level](levelPrefix[lm.Level]), 1)
	}

	h, _, _ := formatTimeHeader(lm.When)
	bytes := append(append(h, msg...), '\n')

	return string(bytes)

}
```
In order to get a custom formatter this `Format` function would need to be overrided somehow with the user defined logging formatter and I'm not sure how to do that.

I also think that putting the formatter in as its own parameter is better as it seperates the logic better
If the formatter was part of the config string it would be something like
```go
logs.SetLoggerWithOpts("console", []string{"something":"key","formatter":"customFormatter"})
// instead of 
logs.SetLoggerWithOpts("console", []string{""}, customFormatter)
```
I also don't know how you'd put a function into a JSON string then extract the a pointer to the function when you decode the JSON config in the `init` function

flycash: > One problem that I'm thinking about that if every adapter has a LogFormatter field it means it must have a Format method. This isn't an issue when you don't want to use custom formatting. In my last PR I hardcoded in a format method for each adapter like so:

My idea is that:
```go
func (c *consoleWriter) Format(lm *LogMsg) string {
	msg := lm.Msg

	if c.Colorful {
		msg = strings.Replace(lm.Msg, levelPrefix[lm.Level], colors[lm.Level](levelPrefix[lm.Level]), 1)
	}

	h, _, _ := formatTimeHeader(lm.When)
	bytes := append(append(h, msg...), '\n')

	return string(bytes)

}
```

Making this method as the default formatter for consoleWritter.
I prefer to define a new default formatter for consoleWritter and remove `Format` method from `Logger` interface.

```
type consoleWritterDefaultFormatter {
     colorful bool
}


func (c *consoleWritterDefaultFormatter) Format(lm *LogMsg) string {
	msg := lm.Msg

	if c.colorful{
		msg = strings.Replace(lm.Msg, levelPrefix[lm.Level], colors[lm.Level](levelPrefix[lm.Level]), 1)
	}

	h, _, _ := formatTimeHeader(lm.When)
	bytes := append(append(h, msg...), '\n')

	return string(bytes)
}

type consoleWriter struct {
     lg *logWriter
     formatter LogFormatter // use the interface defined in your last PR
    Level int
    Colorful bool
}

// using "opts", so that not only LogFormatter but also other parameter could be passed.
// Orm use common.Kv as options. But it's not golang typical practice. 
func (c *consoleWriter) Init(jsonConfig string, opts...Option) error {
	if len(jsonConfig) == 0 {
		return nil
	}
        res := json.Unmarshal([]byte(jsonConfig), c)
        formatter := findFormatter(opts)
        if formatter != nil  {
                res.formatter = formatter
        } else {
                res.formatter = &consoleWritterDefaultFormatter{
                      colorful: res.Colorful
                }
        }
	return res
}

// WriteMsg writes message in console.
func (c *consoleWriter) WriteMsg(lm *LogMsg) error {
	msg := c.formatter.Format(lm)
	c.lg.writeln(msg)
	return nil
}

```

The idea is that we use `LogFormatter` too.

IamCathal: Ok this is quite similar to my 2nd implementation. I'll get to work on it

flycash:
Please add unit tests for those changes

pkg/logs/console.go
```diff
@@ -94,7 +109,16 @@ func (c *consoleWriter) WriteMsg(lm *LogMsg) error {
 	if c.Colorful {
 		lm.Msg = strings.Replace(lm.Msg, levelPrefix[lm.Level], colors[lm.Level](levelPrefix[lm.Level]), 1)
 	}
-	msg := c.Format(lm)
+
```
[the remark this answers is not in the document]
IamCathal: Actually this if colorful check needs to be in this `WriteMsg` function before the `LogMsg` is sent to either the default or custom logging formatter. 

If the user is using a custom formatter their function only has access to the `LogMsg` and not the console writer (where `Colorful` is a field). To keep it simple for the user the message is replaced with it's colorful version before it's sent to the custom logging formatter implemented by the user

IamCathal:
Added simple tests ( `pkg/logs/logformattertest` ) to check if the custom logging formatting is working. Let me know if this is or isn't enough.



```go
func TestCustomLoggingFormatter(t *testing.T) {
	// beego.BConfig.Log.AccessLogs = true

	logs.SetLoggerWithOpts("console", []string{`{"color":true}`}, common.SimpleKV{Key: "formatter", Value: customFormatter})

	// Message will be formatted by the customFormatter with colorful text set to true
	logs.Informational("Test message")
}

func TestGlobalLoggingFormatter(t *testing.T) {
	logs.SetGlobalFormatter(globalFormatter)

	logs.SetLogger("console", `{"color":true}`)

	// Message will be formatted by globalFormatter
	logs.Informational("Test message")

}

```

flycash (approved):

flycash:
I will merge ftr/log_format to develop-2.0 this weekend or next week. It may be hard for you to resolve conflicts. And you can make pr to ftr/log_format too.

IamCathal:
I think this should be it for the implementation. I'll make a PR on [beedoc](https://github.com/beego/beedoc) with the updated documentation for this new feature. Thanks for all the feedback over the 3 different version of this new feature :)

flycash:
> I think this should be it for the implementation. I'll make a PR on [beedoc](https://github.com/beego/beedoc) with the updated documentation for this new feature. Thanks for all the feedback over the 3 different version of this new feature :)

ok, but I do some refactor, please check it.

PR #4198, opened 2020-08-30
Enhancement/Add a new MySQL operator for strict case sensitive query

CadenGuo:
Using `BINARY` key for `exact` operator in MySQL query.<br/>
Since the default collation shipped from MySQL is case insensitive and accent insensitive, making `exact` always case sensitive could avoid confusions.<br/>
Besides, adding a `BINARY` to `exact` operator makes it align with other operators, like [contains](https://github.com/astaxie/beego/blob/f946a35acd6fa002da765cfbdcda9c393b2f30ee/orm/db_mysql.go#L27) operator.

### DIFF
--- a/orm/db_mysql.go
+++ b/orm/db_mysql.go
@@ -22,10 +22,11 @@ import (
 
 // mysql operators.
 var mysqlOperators = map[string]string{
-	"exact":     "= BINARY ?",
-	"iexact":    "LIKE ?",
-	"contains":  "LIKE BINARY ?",
-	"icontains": "LIKE ?",
+	"exact":       "= ?",
+	"iexact":      "LIKE ?",
+	"strictexact": "= BINARY ?",
+	"contains":    "LIKE BINARY ?",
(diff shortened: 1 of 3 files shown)


orm/db_mysql.go
```diff
@@ -22,7 +22,7 @@ import (
 
 // mysql operators.
 var mysqlOperators = map[string]string{
-	"exact":     "= ?",
+	"exact":     "= BINARY ?",
```
flycash, line 25: I think the better solution is adding a new operator.

CadenGuo: Hi, I have added a MySQL operator `strictexact` to implement `BINARY` keyword. Please kindly advice on the naming, related testing, etc. Thank you.

flycash (approved):

CadenGuo:
Hi, I have change the orm_test.go to skip test for `stricexact` operator on db drivers other than mysql. While the test always fails on some code that's not written by me. Still find no reasons. Please kindly advice. Thank you.

PR #4220, opened 2020-09-11
refactor log module

flycash:
I remove the utils.KV from logs module.


IamCathal:
I see that `pkg/infrastructure/logs/log_formatter_test.go` was removed in this PR. I'm guessing this is because it was failing the tests right? Because of the way it's being tested in my PR I had to have it in it's own package (`logformattertest`) due to cyclic dependencies instead of using the `logs` package. Maybe there is a simpler way around this but I'm not sure.

The change to using utils.KV is also fine and apart from this first test issue it looks fine to me

flycash:
> I see that `pkg/infrastructure/logs/log_formatter_test.go` was removed in this PR. I'm guessing this is because it was failing the tests right? Because of the way it's being tested in my PR I had to have it in it's own package (`logformattertest`) due to cyclic dependencies instead of using the `logs` package. Maybe there is a simpler way around this but I'm not sure.
> 
> The change to using utils.KV is also fine and apart from this first test issue it looks fine to me

I am writing tests for logs module. I will put it back after I finish this job.

Issue #4222, opened 2020-09-12
Empty field in validator.Error when label struct tag is not declared


HITWHTigerLiu:
Please answer these questions before submitting your issue. Thanks!

1. What version of Go and beego are you using (`bee version`)?
go version: 1.14
beego: 1.12.2

2. What operating system and processor architecture are you using (`go env`)?
windows

3. What did you do?
the issue #4001 is not be fixed yet.
using validator but errors with empty field when validating failed.
It seems that the bug is caused when label struct tag is empty.

beego/validation/validation.go

https://github.com/astaxie/beego/blob/8f3d1c5f42fce57e83e1c3f7d180477595db7cca/validation/validation.go#L272
What's more, there is no introduction about the new label tag, and tells me it's a must to declare it.
len == 3 caused the new label tag must to declare it，otherwise the code that did not use the label tag will be incompatible, resulting in empty feild.

4. What did you expect to see?
When validating failed and the Label tag not used,non-empty field should be return no matter label is declared or not.

5. What did you see instead?
empty field in validator error

PR #4225, opened 2020-09-14
Empty field in validator.Error when label struct tag is not declared #4222

HITWHTigerLiu:
to fix issue #4222

### DIFF
--- a/pkg/infrastructure/validation/validation_test.go
+++ b/pkg/infrastructure/validation/validation_test.go
@@ -607,3 +607,28 @@ func TestCanSkipAlso(t *testing.T) {
 	}
 
 }
+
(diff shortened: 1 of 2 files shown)


pkg/infrastructure/validation/validation_test.go
```diff
@@ -607,3 +607,28 @@ func TestCanSkipAlso(t *testing.T) {
 	}
 
 }
+
+func TestFieldNoEmpty(t *testing.T) {
```
flycash, line 611: I think we should build invalid user to check the error message.

HITWHTigerLiu: it is a invalid user already，i check the length and feild of the error.

flycash (approved):

PR #4251, opened 2020-10-06
Fix typo

sc0Vu:
Fix typo


admin_test.go
```diff
@@ -154,7 +154,7 @@ func TestBuildHealthCheckResponseList(t *testing.T) {
 		[]string{
 			"error",
 			"Database",
-			"Error occured whie starting the db",
+			"Error occurred whie starting the db",
```
flycash, line 157: whie => while ?

sc0Vu: thanks,  miss that word!

PR #4259, opened 2020-10-10
Add MaxFileSize configuration

### LINKED ISSUE #4224: multipart form数据自动缓存可能导致磁盘被恶意提交的数据占满，能否增加开关控制
描述：
我发现框架会自动在接口路由匹配前将multipart form数据缓存到磁盘中（当大于设定的允许内存读取大小时），并且缓存时是没有对大小进行限制的。该设计可能会导致客户端通过少量请求恶意提交大量数据将服务端磁盘临时占满。

当前处理方式：
增加一个BeforeStatic的过滤器，用于拦截所有携带multipart form数据的请求。不过该场景还是比较通用的，让每个使用beego框架的项目增加这个逻辑会比较冗余。

需求：
能否增加一个配置项用来控制是否由框架自动缓存呢？关闭的情况下，如果接口需要缓存，可以自行在handler中调用parsemultiform方法进行缓存。

相关代码行：
https://github.com/astaxie/beego/blob/f6519b29a846bdf59a2b86baa011c242f78387d5/router.go#L753

AllenX2018:
The Go stdlib multipart has no limitation while parsing multipart file.

Fix #4224

### DIFF
--- a/router.go
+++ b/router.go
@@ -750,6 +750,12 @@ func (p *ControllerRegister) ServeHTTP(rw http.ResponseWriter, r *http.Request) 
 			}
 			context.Input.CopyBody(BConfig.MaxMemory)
 		}
+
+		if r.ContentLength > BConfig.MaxFileSize {
+			logs.Error(errors.New("payload too large"))
+			exception("413", context)
+			goto Admin
+		}
 		context.Input.ParseFormOrMulitForm(BConfig.MaxMemory)
 	}
 
--- a/config.go
+++ b/config.go
@@ -42,6 +42,7 @@ type Config struct {
 	CopyRequestBody     bool
 	EnableGzip          bool
 	MaxMemory           int64
+	MaxFileSize         int64
 	EnableErrorsShow    bool
 	EnableErrorsRender  bool
 	Listen              Listen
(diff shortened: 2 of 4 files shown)


router.go
```diff
@@ -750,6 +750,12 @@ func (p *ControllerRegister) ServeHTTP(rw http.ResponseWriter, r *http.Request)
 			}
 			context.Input.CopyBody(BConfig.MaxMemory)
 		}
+
+		if r.ContentLength > BConfig.MaxFileSize {
```
flycash, line 754: So if I upload a file with 129MB, does it fail?

AllenX2018: Yes.
It will return http status 413 if file size is over 128MB by default.

flycash: So this behavior is not acceptable. For big files, we need to think about how to upload it by splitting into several parts.

AllenX2018: No, the "multipart" DOES NOT limit file size.
The OP wants limitation on file size and that is quite normal.

flycash (commented):
I will test it this weekend

AllenX2018:
ping

flycash:
#4272

PR #4272, opened 2020-10-18
fix 4224：form entity too large casue run out of memory

jianzhiyao:
close https://github.com/astaxie/beego/issues/4224


flycash (approved):
Good! I want to explain more. This PR limit the max file size. If its size is over MaxMemory, the request will be rejected. I believe that this makes the application safer and more stable.

If you want to upload large files, you need to think about splitting the file into multiple chunks.

Issue #4144, opened 2020-08-08
Support order by random()


khanzadimahdi:
how can i order records by random using orm?

flycash:
Do you use `QuerySetter`? It contains method:
```go
	// add ORDER expression.
	// "column" means ASC, "-column" means DESC.
	// for example:
	//	qs.OrderBy("-status")
	OrderBy(exprs ...string) QuerySeter
```

khanzadimahdi:
yes, however `RAND()` not works and generates error.

```go
OrderBy("RAND()")
```

flycash:
I think you can use `QueryRaw`

khanzadimahdi:
@flycash 
I want to filter my Model, using `QueryRaw` is not a clean way.

flycash:
We will fix it in the future. We want to release v2.0.0-beta this weekend, and we don't have enough time to fix it. So it will be fixed in v2.0.1.

khanzadimahdi:
so please add `UpdateOrCreate` method too.

flycash:
There is a similar method `InsertOrUpdate`.

khanzadimahdi:
the concept of `insertOrUpdate` is wrong!!! it means insert it and if it failed, update it!!!

i want the concept of `update it, and if it failed, create it`

flycash:
> the concept of `insertOrUpdate` is wrong!!! it means `insert it and if it failed, update it`!!!
> 
> i want the concept of `update it, and if it failed, create it`
> 
> it must be **`UpdateOrInsert`**

Please correct me if I make some mistakes.

I think there is no SQL like `Update xxx Or Insert`. Actually, the `InsertOrUpdate` is implemented by using `Insert xxx ON DUPLICATE KEY` in MYSQL and `Insert XXX ON CONFLICT (XXXX) DO UPDATE SET`.

If we want to provide `UpdateOrInsert`, we need to start a transaction to do an update. And if we find affected rawnum is 0 and then we do an insert.

I think in most cases, `insertOrUpdate` is good.

So, could you tell me the scenario that you can't use `insertOrUpdate`?

khanzadimahdi:
I asked for `UpdateOrCreate` method and you said me, use `InsertOrUpdate `.

`UpdateOrCreate` means ==> Update the record if exists, or create a new record.

flycash:
> You may come across situations where you want to update an existing model or create a new model if none exists.

[...]

PR #4279, opened 2020-10-25
supports_for_4144

jianzhiyao:
close https://github.com/astaxie/beego/issues/4144


client/orm/structs/clauses/order.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	return ``
+}
+
+func (o *Order) IsRaw() bool {
+	return o.isRaw
+}
+
+func ParseOrder(expressions ...string) []*Order {
```
flycash, line 53: Please add more unit tests to test this method. And I think that putting this file in the orm directory is better.
The unit test should demo all cases. It's hard for me to understand how this method work.

And how about add `error` as return value? I think sometimes users may pass some invalid expressions.

client/orm/types.go
```diff
@@ -289,6 +290,22 @@ type QuerySeter interface {
...
+	// for example:
+	// OrderClauses(clauses.OrderClause(
+	//		clauses.OrderColumn(`status`),
+	//		clauses.OrderSortAscending(),//default None
+	//	))
+	// OrderClauses(clauses.OrderClause(
+	//		clauses.OrderColumn(`user__status`),
+	//		clauses.OrderSortDescending(),//default None
+	//	))
+	// OrderClauses(clauses.OrderClause(
```
flycash, line 303: When I pass multiple columns, does it work?

flycash (commented):
Need more unit tests

PR #4300, opened 2020-11-10
Enhance issue 4209

### LINKED ISSUE #4209: How to use aggregated function, sum, count...

3. What did you do?

5. What did you see instead?

AllenX2018:
close #4209 
1.support dynamic registration model.
2.support aggregate func


flycash (commented):
It will be released with v2.0.1.

client/orm/models.go
```diff
@@ -353,6 +353,9 @@ func (mc *_modelCache) register(prefixOrSuffixStr string, prefixOrSuffix bool, m
 			return
 		}
 
+		if val.Elem().Kind() == reflect.Slice {
```
flycash, line 356: I can not understand this part. Do you mean that model could be Array or Slice?

client/orm/models.go
```diff
@@ -389,12 +392,6 @@ func (mc *_modelCache) register(prefixOrSuffixStr string, prefixOrSuffix bool, m
 					}
 				}
 			}
-
-			if mi.fields.pk == nil {
```
flycash, line 393: I think output some warn information would be better

client/orm/types.go
```diff
@@ -405,6 +405,15 @@ type QuerySeter interface {
...
 	RowsToStruct(ptrStruct interface{}, keyCol, valueCol string) (int64, error)
+	// aggregate func.
+	// for example:
+	// type result struct {
+	//  DeptName string
+	//	Total    int
+	// }
+	// var res []result
+	//  o.QueryTable("dept_info").Aggregate("dept_name,sum(salary) as total").GroupBy("dept_name").All(&res)
+	Aggregate(s string) QuerySeter
```
flycash, line 416: From this method, my understanding is that we could not use `qs.Count(XXX).GroupBy('xxx').All`?

client/orm/models.go
AllenX2018: Yes, for example
```go
var sun []Sum
orm.Aggregate("dept_name,sum(salary) as total").GroupBy("dept_name").OrderBy("dept_name").All(&sun)
```
`[]Sum` is a dynamic registration model, so add this judgment.

client/orm/types.go
AllenX2018: This method returns two values: the total number of queries and error information.
It is different from the aggregate function `count`.
https://github.com/astaxie/beego/blob/26bc9e72645344d65dc79fd39f2ec0dddeb3d617/client/orm/types.go#L326-L329

client/orm/models.go
AllenX2018: What kind of warning message is output?

client/orm/models.go
flycash: How about 
```
var sun []Sum
Register(&Sum{})
orm.Aggregate("dept_name,sum(salary) as total").GroupBy("dept_name").OrderBy("dept_name").All(&sun)
```
Because I think registering a slice looks not good.

client/orm/models.go
flycash: ("<orm.RegisterModel> `%s`lack of a primary key field, default is to use 'id' if not set\n", name)

client/orm/models.go
AllenX2018: No, this logic is to extract the struct information.
```go
if val.Elem().Kind() == reflect.Slice {
   val = reflect.New(val.Elem().Type().Elem())
}
```

client/orm/models.go
AllenX2018: ok

flycash (commented):
It crash when I run test:
```go
type User struct {
	Id       int
	Username string
	Password string
	Gender string
}

type CntResult struct {
	Cnt int
	Gender string
}

func CountMale() []int {
	s := orm.NewOrm().QueryTable(&User{})
	var cnts []CntResult
	_, err := s.Aggregate("gender, count(id) as cnt").GroupBy("gender").All(&cnts)
	if err != nil {
		logs.Error("could not get aggregate result: %+v", err)
		return nil
	}
	res := make([]int, 0, len(cnts))
	for _, i := range cnts {
		res = append(res, i.Cnt)
	}
	return res
}
```
When I run this test multiple times, got error:
```shell
2021/01/01 21:41:23.719 [C] [config.go:193]  Handler crashed with error <orm.RegisterModel> table name `cnt_result` repeat register, must be unique


2021/01/01 21:41:23.719 [C] [config.go:199]  /usr/local/go/src/runtime/panic.go:975

2021/01/01 21:41:23.719 [C] [config.go:199]  /Users/mindeng/go-workspace/src/beego/client/orm/models_boot.go:25

2021/01/01 21:41:23.719 [C] [config.go:199]  /Users/mindeng/go-workspace/src/beego/client/orm/models_boot.go:19

2021/01/01 21:41:23.719 [C] [config.go:199]  /Users/mindeng/go-workspace/src/beego/client/orm/db.go:977

2021/01/01 21:41:23.719 [C] [config.go:199]  /Users/mindeng/go-workspace/src/beego/client/orm/orm_queryset.go:245

2021/01/01 21:41:23.719 [C] [config.go:199]  /Users/mindeng/go-workspace/src/hello-api/models/user.go:80

2021/01/01 21:41:23.720 [C] [config.go:199]  /Users/mindeng/go-workspace/src/hello-api/controllers/user.go:22

2021/01/01 21:41:23.720 [C] [config.go:199]  /usr/local/go/src/reflect/value.go:476

2021/01/01 21:41:23.720 [C] [config.go:199]  /usr/local/go/src/reflect/value.go:337

2021/01/01 21:41:23.720 [C] [config.go:199]  /Users/mindeng/go-workspace/src/beego/server/web/router.go:877

2021/01/01 21:41:23.720 [C] [config.go:199]  /Users/mindeng/go-workspace/src/beego/server/web/filter.go:81

2021/01/01 21:41:23.720 [C] [config.go:199]  /Users/mindeng/go-workspace/src/beego/server/web/router.go:658

2021/01/01 21:41:23.720 [C] [config.go:199]  /usr/local/go/src/net/http/server.go:2843

2021/01/01 21:41:23.720 [C] [config.go:199]  /usr/local/go/src/net/http/server.go:1925

2021/01/01 21:41:23.720 [C] [config.go:199]  /usr/local/go/src/runtime/asm_amd64.s:1374

```

PR #4303, opened 2020-11-11
fix issue #3430: reverse sequence of generating leaves of router prefix tree

AllenX2018:
fix issue #3430 

```go
// tree.go
// ...
		if len(params) > 0 && params[0] == ":" {
			t.addseg(segments[1:], route, wildcards, reg)
			params = params[1:]
		}
// ...
```
The recursive call of addseg here make the t.leaves is rightmost match when append slice t.leaves sequentially.
To make the t.leaves is leftmost match, this PR modify the append sequence of t.leaves.

In addition, test cases for verifying the leftmost matching of regular routes are provided in this PR.


flycash (commented):
I believe that this is a big update so I will merge it after releasing v2.0.0

Issue #4176, opened 2020-08-19
Proposal: Ignore session for some request even SessionOn = true


flycash:
#2787 #4169 

For now, we only have one global configuration SessionOn to control whether using session.

If SessionOn = true, all request except static resources carry session information. 

But sometimes, we want to close the session for some path.

For example, we have two routes, one is `/session/xxxx`, which means that we need session;
another is `/no-session/xxxx`.

But I don't have any good design. One simple way is that change `Router` method, add more option to control the behavior:
```go
func Router(rootpath string, c ControllerInterface, mappingMethods ...string) *App {
	BeeApp.Handlers.Add(rootpath, c, mappingMethods...)
	return BeeApp
}
```

```go
func Router(rootpath string, c ControllerInterface, opts Opts) *App {
	BeeApp.Handlers.Add(rootpath, c, mappingMethods...)
	return BeeApp
}
```

luudis:
I have the same demand and is there any further progress about this proposal?

flycash:
No, would you like to support this feature?

AllenX2018:
Hello

I came up with two options. 

### Modify interface
The first one is like you provided, change the original function
```go
func Router(rootpath string, c ControllerInterface, mappingMethods ...string) *HttpServer
```
to
```go
func Router(rootpath string, c ControllerInterface, opts Opts, mappingMethods ...string) *HttpServer
```
then use `opts` to transfer the bool value of the session switch.
However, this will change the `Router` interface. Will the users who switch from the earlier version need to modify the interface in a large scope?

### Add new interface
The second method is to add a new interface for configuring the session switch. For example,
```go
func Session(rootpath string, opts Opts, mappingMethod ...string) error
```
Calling this interface can search for the corresponding route in the prefix tree `routers` in `ControllerRegister` and set `sessionOn` in `ControllerInfo`。Correspondingly, the `sessionOn bool` field is added to the `ControllerInfo` structure.

Adding this interface means that route binding and route session setting are performed in two steps. However, this doesn't affect the original interface `Router`. In addition, if the user is migrated to the new version but doesn't call the session setting interface, the session is configured globally and doesn't affect the running of the original user code.

flycash:
> Hello
> 
> I came up with two options.
> 
> ### Modify interface
> The first one is like you provided, change the original function
> 
> ```go
> func Router(rootpath string, c ControllerInterface, mappingMethods ...string) *HttpServer
> ```
> 
> to
> 
> ```go
[...]

PR #4318, opened 2020-11-24
Fix issue #4176

AllenX2018:
Fix issue #4176 

1. Refactor `Router` with option pattern
2. Add `SessionOn` option for `web.Router`


flycash (commented):
Good! I will check it and merge it after release 2.0.0

server/web/router.go
```diff
@@ -171,40 +186,67 @@ func NewControllerRegisterWithCfg(cfg *Config) *ControllerRegister {
...
+
+func (p *ControllerRegister) addWithMethodParams(pattern string, c ControllerInterface, methodParams []*param.MethodParam, opts ...ControllerOptions) {
+	reflectVal := reflect.ValueOf(c)
+	t := reflect.Indirect(reflectVal).Type()
+
 	route := &ControllerInfo{}
 	route.pattern = pattern
-	route.methods = methods
 	route.routerType = routerTypeBeego
+	route.sessionOn = BConfig.WebConfig.Session.SessionOn
```
flycash, line 249: ControllerRegister has a field name 'cfg', please using it.

server/web/router.go
```diff
@@ -742,7 +778,16 @@ func (p *ControllerRegister) serveHttp(ctx *beecontext.Context) {
 	}
 
 	// session init
-	if p.cfg.WebConfig.Session.SessionOn {
+	currentSessionOn = BConfig.WebConfig.Session.SessionOn
+	originRouterInfo, originFindRouter = p.FindRouter(ctx)
+	if originFindRouter {
+		if !currentSessionOn && originRouterInfo.sessionOn {
```
flycash, line 784: could we calcaulte the value of `sessionOn` when we create this route?

flycash (commented):
Do not use global config BConfig anymore, it should only be used in initialization.

Please use ControllerRegister.cfg instead.

server/web/router.go
AllenX2018: OK

server/web/router.go
AllenX2018: Good idea. I'll try it.

jianzhiyao (approved):
checked

PR #4328, opened 2020-11-29
expose more error code in web module

flycash:
Many users complain that Beego hide error information.
So we will expose more error.


jianzhiyao (approved):
checked

PR #4380, opened 2020-12-19
fix-2840

jianzhiyao:
no Set-Cookie in response when sessionid is uploaded
close https://github.com/beego/beego/issues/2840


flycash:
We need a long term solution.

Issue #4166, opened 2020-08-18
Proposal-Cache usage pattern: cache-aside, write-through, read-through and more...


flycash:
Some common usage like
```
if Get(key) == nil {
    value := LoadFromDB(key)
    Put(key, value)
}

return Get(key)
```
I think we can provide some cache plugins to solve similar problems.
The above example is "cache-aside pattern". 

And there are many other patterns.
1. write-through pattern:  write cache firstly and then write storage;
2. read-through pattern: read from cache and if not found, cache load value from storage;
3. write-back pattern: write cache firstly and return. Cache will write data to storage asynchronously;
4. write-around pattern: write storage and then remove the key from cache;

All of those patterns may cause inconsistent data, but it's very useful.

AllenX2018:
It seems that current version doesn't support write/read between cache and database.
So this proposal is going to create new interface including method such as  `func Get(key string) string`, `func Put(key string, val  string)` and so on? 
And we will implement the interface to realize differnt patterns, such as "cache-aside pattern", "write-through pattern" and so on ?

flycash:
> It seems that current version doesn't support write/read between cache and database.
> So this proposal is going to create new interface including method such as `func Get(key string) string`, `func Put(key string, val string)` and so on?
> And we will implement the interface to realize differnt patterns, such as "cache-aside pattern", "write-through pattern" and so on ?

Actually we have methods:
```go
type Cache interface {
	// get cached value by key.
	Get(key string) interface{}
	// GetMulti is a batch version of Get.
	GetMulti(keys []string) []interface{}
	// set cached value with key and expire time.
	Put(key string, val interface{}, timeout time.Duration) error
	// delete cached value by key.
	Delete(key string) error
	// increase cached int value by key, as a counter.
	Incr(key string) error
	// decrease cached int value by key, as a counter.
	Decr(key string) error
	// check if cached value exists or not.
	IsExist(key string) bool
	// clear all cache.
	ClearAll() error
	// start gc routine based on config string settings.
	StartAndGC(config string) error
}
```
In fact, I think we need something like, taking `read-through` as example:
```go
type ReadThroughCache struct {
     dataReader XXXXType
     cache Cache
}
```

When we could not find the value with a specific key, we will use `dataReader.Read(key)`.
Usually, it means that `dataReader` read data from DB.

I think using `decorator` design pattern is pretty good.

AllenX2018:
If load from db,  establishing the relationship between a key value and an SQL statement is needed.

[...]

PR #4389, opened 2020-12-26
Fix issue #4166

AllenX2018:
Fix issue #4166 : Add cache-aside pattern

I only add cache-aside pattern in this pr.
If the implement of the cache-pattern in this pr is ok, I will try implement of other cache pattern, such as read-through, write-through...

### DIFF
--- a/client/cache/cache_pattern.go
+++ b/client/cache/cache_pattern.go
@@ -1,0 +1,110 @@
+package cache
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+)
+
+// ValueReader reads data from data source, such as database
+type ValueReader interface {
+	Query(key string) (interface{}, error)
+}
+
+// ValueWriter writes data to data source, such as database
+type ValueWriter interface {
+	Update(key string, val interface{}) error
+}
+
+// CacheAside is a decorator of Cache, it implements "cache aside" pattern
+type CacheAside struct {
+	Cache
+	reader   ValueReader
+	writer   ValueWriter
+	lifeSpan time.Duration
+}
+
+func NewCacheAside(cache Cache, reader ValueReader, writer ValueWriter, lifeSpan time.Duration) *CacheAside {
+	return &CacheAside{Cache: cache, reader: reader, writer: writer, lifeSpan: lifeSpan}
+}
+
+func (c *CacheAside) Get(ctx context.Context, key string) (interface{}, error) {
+	if val, _ := c.Cache.Get(ctx, key); val != nil {
+		return val, nil
+	}
+
+	val, err := c.reader.Query(key)
+	if err != nil {
+		return nil, err
+	}
+
+	return val, c.Cache.Put(ctx, key, val, c.lifeSpan)
+}
+
+func (c *CacheAside) GetMulti(ctx context.Context, keys []string) ([]interface{}, error) {
+	keysErr := make([]string, 0)
+	vals := make([]interface{}, len(keys))
+
+	for i := range keys {
+		val, err := c.Get(ctx, keys[i])
+		vals[i] = val
+		if err != nil {
+			keysErr = append(keysErr, err.Error())
+		}
+	}
+
+	if len(keysErr) == 0 {
+		return vals, nil
+	}
+
+	return vals, fmt.Errorf(strings.Join(keysErr, "; "))
+}
+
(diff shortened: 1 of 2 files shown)


client/cache/cache_pattern.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	if len(keysErr) == 0 {
+		return vals, nil
+	}
+
+	return vals, fmt.Errorf(strings.Join(keysErr, "; "))
+}
+
+func (c *CacheAside) Put(ctx context.Context, key string, val interface{}, timeout time.Duration) error {
```
jianzhiyao, line 64: this operations shold be locked?

AllenX2018: > this operations should be locked?

I think the lock will be implemented by `cache`, `reader` and `writer` passed to `NewCacheAside`.
For example, if we pass `MemoryCache` to `NewCacheAside`,  `CacheAside.Put` will use lock in `MemoryCache.Delete` while deleting cache.

client/cache/cache_pattern.go
```diff
@@ -0,0 +1,110 @@
...
+
+	val, err := c.reader.Query(key)
+	if err != nil {
+		return nil, err
+	}
+
+	return val, c.Cache.Put(ctx, key, val, c.lifeSpan)
+}
+
+func (c *CacheAside) GetMulti(ctx context.Context, keys []string) ([]interface{}, error) {
```
flycash, line 45: I think add a new method named "QueryBatch" into `ValueReader` would be better. Users always expect that the `CacheAside` work the same way as `Cache`.

For example, users create the `CacheAside` instance by passing a redis Cache instance which read all values  from redis server in one command `MGET`.

But this method change the behavior.

client/cache/cache_pattern.go
```diff
@@ -0,0 +1,110 @@
...
+
+// CacheAside is a decorator of Cache, it implements "cache aside" pattern
+type CacheAside struct {
+	Cache
+	reader   ValueReader
+	writer   ValueWriter
+	lifeSpan time.Duration
+}
+
+func NewCacheAside(cache Cache, reader ValueReader, writer ValueWriter, lifeSpan time.Duration) *CacheAside {
```
flycash, line 28: I think both reader, writer and lifeSpan are optional.

For example, users only want `CacheAside` read data from DB if the cache didn't have the specific key. But users update values by ORM directly:
```go

func demo() {
   aside.NewCacheAside(redisCache, ormReader)
   aside.Get(ctx, key1)
   orm.update(key1, xxx)
}

```

client/cache/cache_pattern.go
```diff
@@ -0,0 +1,110 @@
...
+func (c *CacheAside) Put(ctx context.Context, key string, val interface{}, timeout time.Duration) error {
+	if err := c.writer.Update(key, val); err != nil {
+		return err
+	}
+
+	if ok, _ := c.IsExist(ctx, key); !ok {
+		return nil
+	}
+
+	return c.Cache.Delete(ctx, key)
```
flycash, line 73: we don't need to check whether exist, just delete it.

client/cache/cache_pattern.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		return err
+	}
+
+	if err := c.writer.Update(key, updateVal); err != nil {
+		return err
+	}
+
+	return c.Cache.Delete(ctx, key)
+}
```
flycash, line 110: Both two methods have concurrency issue.

flycash (changes requested):
Please resolve the concurrency problem

client/cache/cache_pattern.go
AllenX2018: > I think both reader, writer and lifeSpan are optional.
> 
> For example, users only want `CacheAside` read data from DB if the cache didn't have the specific key. But users update values by ORM directly:
> 
> ```go
> func demo() {
>    aside.NewCacheAside(redisCache, ormReader)
>    aside.Get(ctx, key1)
>    orm.update(key1, xxx)
> }
> ```

I think `CacheAside` encapsulates operation of data source and cache for user. If the user implements the data source operation by himself, I think it loses the meaning of `CacheAside`.

AllenX2018:
> Please resolve the concurrency problem

OK. I will fix and commit after all problems solved.

jianzhiyao:
After discussion with @AllenX2018 , close that pr for now.

PR #4390, opened 2020-12-26
fix-3928

jianzhiyao:
close https://github.com/beego/beego/issues/3928

### DIFF
--- a/server/web/router.go
+++ b/server/web/router.go
@@ -548,7 +548,7 @@ func (p *ControllerRegister) getURL(t *Tree, url, controllerName, methodName str
 	for _, l := range t.leaves {
 		if c, ok := l.runObject.(*ControllerInfo); ok {
 			if c.routerType == routerTypeBeego &&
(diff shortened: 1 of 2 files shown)


server/web/router.go
```diff
@@ -548,7 +548,7 @@ func (p *ControllerRegister) getURL(t *Tree, url, controllerName, methodName str
 	for _, l := range t.leaves {
 		if c, ok := l.runObject.(*ControllerInfo); ok {
 			if c.routerType == routerTypeBeego &&
-				strings.HasSuffix(path.Join(c.controllerType.PkgPath(), c.controllerType.Name()), controllerName) {
+				strings.HasSuffix(path.Join(c.controllerType.PkgPath(), c.controllerType.Name()), `/`+controllerName) {
```
flycash, line 551: You should not user `/` because Windows platform doesn't use `/` as separator

flycash (changes requested):

PR #4418, opened 2021-01-06
Fix issue #4060

liaoyijun:


flycash:
@liaoyijun Hi, please add some unit tests. Beego has terrible code converage, so we are trying our best to improve the coverage

PR #4424, opened 2021-01-08
Add context support

paco0x:
Implement all the missing context support in `client/orm`

### DIFF
--- a/client/orm/cmd.go
+++ b/client/orm/cmd.go
@@ -77,7 +77,6 @@ func RunCommand() {
 
 // sync database struct command interface.
 type commandSyncDb struct {
-	ctx       context.Context
 	al        *alias
 	force     bool
 	verbose   bool
@@ -143,6 +142,7 @@ func (d *commandSyncDb) Run() error {
 		fmt.Printf("    %s\n", err.Error())
 	}
 
+	ctx := context.Background()
 	for i, mi := range modelCache.allOrdered() {
 
 		if !isApplicableTableForDB(mi.addrField, d.al.Name) {
@@ -156,7 +156,7 @@ func (d *commandSyncDb) Run() error {
 			}
 
 			var fields []*fieldInfo
-			columns, err := d.al.DbBaser.GetColumns(d.ctx, db, mi.table)
+			columns, err := d.al.DbBaser.GetColumns(ctx, db, mi.table)
 			if err != nil {
 				if d.rtOnError {
 					return err
@@ -190,7 +190,7 @@ func (d *commandSyncDb) Run() error {
 			}
 
 			for _, idx := range indexes[mi.table] {
-				if !d.al.DbBaser.IndexExists(d.ctx, db, idx.Table, idx.Name) {
+				if !d.al.DbBaser.IndexExists(ctx, db, idx.Table, idx.Name) {
 					if !d.noInfo {
 						fmt.Printf("create index `%s` for table `%s`\n", idx.Name, idx.Table)
 					}
@@ -292,7 +292,6 @@ func RunSyncdb(name string, force bool, verbose bool) error {
 
 	al := getDbAlias(name)
 	cmd := new(commandSyncDb)
-	cmd.ctx = context.TODO()
 	cmd.al = al
 	cmd.force = force
 	cmd.noInfo = !verbose
--- a/client/orm/cmd.go
+++ b/client/orm/cmd.go
@@ -15,6 +15,7 @@
 package orm
 
 import (
+	"context"
 	"flag"
 	"fmt"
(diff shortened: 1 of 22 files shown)


client/orm/cmd.go
```diff
@@ -76,6 +77,7 @@ func RunCommand() {
 
 // sync database struct command interface.
 type commandSyncDb struct {
+	ctx       context.Context
```
flycash, line 80: I think we don't need this `ctx` field. We could pass `context.Backgroud` if we need context.

From my understanding, we can do nothing with context when we run command fron temminal.

client/orm/orm_queryset.go
```diff
@@ -64,22 +64,21 @@ func ColValue(opt operator, value interface{}) interface{} {
 
 // real query struct
 type querySet struct {
```
flycash, line 66: We should not using `context.Context` as field, it's not good practice.

And for this structure, the key point is to change `QuerySeter` interface. There are many methods in this interface but most of them don't need the `context` parameter.

But the "teminal method", which means that actually do query to DB, need to add parameter `ctx context.Context` to its signature.

Those methods are changed in this file now, you could check it.

But in order to keep compatiable, we should think about add new method, for example:
```go
Exist()
ExistWitCtx()
```

client/orm/types.go
```diff
@@ -235,6 +235,8 @@ type Inserter interface {
 
 // QuerySeter query seter
 type QuerySeter interface {
+	// add query context for querySeter
+	WithContext(context.Context) QuerySeter
```
flycash, line 239: As mentioned before, we don't need this method.

flycash (changes requested):
I think using context.Context as strucure's field is not a good practice.

client/orm/cmd.go
paco0x: this can be removable

client/orm/orm_queryset.go
paco0x: In some aspects, I agree with you. But idk what's the plan for your team on this orm module. Your comment seems inconsistent with the existing code. 

You can see there's already `QueryM2MWithCtx` and `QueryTableWithCtx` method in the codebase.  If we remove `ctx` field on `querySet` struct. These two methods and related code also need to be removed. I reserved the `ctx` just for compatibility with the existing code.

I can do a revise as you suggested. But just wanna make sure what the idea of the guy who wrote and merged `QueryM2MWithCtx` and `QueryTableWithCtx`.

client/orm/types.go
paco0x: Same as the above comment.

client/orm/orm_queryset.go
flycash: let me check that.

flycash: I believe this is my mistake. 
I think one way is that keep both two methods and your solution is acceptable. But as I said that, using `ctx` as structure's field is not a good practice.
Another way is that mark both `QueryM2MWithCtx` and `QueryTableWithCtx` as deprecated. In this case, we want you remove the ctx field.

What do you think?

flycash: If we think querySeter is "request-scope", using ctx as structure field looks not bad.

paco0x: To me, I'd prefer the suggestion on the first comment you make. That is to make `context.Context` parameter on the `Insert/Delete/Update` method on `QuerySetter`. I think it's good for any use case cause we don't know how users gonna use it in "request-scope" or something other. 

For the compatibility side. I saw these two `QueryM2MWithCtx ` and `QueryTableWithCtx ` methods are already in the master branch and some tagged versions. I'll push another commit to remove all the `ctx` filed and mark that two methods as deprecated.

paco0x:
done. the sonarcloud code analysis seems to keep failing

client/orm/orm_object.go
```diff
@@ -15,12 +15,14 @@
...
 
 import (
+	"context"
 	"fmt"
 	"reflect"
 )
 
 // an insert queryer struct
 type insertSet struct {
+	ctx    context.Context
```
flycash, line 25: it's similar to QuerySter, and I think we could remove this field and trigger it like other interface.

flycash (changes requested):
Good, and the insertSet and sonar need to be fixed.

flycash:
Oh, you could ignore Sonar problem. It's a little big hard to fix them.

client/orm/orm_object.go
paco0x: done

flycash (approved):

Issue #4411, opened 2021-01-04
Proposal: Reduce memory allocation and gc of function Controller.SaveToFile


AllenX2018:
For now,  we have `Controller.SaveToFile` as following: 
https://github.com/beego/beego/blob/c1b7fa53817c6dbabbf9bcfb61f1a023e36690f4/server/web/controller.go#L606-L621

`io.Copy` called by `SaveToFile` will allocate 32KB buffer every time as following:
```go
// src/io/io.go
func copyBuffer(dst Writer, src Reader, buf []byte) (written int64, err error) {
        // ...
        if buf == nil {
		size := 32 * 1024
		if l, ok := src.(*LimitedReader); ok && int64(size) > l.N {
			if l.N < 1 {
				size = 1
			} else {
				size = int(l.N)
			}
		}
		buf = make([]byte, size)
	}
        // ...
}
```
In high concurrency scenarios, `buf = make([]byte, size)` will allocate memory frequently and increase the gc pressure.

To solve this problem, I think we can add a new function, like: 
```go
type onlyWriter struct {
        io.Writer
}

func (c *Controller) SaveToFileWithBuffer(fromfile, tofile string, buf []byte) error { 
        // ...
        io.CopyBuffer(onlyWriter{f}, file, buf) 
        // ...
}
```
User can pass reused buffer to `buf []byte` of `SaveToFileWithBuffer(fromfile, tofile string, buf []byte)`.  One way to reuse is to use object pool, such as `sync.Pool`, like:
```go
buf := objPool.Get()
io.CopyBuffer(onlyWriter{f}, file, buf.([]byte))
objPool.Put(buf)
```

AllenX2018:
@flycash @jianzhiyao

flycash:
Nice, I believe this is a valuable enhancement. My suggestion is that:
```
SaveToFile(xx) {
    buf := // created by beego and, should we provide ths objPool?
    SaveToFileWithBuffer(xxx, buf)
}
```

AllenX2018:
> Nice, I believe this is a valuable enhancement. My suggestion is that:
> 
> ```
> SaveToFile(xx) {
>     buf := // created by beego and, should we provide ths objPool?
>     SaveToFileWithBuffer(xxx, buf)
> }
> ```

Yes. I think we can use `sync.Pool` as the default buffer management method in `SaveToFile`.
User can use other object pool or other buffer management method through calling `SaveToFileWithBuffer` directly.

flycash:
[...]

PR #4437, opened 2021-01-12
Fix issue #4411

AllenX2018:
Fix issue #4411


flycash (changes requested):
Please add tests and fix goling and sonar problem

flycash (approved):

Issue #4440, opened 2021-01-13
Refactor httplib: Remove debug code to a filter


flycash:
The httplib.BeegoHTTPSettings has field `ShowDebug        bool`
It only be used in method `doRequest`:
```go
	if b.setting.ShowDebug {
		dump, err := httputil.DumpRequest(b.req, b.setting.DumpBody)
		if err != nil {
			log.Println(err.Error())
		}
		b.dump = dump
	}
```
But I don't see any code use `b.dump`. So I think we could remove `b.dump` and this part.

And  then we provide a filter named `LogFilter`. It's default filter and its reponsibility is to log request and response. And if the request.ShowDebug = true, it logs request and response:

```go
log.Debug(request)
next()
log.Debug(response)
```

anoymouscoder:
If you do not hurry, I want to complete this task. Because I am not familiar with the source code of beego project, so I need a few days to get familiar with it. If all goes ok, I will submit a pr next week. Do you think this is ok?

anoymouscoder:
@flycash `DumpRequest()` uses `b.dump`, so I think we can not remove it directly?

https://github.com/beego/beego/blob/develop/client/httplib/httplib.go#L184

anoymouscoder:
I found that there is no place to call `DumpRequest()`.  What scenarios will call this method at present? If not, consider deleting `DumpRequest()` and `b.dump`

flycash:
Thank you @anoymouscoder . This is not an urgent issue, and I assigned it to you. 

> I found that there is no place to call DumpRequest(). What scenarios will call this method at present? If not, consider deleting DumpRequest() and b.dump

I think we could remove both `DumpBody  bool`,  `b.dump`, method `DumpRequest` and `DumpBody` because users are able to call `GetRequest` to get original http.Request and then dump request by themselves.

The scenario is that users want to log request and response when they use httplib. In the future, they could use this `LogFilter` to log request.

anoymouscoder:
@flycash Thank you for your detailed instrutions, I will start now!

PR #4445, opened 2021-01-19
Refactor httplib: Move debug code to a filter #4440

anoymouscoder:
#4440
https://github.com/beego/beego/compare/develop...anoymouscoder:refactor-filter?expand=1#diff-2d3ddf8be721d25ee0d78c8dc8b3a984800822258a894fdef34d2da749676673R638-R645

@flycash I'm not sure this is the result what you want. Becase I think `logFilter()` is meaningless, it prints unreadable golang object. Maybe this is not the debug output that users expect? Looking forward to your suggestions.

### DIFF
--- a/client/httplib/filter/log/filter.go
+++ b/client/httplib/filter/log/filter.go
@@ -1,0 +1,130 @@
+// Copyright 2020 beego
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package log
+
+import (
+	"context"
+	"io"
+	"net/http"
+	"net/http/httputil"
+
+	"github.com/beego/beego/v2/client/httplib"
+	"github.com/beego/beego/v2/core/logs"
+)
+
+// FilterChainBuilder can build a log filter
+type FilterChainBuilder struct {
+	printableContentTypes []string                              // only print the body of included mime types of request and response
+	log                   func(f interface{}, v ...interface{}) // custom log function
+}
+
+// BuilderOption option constructor
+type BuilderOption func(*FilterChainBuilder)
+
+type logInfo struct {
+	req  []byte
+	resp []byte
+	err  error
+}
+
+var defaultprintableContentTypes = []string{
+	"text/plain", "text/xml", "text/html", "text/csv",
+	"text/calendar", "text/javascript", "text/javascript",
+	"text/css",
+}
+
+// NewFilterChainBuilder initialize a filterChainBuilder, pass options to customize
+func NewFilterChainBuilder(opts ...BuilderOption) *FilterChainBuilder {
+	res := &FilterChainBuilder{
+		printableContentTypes: defaultprintableContentTypes,
+		log:                   logs.Debug,
+	}
+	for _, o := range opts {
+		o(res)
+	}
+
+	return res
+}
+
+// WithLog return option constructor modify log function
+func WithLog(f func(f interface{}, v ...interface{})) BuilderOption {
+	return func(h *FilterChainBuilder) {
+		h.log = f
+	}
+}
+
+// WithprintableContentTypes return option constructor modify printableContentTypes
+func WithprintableContentTypes(types []string) BuilderOption {
+	return func(h *FilterChainBuilder) {
+		h.printableContentTypes = types
+	}
+}
+
+// FilterChain can print the request after FilterChain processing and response before processsing
+func (builder *FilterChainBuilder) FilterChain(next httplib.Filter) httplib.Filter {
+	return func(ctx context.Context, req *httplib.BeegoHTTPRequest) (*http.Response, error) {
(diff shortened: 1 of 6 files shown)


client/httplib/httplib.go
```diff
@@ -654,3 +634,12 @@ func TimeoutDialer(cTimeout time.Duration, rwTimeout time.Duration) func(net, ad
...
 }
+
+func logFilter(next Filter) Filter {
+	return func(ctx context.Context, req *BeegoHTTPRequest) (*http.Response, error) {
+		log.Println(req)
+		resp, err := next(ctx, req)
+		log.Println(resp)
+		return resp, err
+	}
+}
```
flycash, line 645: Please move this method to a new file.
We don't print `req` and `resp`, and we could use `httputil.DumpRequest` and `httputil.DumpResponse`.

The keypint is that we should check http header content-type, and print the body if we got expected content-type.

For example, if the content-type is application/json, we print body, or we don't print body.

Let's assume that you have a builder and it has two options:
```go
type logFilterBuilder struct{
    PrintableContentType []string // default value are those textual content-type
    MaxBodyLength  int // default 0
}
```
if we the content-type of request(or response) is in `PrintableContentType` array, we print the body.

When we print the body, if `MaxBodyLength <=0` we print whole body, or we print substring of body.

If you are still confused, and prefer online discussion, we could have a short discussion on zoom.

anoymouscoder: @flycash There are three questiones need your help.
1. What is the relationship between `ShowDebug` and `logFilterBuilder`. As you said, we provide a default filter like Opentracing Filter, user can initialize it by themselves, and add it to `FilterChain`. So `ShowDebug` plays a what kind of  role?
2. User can customize the `PrintableContentType` parameter, do we need to verify whether the parameter really printable? (eg. The default value contains all kinds of printable content types, the user input parameter should be a subset of default.)
3. `logFilter` should print the request processed by `FilterChain` including `doRequest()` and response before filterChain processing?
I am willing to discuss online, when do you have time?

flycash: 1. I prefer that making `logFilter` as default filter which means that we will register this filter when httplib module initialized. So if `ShowDebug` is true, we log request and response, or `logFilter` do nothing. Another option is that we remove `ShowDebug` field, but user should register this filter by themselves.
2. I think users should be responsible for their choices. We only provide default value of `PrintableContentType`, but don't need to verify it when users change it.
3. `logFilter` should log request before `doRequest` and log response after `doReuest`. In our design, `doRequest` is the last "filter", it looks like: `logFilter.next.next...=doRequest`:
```go
func FilterChain(next) Filter{
    return func() {
     logRequest()
     next() // finally we will run doRequest.
     logResponse()
   }
}
```

I always have time except UTC+8 00:00 - 10:00. Please send me an email to tell me your time.

client/httplib/filter/log/filter.go
```diff
@@ -0,0 +1,106 @@
...
+
+	"github.com/beego/beego/v2/client/httplib"
+)
+
+// FilterChainBuilder is used to register logFilter to FilterChain
+type FilterChainBuilder struct {
+	PrintableContentTypes  []string               // only print the included request and response content types
+	MaxPrintableBodyLength int                    // print all content when value <= 0
+	Log                    func(v ...interface{}) // custom log function
+	Out                    io.Writer              // destination for output, only take effect when the Log function is nil
```
flycash, line 33: only keep one filed.
The better way is that you define a NewBuilder function, it looks like
```go
func NewFilterBuilder(opts ...Opt) *FilterChainBuilder {
    res = newDefaultBuilder() // in this method, you will use all default value to build the instance.
    for _, o := range opts {
       o(res)
    }
}
```
Using the `Option` pattern. And then you could make FilterChainBuilder as private(optional).
The default value of `Log` field, it should be `logs.Debug`

client/httplib/filter/log/filter.go
```diff
@@ -0,0 +1,106 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package log
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"log"
```
flycash, line 21: do not use this, please use beego/v2/core/logs

flycash (changes requested):
Good!

client/httplib/filter/log/filter.go
anoymouscoder: Alouthough I followed your suggestion, but want to know why `log` package can not be used.

client/httplib/filter/log/filter.go
anoymouscoder: Considering that other filter builders also use `FilterBuilder`(not private), I think this should be unified.  Using option pattern will indeed be more in line with the Open/Close Principle.

anoymouscoder:
I removed maxPrintableBodyLength because I think maxPrintableContentTyps already controls the printing behavior of the body.  I currently can’t think of a scenario where user only needs to print part of the body.

client/httplib/filter/log/filter.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+}
+
+// FilterChain can print the request after FilterChain processing and response before processsing
+func (builder *FilterChainBuilder) FilterChain(next httplib.Filter) httplib.Filter {
+	return func(ctx context.Context, req *httplib.BeegoHTTPRequest) (*http.Response, error) {
+		info := &logInfo{}
+		defer info.print(builder.log)
+		resp, err := next(ctx, req)
+		if err != nil {
```
flycash, line 80: `info.err = err`  because this error is from user's business, not framework.

flycash (changes requested):

flycash (approved):

PR #4453, opened 2021-01-23
Error Module desgin and apply it to httplib module

flycash:
We are trying to build a error handling mechanism.

From it, users are able to know all errors returned by Beego. For a specific error code, users are able to know how to fix it.

Users could define their own business error code. They only need to invoke `DefineCode` to define codes.

And then they could find codes definition from admin service. We could generate error handbook, users could provide thsi error handbook to their customers. 

When application users find something wrong, they could report error code to application developer.


client/httplib/httplib_test.go
```diff
@@ -345,6 +345,28 @@ func TestNewBeegoRequest(t *testing.T) {
 	req := NewBeegoRequest("http://beego.me", "GET")
 	assert.NotNil(t, req)
 	assert.Equal(t, "GET", req.req.Method)
+
+	req = NewBeegoRequest("httpa\ta://beego.me", "GET")
```
nurlansu, line 349: A typo I guess.

```suggestion
	req = NewBeegoRequest("http://beego.me", "GET")
```

flycash: This is a invalid url test case. 
The NewRequest alway return not nil result and I believe it's not a good behavior. 
image
We should return (*BeegoHTTPRequest, error).

But this is a public API, so I am not sure when to do this change.

nurlansu: > This is a invalid url test case.

Oh, I apologise, I didn't get this right away.

PR #4486, opened 2021-02-09
Remove generate route HTTP hook

flycash:
Now, beego will generate route based on controller's definition. But we already have bee too. So I think moving this function to bee tool is better.

Beego only need to public related functions.


flycash:
https://github.com/beego/bee/pull/762

Issue #4429, opened 2021-01-09
Proposal: Easy cron support for Task module


flycash:
Now we have method:
```
func NewTask(tname string, spec string, f TaskFunc) *Task {
}
```
In fact, users are always trouble with writing exact `spec`, the cron expression. 

We could provide methods:
```go
func Schdule(taskFunc, taskName, time.Duration, ... opts)

// user's code, it means running this task every 30 minutes
Schedule(mytask, "myTask", time.Minute * 30)
```

If user wants the task be scheduled on specific datetime:
```go
func ScheduleOn(taskFunc, taskName, type, opts...) 

// users code, it means that this task run every day on 01:30, 13:30
ScheduleOn(mytask, 'mytask', daily, DailyStartAt(1, 30), DailyStartAt(13, 30)))

// every moday 01:30 
ScheduleOn(mytask, 'mytask', weekly, WeeklyStartAt(MODAY, 1, 30))
```

changsongl:
Do we need to implement all the features of cron expression? I think we just need add some common features. Since we keep the cron expression, they can still use it for some complicated ones.

flycash:
> Do we need to implement all the features of cron expression? I think we just need add some common features. Since we keep the cron expression, they can still use it for some complicated ones.

Yes

PR #4511, opened 2021-02-24
[WIP]Feature easy cron

changsongl:
This is for issue #4429. I finish all APIS for easy crontab. Do you have any suggestion for this implementation? It might be some error in the code, and I will fix it during writing unit tests. @flycash @jianzhiyao

### DIFF
--- a/task/task.go
+++ b/task/task.go
@@ -192,11 +192,8 @@ type Schedule struct {
 // TaskFunc task func type
 type TaskFunc func(ctx context.Context) error
 
-// TaskFunc task func type
-type EasyTaskFunc func(task *Task)
-
-// nextFunc is a function to set next execution time
-type nextFunc func(t *Task, now time.Time)
+// TimeFunc is a function to set next execution time
+type TimeFunc func(t *Task, now time.Time)
 
 // Tasker task interface
 type Tasker interface {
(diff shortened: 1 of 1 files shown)


flycash (commented):
nextFunc and EasyTaskFunc is similar.
One option is that use nextFunc, and remove EasyTaskFunc, and then calculate the time in *StartAt methods;
Another option is no need to introduce these two abstactions. Methods `Repeat` and `*StartAt`
return cron expression;

changsongl:
> nextFunc and EasyTaskFunc is similar.
> One option is that use nextFunc, and remove EasyTaskFunc, and then calculate the time in *StartAt methods;
> Another option is no need to introduce these two abstactions. Methods `Repeat` and `*StartAt`
> return cron expression;

I think I will go with plan 1. I don't think we can transfer repeat method to cron expression. 

For example, if I want to repeat a job for every 11 minutes, the expression could be * */11 * * * *. The fact it will cause a false running time (4:00:00).
````shell
2021-02-28 03:44:00
2021-02-28 03:55:00
2021-02-28 04:00:00
````

changsongl:
I have made the change for the review, let me know if you have more suggestions.

flycash (commented):
Nice job, and please add more unit tests.

flycash:
You could ignore the Sonar problems because I think it's hard to refactor the code to reduce the duplication.

PR #4519, opened 2021-03-02
Add: Helper APIs in Controller

preetham:
Linked Issue: https://github.com/beego/beego/issues/4492

Signed-off-by: Preetham <kamidipreetham@gmail.com>

### DIFF
--- a/server/web/controller.go
+++ b/server/web/controller.go
@@ -722,3 +722,34 @@ func (c *Controller) XSRFFormHTML() string {
 func (c *Controller) GetControllerAndAction() (string, string) {
 	return c.controllerName, c.actionName
 }
+
+// JSONResp sends JSON response directly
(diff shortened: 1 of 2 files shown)


jianzhiyao:
May be you can add some unit test cases.

server/web/controller.go
```diff
@@ -722,3 +722,34 @@ func (c *Controller) XSRFFormHTML() string {
 func (c *Controller) GetControllerAndAction() (string, string) {
 	return c.controllerName, c.actionName
 }
+
+// JSONResp sends JSON response directly
+func (c *Controller) JSONResp(data interface{}, encoding ...bool) error {
```
flycash, line 727: I think we don't need `encoding` params in most cases. So I think if users want to using this param, they could invoke `ServeJSON` manually.

flycash (changes requested):
Please provide unit tests because we want to improve the code quality.

PR #4535, opened 2021-03-15
Fix issue #4528

### LINKED ISSUE #4528: Using reference to loop iterator variable
when i upgrade beego 2.0, something went wrong.

then, i found some code in `github.com/beego/beego/v2/adapter` like this: 

```go
func AddAPPStartHook(hf ...hookfunc) {
	for i := 0; i < len(hf); i++ {
		f := hf[i]
		web.AddAPPStartHook(func() error {
			return f()
		})
	}
}
```
it lead to unintended behavior.

mjyi:
Close #4528


jianzhiyao:
Could you add some unit test cases for #4528 ?

flycash:
> Could you add some unit test cases for #4528 ?

:)

PR #4556, opened 2021-04-01
fix(go.mod): fixed an issue where "go mod tidy" failed to execute

cndoit18:


flycash:
I think it's not beego's problem because I try `go mod tidy` commands on my computers, it works well.
Could you provide more error msg?

cndoit18:
> I think it's not beego's problem because I try `go mod tidy` commands on my computers, it works well.
> Could you provide more error msg?

I tested it again on golang 1.14 and golang 1.16, and both had this problem.
Error message:
go: github.com/beego/beego/v2/core/config/etcd imports
github.com/coreos/etcd/clientv3 tested by
github.com/coreos/etcd/clientv3.test imports
github.com/coreos/etcd/auth imports
github.com/coreos/etcd/mvcc/backend imports
Github.com/coreos/bbolt: github.com/coreos/bbolt@v1.3.5: parsing go. Mod:
module declares its path as: go.etcd.io/bbolt
but was required as: github.com/coreos/bbolt

cndoit18:
> I think it's not beego's problem because I try `go mod tidy` commands on my computers, it works well.
> Could you provide more error msg?

It might have something to do with your go mod cache

flycash:
> > I think it's not beego's problem because I try `go mod tidy` commands on my computers, it works well.
> > Could you provide more error msg?
> 
> It might have something to do with your go mod cache

I recently reinstalled OS and setup golang environment but I didn't get similar problem.

cndoit18:
> > > I think it's not beego's problem because I try `go mod tidy` commands on my computers, it works well.
> > > Could you provide more error msg?
> > 
> > 
> > It might have something to do with your go mod cache
> 
> I recently reinstalled OS and setup golang environment but I didn't get similar problem.

`docker run --rm -v "$PWD":/usr/src/beego -w /usr/src/beego golang:1.15 go mod tidy`
You can verify this by executing this command from beego directory.

flycash:
Thank you, I had reproduce this problem. But your pr could not pass all checks. 
I create a new PR to master branch because I found it has the same problem.
#4561

PR #4569, opened 2021-04-12
Develop luispater

luispater:


flycash:
I am confused about this PR, could you please provide more details about it? What's the problem it want to resolve?

core/config/ini-string.go
```diff
@@ -0,0 +1,450 @@
...
+	"os"
+	"strconv"
+	"strings"
+	"sync"
+
+	"github.com/mitchellh/mapstructure"
+)
+
+// IniStringConfig implements Config to parse ini file.
+type IniStringConfig struct {
```
flycash, line 32: Why do we need this?

server/web/router.go
```diff
@@ -823,6 +818,13 @@ func (p *ControllerRegister) serveHttp(ctx *beecontext.Context) {
 			}
 		}
 	}
+	ctx.Input.ExecMethod = runMethod
+	ctx.Input.ExecController = runRouter
+
+	// execute middleware filters
+	if len(p.filters[BeforeExec]) > 0 && p.execFilter(ctx, urlPath, BeforeExec) {
```
flycash, line 825: you could not move this block here. `BeforeExec` means that we must invoke these filters before `serveHttp`

Issue #4455, opened 2021-01-24
Proposal: Easy way to communitate with multiple HTTP downstream by httplib


flycash:
httplib is a simple module and it works well for a long time.

Recently, when I try to use it to call multiple downstreams, I found it's too simple to support `HTTP-REST` architecture.

Assume that we have different services, and they have different behavior.

They need different headers, different token, and have different SLA.

Here is a simple example:

```go
func BuySomething() error  {

	userId := 123
	userGetReq := Get(fmt.Sprintf("http://api.my-company.com/user/%d", userId)) //
	userGetReq.SetTimeout(100 * time.Millisecond, 100 * time.Millisecond)
	user := &User{}
	err := userGetReq.ToJSON(user)
	if err != nil {
		return err
	}

    userAddressReq := Get(fmt.Sprintf("http://api.my-company.com/user/address/%d", user.PrimaryAddressId)) //
    userAddressReq.SetTimeout(100 * time.Millisecond, 100 * time.Millisecond)
	address := &Address{}

	err = userAddressReq.ToJSON(address)
	if err != nil {
		return nil
	}


	token, err := generateToken()
	if err != nil {
		return err
	}
	createReq := Post("http://order.my-company.com/checkout")

	createReq.Header("Authorization", token)
	createReq.SetTimeout(time.Second, time.Second)
	createReq, err = createReq.JSONBody(&CreateOrderRequest{
		Amount: 10,
		Currency: "USD",
		Buyer: user,
		BuyerAddress: address,
	})

	if err != nil {
		return err
	}

	createResp := &CreateResponse{}
	err = createReq.ToJSON(createResp)
	if err != nil {
		return err
	}

	// handle user response, refund response
	return nil
}
```

In this case, I have to build request, parse response...

This is bad smell code.

So that I can not focus on my business logic. I spent much time on struggling with HTTP protocol

## Client abstraction

We could introduce a new abstraction `Client`:

```go
type Client struct {
	BeegoHTTPSettings
	Name string
	Endpoint string
	CommonOpts []BeegoHttpRequestOption
}

type ClientOption func(client *Client) error

func NewClient(name string, endpoint string, opts...ClientOption) *Client {
	res := &Client{
	}
	return res
}

func WithTimeout(connectTimout time.Duration, rwTimeout time.Duration) ClientOption {
	return func(client *Client) error {
		client.ConnectTimeout = connectTimout
		client.ReadWriteTimeout = rwTimeout
		return nil
	}
}


[...]

PR #4584, opened 2021-04-20
add http client and option func

holooooo:
#4455


client/httplib/httpclient.go
```diff
@@ -0,0 +1,296 @@
...
+)
+
+// Client provides an HTTP client supporting chain call
+type Client struct {
+	Name       string
+	Endpoint   string
+	CommonOpts []BeegoHttpRequestOption
+
+	Setting *BeegoHTTPSettings
+	pointer *responsePointer
```
flycash, line 31: It's not thread-safe.

flycash: There may be many goroutines use the same client. So you could not store `response` inside Client

client/httplib/httpclient.go
```diff
@@ -0,0 +1,296 @@
...
+	if c.pointer.headerValues != nil {
+		for k, v := range c.pointer.headerValues {
+			s := resp.Header.Get(k)
+			*v = &s
+		}
+	}
+	if c.pointer.contentLength != nil {
+		*c.pointer.contentLength = &resp.ContentLength
+	}
+}
```
flycash, line 157: See my above comments. `pointer` is not thread-safe

client/httplib/httpclient.go
```diff
@@ -0,0 +1,296 @@
...
+		}
+	}
+	for _, o := range opts {
+		err := o(req)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return req, nil
+}
```
flycash, line 189: When you are building request, you know the http method.

flycash: So I think we don't need this method. But we could define other method:
```go
func (c *Client) customReq(req *BeegoHTTPRequest, opts []BeegoHttpRequestOption) {
	opts = append(c.commonOpts, opts...)
	for _, o := range opts {
		err := o(req)
		if err != nil {
			return nil, err
		}
	}
}
```

client/httplib/httpclient.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+	if err := req.ToYAML(value); err == nil {
+		return nil
+	}
+	if err := req.ToXML(value); err == nil {
+		return nil
+	}
+
+	// TODO add new error type about can't parse body
+	return berror.Error(UnsupportedBodyType, "unsupported body data")
```
flycash, line 244: Defining new method in struct `BeegoHttpRequest`:
```go
func (req *BeegoHTTPRequest) ResponseForValue(value interface) error {
    contentType := xxxx
    // xxx
}
```

client/httplib/httpclient.go
```diff
@@ -0,0 +1,296 @@
...
+		bs, err := req.Bytes()
+		if err != nil {
+			return nil
+		}
+		*v = &bs
+		return nil
+	}
+
+	// try to parse it as content type
+	switch strings.Split(resp.Header.Get("Content-Type"), ";")[0] {
```
flycash, line 223: Thinking about the case that we don't know the `Content-Type` which means that server didn't return this header.

client/httplib/client_option.go
```diff
@@ -0,0 +1,175 @@
...
+
+import (
+	"crypto/tls"
+	"net/http"
+	"net/url"
+	"time"
+)
+
+type ClientOption func(client *Client) error
+type BeegoHttpRequestOption func(request *BeegoHTTPRequest) error
```
flycash, line 25: I think we don't need to return error.
```go
type ClientOption func(client *Client) error
type BeegoHttpRequestOption func(request *BeegoHTTPRequest) error
```
that would make codes cleaner.

flycash (commented):
Good. And if you don't mind, we could have online discussion about this PR.

holooooo:
I‘m ok with online&nbsp;discussion about this PR.I think point in response() or statuscode() is&nbsp;necessary, maybe we can discuss another way to do it



------------------&nbsp;原始邮件&nbsp;------------------
发件人: "Ming ***@***.***&gt;; 
发送时间: 2021年4月21日(星期三) 晚上9:59
收件人: ***@***.***&gt;; 
抄送: ***@***.***&gt;; ***@***.***&gt;; 
主题: Re: [beego/beego] add http client and option func (#4584)





 
@flycash commented on this pull request.
 
Good. And if you don't mind, we could have online discussion about this PR.

 
In client/httplib/httpclient.go:
 &gt; + +import ( +	"net/http" +	"strings" + +	"github.com/beego/beego/v2/core/berror" +) + +// Client provides an HTTP client supporting chain call +type Client struct { +	Name       string +	Endpoint   string +	CommonOpts []BeegoHttpRequestOption + +	Setting *BeegoHTTPSettings +	pointer *responsePointer  
It's not thread-safe.
 
 
In client/httplib/httpclient.go:
 &gt; + +// Response will set response to the pointer +func (c *Client) Response(resp **http.Response) *Client { +	if c.pointer == nil { +		newC := *c +		newC.pointer = &amp;responsePointer{ +			response: resp, +		} +		return &amp;newC +	} +	c.pointer.response = resp +	return c +} + +// StatusCode will set response StatusCode to the pointer +func (c *Client) StatusCode(code **int) *Client { +	if c.pointer == nil { +		newC := *c +		newC.pointer = &amp;responsePointer{ +			statusCode: code, +		} +		return &amp;newC +	} +	c.pointer.statusCode = code +	return c +} + +// Headers will set response Headers to the pointer +func (c *Client) Headers(headers **http.Header) *Client { +	if c.pointer == nil { +		newC := *c +		newC.pointer = &amp;responsePointer{ +			header: headers, +		} +		return &amp;newC +	} +	c.pointer.header = headers +	return c +} + +// HeaderValue will set response HeaderValue to the pointer +func (c *Client) HeaderValue(key string, value **string) *Client { +	if c.pointer == nil { +		newC := *c +		newC.pointer = &amp;responsePointer{ +			headerValues: map[string]**string{ +				key: value, +			}, +		} +		return &amp;newC +	} +	if c.pointer.headerValues == nil { +		c.pointer.headerValues = map[string]**string{} +	} +	c.pointer.headerValues[key] = value +	return c +} + +// ContentType will set response ContentType to the pointer +func (c *Client) ContentType(contentType **string) *Client { +	return c.HeaderValue("Content-Type", contentType) +} + +// ContentLength will set response ContentLength to the pointer +func (c *Client) ContentLength(contentLength **int64) *Client { +	if c.pointer == nil { +		newC := *c +		newC.pointer = &amp;responsePointer{ +			contentLength: contentLength, +		} +		return &amp;newC +	} +	c.pointer.contentLength = contentLength +	return c +} + +// setPointers set the http response value to pointer +func (c *Client) setPointers(resp *http.Response) { +	if c.pointer == nil { +		return +	} +	if c.pointer.response != nil { +		*c.pointer.response = resp +	} +	if c.pointer.statusCode != nil { +		*c.pointer.statusCode = &amp;resp.StatusCode +	} +	if c.pointer.header != nil { +		*c.pointer.header = &amp;resp.Header +	} +	if c.pointer.headerValues != nil { +		for k, v := range c.pointer.headerValues { +			s := resp.Header.Get(k) +			*v = &amp;s +		} +	} +	if c.pointer.contentLength != nil { +		*c.pointer.contentLength = &amp;resp.ContentLength +	} +}  
See my above comments. pointer is not thread-safe
 
 
In client/httplib/httpclient.go:
 &gt; +func (c *Client) newRequest(method, path string, opts []BeegoHttpRequestOption) (*BeegoHTTPRequest, error) { +	var req *BeegoHTTPRequest +	switch method { +	case http.MethodGet: +		req = Get(c.Endpoint + path) +	case http.MethodPost: +		req = Post(c.Endpoint + path) +	case http.MethodPut: +		req = Put(c.Endpoint + path) +	case http.MethodDelete: +		req = Delete(c.Endpoint + path) +	case http.MethodHead: +		req = Head(c.Endpoint + path) +	} + +	req = req.Setting(*c.Setting) +	for _, o := range c.CommonOpts { +		err := o(req) +		if err != nil { +			return nil, err +		} +	} +	for _, o := range opts { +		err := o(req) +		if err != nil { +			return nil, err +		} +	} +	return req, nil +}  
When you are building request, you know the http method.
 
 
In client/httplib/httpclient.go:
 &gt; +	switch v := value.(type) { +	case **string: +		s, err := req.String() +		if err != nil { +			return nil +		} +		*v = &amp;s +		return nil +	case **[]byte: +		bs, err := req.Bytes() +		if err != nil { +			return nil +		} +		*v = &amp;bs +		return nil +	} + +	// try to parse it as content type +	switch strings.Split(resp.Header.Get("Content-Type"), ";")[0] { +	case "application/json": +		return req.ToJSON(value) +	case "text/xml", "application/xml": +		return req.ToXML(value) +	case "text/yaml", "application/x-yaml": +		return req.ToYAML(value) +	} + +	// try to parse it anyway +	if err := req.ToJSON(value); err == nil { +		return nil +	} +	if err := req.ToYAML(value); err == nil { +		return nil +	} +	if err := req.ToXML(value); err == nil { +		return nil +	} + +	// TODO add new error type about can't parse body +	return berror.Error(UnsupportedBodyType, "unsupported body data")  
Defining new method in struct BeegoHttpRequest:
 func (req *BeegoHTTPRequest) ResponseForValue(value interface) error {     contentType := xxxx     // xxx }
 
 
In client/httplib/httpclient.go:
 &gt; +		if err != nil { +			return nil +		} +		*v = &amp;s +		return nil +	case **[]byte: +		bs, err := req.Bytes() +		if err != nil { +			return nil +		} +		*v = &amp;bs +		return nil +	} + +	// try to parse it as content type +	switch strings.Split(resp.Header.Get("Content-Type"), ";")[0] {  
Thinking about the case that we don't know the Content-Type which means that server didn't return this header.
 
 
In client/httplib/client_option.go:
 &gt; +type ClientOption func(client *Client) error +type BeegoHttpRequestOption func(request *BeegoHTTPRequest) error  
I think we don't need to return error.
 type ClientOption func(client *Client) error type BeegoHttpRequestOption func(request *BeegoHTTPRequest) error
 
that would make codes cleaner.
 
 
In client/httplib/httpclient.go:
 &gt; + +import ( +	"net/http" +	"strings" + +	"github.com/beego/beego/v2/core/berror" +) + +// Client provides an HTTP client supporting chain call +type Client struct { +	Name       string +	Endpoint   string +	CommonOpts []BeegoHttpRequestOption + +	Setting *BeegoHTTPSettings +	pointer *responsePointer  
There may be many goroutines use the same client. So you could not store response inside Client
 
 
In client/httplib/httpclient.go:
 &gt; +func (c *Client) newRequest(method, path string, opts []BeegoHttpRequestOption) (*BeegoHTTPRequest, error) { +	var req *BeegoHTTPRequest +	switch method { +	case http.MethodGet: +		req = Get(c.Endpoint + path) +	case http.MethodPost: +		req = Post(c.Endpoint + path) +	case http.MethodPut: +		req = Put(c.Endpoint + path) +	case http.MethodDelete: +		req = Delete(c.Endpoint + path) +	case http.MethodHead: +		req = Head(c.Endpoint + path) +	} + +	req = req.Setting(*c.Setting) +	for _, o := range c.CommonOpts { +		err := o(req) +		if err != nil { +			return nil, err +		} +	} +	for _, o := range opts { +		err := o(req) +		if err != nil { +			return nil, err +		} +	} +	return req, nil +}  
So I think we don't need this method. But we could define other method:
 func (c *Client) customReq(req *BeegoHTTPRequest, opts []BeegoHttpRequestOption) { 	opts = append(c.commonOpts, opts...) 	for _, o := range opts { 		err := o(req) 		if err != nil { 			return nil, err 		} 	} }

client/httplib/httpclient.go
holooooo: Here I do three scenarios of return. 

One is that when the value is **string or **byte, the content will be directly put into value as string or [] byte and returned. 

The second is to try to judge the type by the value of its header, and parse and return. 

The third is to try to use various types of parsing directly when contenttype does not exist, and then put back the result that the parsing will not report an error.

client/httplib/httpclient.go
holooooo: maybe we can alway return new Client struct point. like

```
func (c *Client) ContentLength(contentLength **int64) *Client {
        newC := *c
	if c.pointer == nil {
               // setting or pointer change to value type
		newC.pointer = responsePointer{
			contentLength: contentLength,
		}
		return &newC
	}
	c.pointer.contentLength = contentLength
	return &newC
}
```

flycash: I think it's not a good idea. From the original design, creating one client instance for one downstream. For example, when users need to communicate with two different severs, they only need to create two client instances. And users could reuse them everywhere. 

Client should be "immutable", which means that users should not change client after creating it. 

So it's easier for users to use client. If we need to create clients every time, why users do not use `BeegoHttpRequest` directly?

client/httplib/httplib.go
```diff
@@ -660,6 +661,64 @@ func (b *BeegoHTTPRequest) Response() (*http.Response, error) {
...
+		*v = &s
+		return nil
+	case **[]byte:
+		bs, err := b.Bytes()
+		if err != nil {
+			return nil
+		}
+		*v = &bs
+		return nil
+	}
```
flycash, line 689: I think, in those two cases, users should read body from original http response.
It means that if users want to know the original response body, response header, response status code, users read from original http response.

client/httplib/httpclient.go
```diff
@@ -0,0 +1,176 @@
...
+)
+
+// Client provides an HTTP client supporting chain call
+type Client struct {
+	Name       string
+	Endpoint   string
+	CommonOpts []BeegoHttpRequestOption
+
+	Setting BeegoHTTPSettings
+	pointer responsePointer
```
flycash, line 28: I think may be we should give up chain invocation if we must keep this field.
I am thinking about using the design proposed by @askuy that we allow users pass pointer like this:
```go
var status int 
var headers http.Header
var httResp http.Response
client.Get(val, path, params, WithStatus(&status), WithHeaders(&headers), WithHttpResponse(httpResp))
```
But I prefer that we could have something like :
```go
type httpClient struct {

}
func (hc *httpClient) Get() HttpResponse, error {
}

type Client struct {
    HTTP *httpClient
}

func (c *Client) Get(value interface, path, opts...) error {
    resp, err := c.Http.Get(path, opts)
    return resp.ResponseForValue(value)
} 
```

flycash: In second way, if users want to handle the response by themselves, they could do this:
```go
client.Http.Get(xxxx)
```
This is similar to your design.

client/httplib/httplib.go
```diff
@@ -660,6 +661,64 @@ func (b *BeegoHTTPRequest) Response() (*http.Response, error) {
 	return b.getResponse()
 }
 
+// ResponseForValue attempts to resolve the response body to value using an existing method.
+// Calls Response inner.
+// If value type is **string or **[]byte, the func directly passes response body into the pointer.
+// Else if response header contain Content-Type, func will call ToJSON\ToXML\ToYAML.
+// Finally it will try to parse body as json\yaml\xml, If all attempts fail, an error will be returned
+func (b *BeegoHTTPRequest) ResponseForValue(value interface{}) error {
```
flycash, line 669: hhhh, may be you could give it a better name :laughing:

client/httplib/httpclient.go
holooooo: got it.Maybe we can merge two of them? For the since when user want easily get struct value and status code(or header).

client/httplib/httplib.go
holooooo: That's embarrassing😆I thought that was the name of the method

client/httplib/httpclient.go
holooooo: Or can we wrapper request and save the point in new request struct？In this way, Client is also immutable, and  chain invocation also can be done. And it also thread safe, cause every request has one and only request struct instance.

```go
// when client first call contenLen or Header, will create a new request
func (c *Client) ContentLength(contentLength **int64) *req{
	return c.newRequest(c, WithContentLength(contentLength ))
}

// new request struct
type req struct {
  client client  // save current client setting
  opts []resOptions
}

// and new req also have same function signature to keep chain go on
func (r *req) ContentLength(contentLength **int64) *req{
        r.opts = append(r.opts, WithContentLength(contentLength ))
	return r
}
```

client/httplib/httpclient.go
flycash: How do we merge them? could you provide a demo? I could not get your point.

holooooo: Sure.This means that users can use both methods at the same time. For examplet, when user use some method like GetUser, I thought he will perfer  `client.Get(val, path, params, WithStatus(&status)`. This allows the user to judge whether the resource does not exist（403） or an error occurs in the acquisition（5xx）.

In some case, user will perfer to use `client.Http.Get(xxxx)`. Like read file stream or raw reponse body. I think pass point with option and handle original response is not mutex. We can make them coexist。

flycash: Good, I think this is the best solution. So in this solution we don't need to store the response inside the `req`?

holooooo: BeegoHttpRequestOption will recive a new param to save response. And pointer intstance will create after req created.
```go
// add new param
type BeegoHttpRequestOption func(request *BeegoHTTPRequest, pointer *responsePointer)

// init response pointer and return to the handleResponse method
func (c *Client) customReq(req *BeegoHTTPRequest, opts []BeegoHttpRequestOption) *responsePointer {
	req.Setting(c.Setting)
	opts = append(c.CommonOpts, opts...)
	rp := &responsePointer{}
	for _, o := range opts {
		o(req, rp)
	}
	return rp
}
```

flycash: How about this:
```go
var resp http.Response
client.Get(xx, xx, xx, WithHttpResponse(&resp))
```
So we don't need to introduce the new structure `responsePointer`.
In your previous design, I think we actually don't need this structure.

holooooo: Is there some way to get `&resp` in `client.Get(xx, xx, xx, WithHttpResponse(&resp))` without any new structure or edit exist structure? :flushed:  I don't know how to solve it

flycash: Ok, I got your concern.

Let's make this problem simple.
How about this:
1. If user want to handle http response manually, it use `client.Http.Get()`
2. If user don't want to handle http response manually, it use `client.Get()`

Another choice is that we return two value:
```go
func (c *Client) Get(...) (http.Response, error)
```

flycash:
@holooooo I think there are three solutions that support `http response`.
**Solution 1**
```go
client.Get(val, path, ...) (*http.Response, error)
```
In this way, we both return `http.Response` and `error`.

**Solution 2**
```go
client.Http.Get(xxx) (http.Response, error)
client.Get(xxx)
````
In this way, user should make the decision using `client.Http.Get` or `client.Get`. 

**Solution 3**
```go
type HttpResponseCarrier interface {
     SetHttpResponse(*http.Response)
}
func (c *Client) Get(val interface{}, ...) {
    handle(val)
    // check val
    if ok := val.(HttpResponseCarrier) {
              val.SetHttpResponse(response)
    }
}
```

holooooo:
@flycash sorry, these days I'm busy with interview.There will be more time to do it next week

Dose solution 3 use like this ?
```
var respCarrier *HttpResponseCarrier
client.get(&respCarrier)
resp := respCarrier.GetResponse()// return a raw response
```
I think this way also make user decision about get body or get struct.

I think Solution 1 is more flexible, and I want to use it. But mostly, user dont need return response, they need to write lot `_, err  = client.Get(val, path, ...)`

What if use 1 and 2 both? 
```
client.Get(val, ...) error
client.Http.Get(val, ...) (http.Response, error)
```
It looks wired, but can make user write less `_` to ignore resp

flycash:
> @flycash sorry, these days I'm busy with interview.There will be more time to do it next week
> 
> Dose solution 3 use like this ?
> 
> ```
> var respCarrier *HttpResponseCarrier
> client.get(&respCarrier)
> resp := respCarrier.GetResponse()// return a raw response
> ```
> 
> I think this way also make user decision about get body or get struct.
> 
> I think Solution 1 is more flexible, and I want to use it. But mostly, user dont need return response, they need to write lot `_, err = client.Get(val, path, ...)`
> 
> What if use 1 and 2 both?
> 
> ```
> client.Get(val, ...) error
> client.Http.Get(val, ...) (http.Response, error)
> ```
> 
> It looks wired, but can make user write less `_` to ignore resp

I agree with you. You could finish this job in this month. I just want to confirm that you still want to help us. :)
And solution1 is a little bit strange. For example, I am the user, and I want to query user, and I want to know the original http response.
```go
// HttpResponseCarrier defined by beego's httplib module
type HttpResponseCarrier interface {
    SetHttpResponse(resp *http.Response)
}

func (c *Client) Get(value interface{}, path, params, opts...) error {
      resp := queryGet() // resp is *http.Response

      // Here is the keypoint, we check whether "value" implement HttpResponseCarrier. If "value" already implement 
     // HttpResponseCarrier , it means that user want to get the original http.Response, so we put the http.Response into "value"
      if carrier, ok := value.(HttpResponseCarrier); ok {
            carrier.SetHttpResponse(resp)
      }

     // 
      return json.Unmarshal(value, resp.Body)
}

// user's code

// UserResponse implement interface HttpResponseCarrier 
type UserResponse struct {
    resp *http.Response
    
    // user data 
    name string
    id int
}

func (u *UserResponse) SetHttpResponse(resp *http.Response) {
    u.resp = resp
}

// and then user invoke http API
func MyBusiness() {
    userResp := &UserResponse{}
    client.Get(userResp, xxx, xxx, xxx)
    httpStatus = userResp.resp.Status
    // do other things.
}
```

holooooo:
I try to implement Solution 3. But in the end, I found the body of response had been closed when user get resp.So I add another interface `ResponseBytesCarrier` for user who want to get bytes of  body.
user need to define a struct like this:
```go
type respCarrier struct {
	Resp  *http.Response
	bytes []byte
}

func (r *respCarrier) SetHttpResponse(resp *http.Response) {
	r.Resp = resp
}

func (r *respCarrier) SetBytes(bytes []byte) {
	r.bytes = bytes
}

func (r *respCarrier) Bytes() []byte {
	return r.bytes
}

func (r *respCarrier) String() string {
	return string(r.bytes)
}

```

flycash:
> I try to implement Solution 3. But in the end, I found the body of response had been closed when user get resp.So I add another interface `ResponseBytesCarrier` for user who want to get bytes of body.
> user need to define a struct like this:
> 
> ```go
> type respCarrier struct {
> 	Resp  *http.Response
> 	bytes []byte
> }
> 
> func (r *respCarrier) SetHttpResponse(resp *http.Response) {
> 	r.Resp = resp
> }
> 
> func (r *respCarrier) SetBytes(bytes []byte) {
> 	r.bytes = bytes
> }
> 
> func (r *respCarrier) Bytes() []byte {
> 	return r.bytes
> }
> 
> func (r *respCarrier) String() string {
> 	return string(r.bytes)
> }
> ```

I think there are tow possible solution, the first solution is:
```go
bodyRes, err := ioutil.ReadAll(resp.Body)  
resbody :=ioutil.NopCloser(bytes.NewReader(bodyRes))
```
Another solution is that, we use `BeegoHttpRequest.Bytes()` method or its field `body`. But in this solution, we should handle `statue`, `header` and `body`:
```
  if (isCarrier) {
      setHeader()
      setStatus()
      setBody()
  }
```

Maybe solution1 is better. You could have a try

holooooo:
@flycash Dose solution1 means replace resp.body with resbody?

flycash:
> @flycash Dose solution1 means replace resp.body with resbody?

Yes, you read data from body, and then you create a new `body` to reset response's body

holooooo:
> Yes, you read data from body, and then you create a new `body` to reset response's body

I think in this way user need to write the same code as `BeegoHTTPRequest.Bytes()` over and over again. I perfer to write more carrier interface for user to chose what he want.

```
type HttpResponseCarrier interface // replace body with new reader
type HttpHeaderCarrier interface // set a map[string]string
type HttpStatusCarrier interface // 
type HttpBodyCarrier interface // set a reader to value
```

flycash:
> > Yes, you read data from body, and then you create a new `body` to reset response's body
> 
> I think in this way user need to write the same code as `BeegoHTTPRequest.Bytes()` over and over again. I perfer to write more carrier interface for user to chose what he want.
> 
> ```
> type HttpResponseCarrier interface // replace body with new reader
> type HttpHeaderCarrier interface // set a map[string]string to value
> type HttpStatusCarrier interface // 
> type HttpBodyCarrier interface // set a reader to value
> type HttpBodyBytesCarrier interface // set a []byte to value
> ```

Good idea!

flycash:
@holooooo thank you, I think this implementation looks great. Could you please check DeepSource checking? So that I could merge this PR after fix them.

holooooo:
one of deepSource issues is about `encoding/xml`.It looks like will be fixed in go 1.17.
Now I've added `xml-roundtrip-validator` as deepSource advise, but it useless.
Should I use another lib to replace `encoding/xml` .

image

flycash:
Ignore deepsource, we plan to ignore some rules. And please help to fix golint issues.

flycash:
image
@holooooo tests failed.

holooooo:
😢 sorry. now it turn to normal

PR #4605, opened 2021-05-12
postFrom

wandercn:
add postFrom with submit formdata

It's the same as this function http.PostForm in net/http.

### DIFF
--- a/client/httplib/httplib.go
+++ b/client/httplib/httplib.go
@@ -38,6 +38,7 @@ import (
 	"crypto/tls"
 	"encoding/json"
 	"encoding/xml"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"mime/multipart"
@@ -56,6 +57,7 @@ import (
 )
 
 const contentTypeKey = "Content-Type"
+
 // it will be the last filter and execute request.Do
 var doRequestFilter = func(ctx context.Context, req *BeegoHTTPRequest) (*http.Response, error) {
 	return req.doRequest(ctx)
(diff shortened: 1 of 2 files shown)


wandercn:
encoding/xml 这个包怎么提示Security 安全风险。我没改这个代码，原来就是使用的这个包 @astaxie

client/httplib/httplib.go
```diff
@@ -275,6 +276,17 @@ func (b *BeegoHTTPRequest) PostFile(formname, filename string) *BeegoHTTPRequest
 	return b
 }
 
+// PostForm submit formdata like "name1=value1&name2=value2" and encode
+func (b *BeegoHTTPRequest) PostForm(fromdata string) (*BeegoHTTPRequest, error) {
```
flycash, line 280: `formdata`?

client/httplib/httplib.go
```diff
@@ -275,6 +276,17 @@ func (b *BeegoHTTPRequest) PostFile(formname, filename string) *BeegoHTTPRequest
...
+	v, err := url.ParseQuery(fromdata)
+	if err != nil {
+		return b, berror.Wrap(err, InvalidFormData, "fromdata format error")
```
flycash, line 283: form, not from

client/httplib/httplib.go
flycash, line 280: I think maybe users want to use `map` as the `formdata`.
If we declare `formdata`, I guess that users need to build the string in most cases. Maybe using `map` would be better.

flycash (changes requested):
please add unit tests.

flycash:
Please fix CI problem

PR #4612, opened 2021-05-17
Deprecated BeeMap and replace all usage with `sync.Map`

CarolineZhang666:


jianzhiyao:
Why are  so many useless files committed?

CarolineZhang666:
> Why are so many useless files committed?

sorry, I should add  (core/config/env/pkg  files ) to gitignore.

jianzhiyao:
> > Why are so many useless files committed?
> 
> sorry, I should add (core/config/env/pkg files ) to gitignore.

Seems those files are still there.

jianzhiyao:
close it for now, and follow it in https://github.com/beego/beego/pull/4616

PR #4616, opened 2021-05-18
Deprecated BeeMap and replace all usage with

CarolineZhang666:


flycash (approved):

jianzhiyao (approved):
LGTM.
We will merge it into develop after change log provided

PR #4620, opened 2021-05-19
Propoal: Convenient way to generate mock object

### LINKED ISSUE #4397: Propoal: Convenient way to generate mock object
Now, we focus on writing unit tests easier. 

We are building mock feature on each module. And we need a solution that user could generate mock object easier.

The proposal is that:
```go
type MockSubObject struct {
	SubA bool `mock:"false"`
}

type MockObject struct {
	A string          `mock:"aaaaa"`
	B int             `mock:"12"`
	C []MockSubObject `mock:"length:1;"`
}
```
Using tht tag and reflection, we could provide the API looks like:
```go
mockObj := bean.Mock(&MockObject{})
```
the mockObj is the same as:
```go
mockObj= &MockObject {
    A: "aaaaa",
    B: 12,
    C: []MockSubObject {
    {
        SubA: false
    }
}
}
```

I hope that we could use this API to generate mock response to help users to build mock server.

iabetor:
see the issue: https://github.com/beego/beego/issues/4397

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # developing
 
+- Add: Convenient way to generate mock object [4620](https://github.com/beego/beego/issues/4397)
 - Lint: use golangci-lint. [4619](https://github.com/beego/beego/pull/4619)
 - Chore: format code. [4615](https://github.com/beego/beego/pull/4615)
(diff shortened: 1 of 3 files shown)


flycash (approved):
Great, thank you! And please fix golint add put this PR into `CHANGELOG.md` and then I will merge it.

PR #4621, opened 2021-05-19
Fix lint according to golangci-lint results

Loyalsoldier:
Use https://github.com/mvdan/gofumpt and run `gofumpt -l -w .` commands, and so on.


jianzhiyao:
please make sure all of tests passed

flycash:
I think maybe you could split this PR into several PR, so you could do this one package by one package.

PR #4635, opened 2021-05-26
task manager graceful shutdown support

optimistic9527:
https://github.com/beego/beego/issues/4631

### DIFF
--- a/task/task.go
+++ b/task/task.go
@@ -593,7 +593,7 @@ func (m *taskManager) StopTask() {
 
 // GracefulShutdown wait all task done
 func (m *taskManager) GracefulShutdown() <-chan struct{} {
-	done := make(chan struct{}, 0)
+	done := make(chan struct{})
 	go func() {
 		m.stop <- true
 		m.wait.Wait()
--- a/task/task.go
+++ b/task/task.go
@@ -37,6 +37,7 @@ type taskManager struct {
 	stop          chan bool
 	changed       chan bool
 	started       bool
+	wait          sync.WaitGroup
 }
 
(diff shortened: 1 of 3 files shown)


task/task.go
```diff
@@ -581,6 +591,17 @@ func (m *taskManager) StopTask() {
 	}()
 }
 
+// GracefulShutdown wait all task done
+func (m *taskManager) GracefulShutdown() <-chan struct{} {
```
flycash, line 595: How about we don't return the `chan`? I think `GracefulShutdown` means that we will block the users and then wait for the final result. So we just return nothing.

flycash (commented):
Please add this PR into `CHANGELOG.md` and fix `golint` and `deepsource` checking. 

Thank you, nice job!

task/task.go
optimistic9527:
```
   shutdown = m.GracefulShutdown()
	select {
	case <-shutdown:
		log.Printf("wait for done")
	case <-time.After(10 * time.Second):
		log.Printf("time out wait job")
	}
```
Worried about the task execution time being too long, the user can customize the timeout exit according to the situation,When all tasks are over, the channel will be closed,Don't worry about leaks.

optimistic9527:
> Please add this PR into `CHANGELOG.md` and fix `golint` and `deepsource` checking.
> 
> Thank you, nice job!

I can’t fix the log DeepSource check

flycash (approved):
It seems that something wrong with Deepsource, let me fix it in the future.
And thank you very much.

PR #4656, opened 2021-06-06
Fix lint and format code in server dir

Loyalsoldier:


Loyalsoldier:
It seems nothing worths changing in SonarCloud and DeepSource.

flycash (approved):

PR #4662, opened 2021-06-11
Chore: update dependencies

Loyalsoldier:


flycash:
We have github action to upgrade dependencies, but thank you!

PR #4668, opened 2021-06-16
Chore: fix code

Loyalsoldier:


Loyalsoldier:
I'm afraid the golangci-lint should not be fixed for backward compatibility.

PR #4679, opened 2021-06-29
add SameSite for Cookie

xjl662750:
add SameSite for Cookie


jianzhiyao:
Could you provide some unit test cases?

flycash (approved):
we should think about improve the design in the future. Checking length is bad pratice.

PR #4680, opened 2021-06-29
Fix orm.SetMaxOpenConns

zav8:
It should call `SetMaxOpenConns` instead of `SetMaxIdleConns`.


jianzhiyao:
Thx for contributions!

jianzhiyao:
PLAT @flycash  ，
SonarCloud Code Analysis failed on Duplication checks.

PR #4701, opened 2021-07-20
fix json marshal in http request

sunxingboo:
The json.Marshal method will escape HTML special characters by default. 
When the value of a field of the request parameter contains these special characters, and the escape is not desired, the BeegoHTTPRequest.JSONBody method does not meet expectations for the processing of the request parameters.

E.g
```golang
body := map[string]interface{} {
	"escape": "left&right",
}
b, _ := json.Marshal(body)
fmt.Println(string(b))  // get '{"escape":"left\u0026right"}' but what is expected is '{"escape":"left&right"}'
```

### DIFF
--- a/client/httplib/httplib_test.go
+++ b/client/httplib/httplib_test.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"context"
 	"errors"
+	"fmt"
 	"io/ioutil"
 	"net"
 	"net/http"
@@ -434,4 +435,14 @@ func TestBeegoHTTPRequestXMLBody(t *testing.T) {
 
 // TODO
 func TestBeegoHTTPRequestResponseForValue(t *testing.T) {
+}
+
+func TestBeegoHTTPRequestMarshal(t *testing.T) {
+	req := Post("http://beego.me")
+	req.SetEscapeHTML(false)
+	body := map[string]interface{} {
+		"escape": "left&right",
+	}
+	b, _ := req.Marshal(body)
+	assert.Equal(t,fmt.Sprintf(`{"escape":"left&right"}%s`, "\n"), string(b))
 }
--- a/client/httplib/httplib_test.go
+++ b/client/httplib/httplib_test.go
@@ -437,12 +437,12 @@ func TestBeegoHTTPRequestXMLBody(t *testing.T) {
 func TestBeegoHTTPRequestResponseForValue(t *testing.T) {
 }
 
-func TestBeegoHTTPRequestMarshal(t *testing.T) {
+func TestBeegoHTTPRequestJSONMarshal(t *testing.T) {
 	req := Post("http://beego.me")
 	req.SetEscapeHTML(false)
 	body := map[string]interface{} {
 		"escape": "left&right",
 	}
-	b, _ := req.Marshal(body)
+	b, _ := req.JSONMarshal(body)
 	assert.Equal(t,fmt.Sprintf(`{"escape":"left&right"}%s`, "\n"), string(b))
 }
--- a/client/httplib/httplib_test.go
+++ b/client/httplib/httplib_test.go
@@ -440,9 +440,9 @@ func TestBeegoHTTPRequestResponseForValue(t *testing.T) {
 func TestBeegoHTTPRequestJSONMarshal(t *testing.T) {
 	req := Post("http://beego.me")
 	req.SetEscapeHTML(false)
-	body := map[string]interface{} {
+	body := map[string]interface{}{
 		"escape": "left&right",
 	}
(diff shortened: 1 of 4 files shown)


jianzhiyao:
What your talk about is your need a method to avoid escaping some character?
As I see, ``{"escape":"left\u0026right"}`` is acceptable.

sunxingboo:
> What your talk about is your need a method to avoid escaping some character?
> 
> As I see, ``{"escape":"left\u0026right"}`` is acceptable.

My usage scenario is that when calling a third-party interface, there is a splicing of ampersands in a field in the request parameter, but the third-party interface does not want to escape the &, but will directly use this parameter value.

The JSON method in the framework does not provide custom options, but is escape by default. I wonder if we should add custom options in this regard.

jianzhiyao:
> > What your talk about is your need a method to avoid escaping some character?
> > As I see, `{"escape":"left\u0026right"}` is acceptable.
> 
> My usage scenario is that when calling a third-party interface, there is a splicing of ampersands in a field in the request parameter, but the third-party interface does not want to escape the &, but will directly use this parameter value.
> 
> The *BeegoHTTPRequest.JSONBody method in the framework does not provide custom options, but is escape by default. I wonder if we should add custom options in this regard.

Alright, Thx for you contribution

sunxingboo:
> > > What your talk about is your need a method to avoid escaping some character?
> 
> > > As I see, `{"escape":"left\u0026right"}` is acceptable.
> 
> > 
> 
> > My usage scenario is that when calling a third-party interface, there is a splicing of ampersands in a field in the request parameter, but the third-party interface does not want to escape the &, but will directly use this parameter value.
> 
> > 
> 
> > The *BeegoHTTPRequest.JSONBody method in the framework does not provide custom options, but is escape by default. I wonder if we should add custom options in this regard.
> 
> 
> 
> Alright, Thx for you contribution

It's my honor. I submit PR for the first time. Will it be adopted?

jianzhiyao (commented):
> adopted

Look good tome

> > > > What your talk about is your need a method to avoid escaping some character?
> > 
> > 
> > > > As I see, `{"escape":"left\u0026right"}` is acceptable.
> > 
> > 
> > > 
> > 
> > 
> > > My usage scenario is that when calling a third-party interface, there is a splicing of ampersands in a field in the request parameter, but the third-party interface does not want to escape the &, but will directly use this parameter value.
> > 
> > 
> > > 
> > 
> > 
> > > The *BeegoHTTPRequest.JSONBody method in the framework does not provide custom options, but is escape by default. I wonder if we should add custom options in this regard.
> > 
> > 
> > Alright, Thx for you contribution
> 
> It's my honor. I submit PR for the first time. Will it be adopted?

It's good to me, but you should rename the function from Marshal to JSONMarshal.

sunxingboo:
> > adopted
> 
> 
> 
> Look good tome
> 
> 
> 
> > > > > What your talk about is your need a method to avoid escaping some character?
> 
> > > 
> 
> > > 
> 
> > > > > As I see, `{"escape":"left\u0026right"}` is acceptable.
> 
> > > 
> 
> > > 
> 
> > > > 
> 
> > > 
> 
> > > 
> 
> > > > My usage scenario is that when calling a third-party interface, there is a splicing of ampersands in a field in the request parameter, but the third-party interface does not want to escape the &, but will directly use this parameter value.
> 
> > > 
> 
> > > 
> 
> > > > 
> 
> > > 
> 
> > > 
> 
> > > > The *BeegoHTTPRequest.JSONBody method in the framework does not provide custom options, but is escape by default. I wonder if we should add custom options in this regard.
> 
> > > 
> 
> > > 
> 
> > > Alright, Thx for you contribution
> 
> > 
> 
> > It's my honor. I submit PR for the first time. Will it be adopted?
> 
> 
> 
> It's good to me, but you should rename the function from Marshal to JSONMarshal.

ok

sunxingboo:
@jianzhiyao 
the function name has been renamed

client/httplib/httplib_test.go
```diff
@@ -435,3 +436,13 @@ func TestBeegoHTTPRequestXMLBody(t *testing.T) {
 // TODO
 func TestBeegoHTTPRequestResponseForValue(t *testing.T) {
 }
+
+func TestBeegoHTTPRequestJSONMarshal(t *testing.T) {
+	req := Post("http://beego.me")
+	req.SetEscapeHTML(false)
+	body := map[string]interface{} {
```
jianzhiyao, line 443: Please format your code

sunxingboo: ok

sunxingboo: done

PR #4713, opened 2021-08-04
PatternLogFormatter supports Line Feed（自定义formatter支持换行符）

dada0z:
PatternLogFormatter supports Line Feed。（自定义formatter支持换行符）

### DIFF
--- a/core/logs/formatter.go
+++ b/core/logs/formatter.go
@@ -62,6 +62,7 @@ func GetFormatter(name string) (LogFormatter, bool) {
 
 // 'w' when, 'm' msg,'f' filename，'F' full path，'n' line number
 // 'l' level number, 't' prefix of level type, 'T' full name of level type
+// 'N' new line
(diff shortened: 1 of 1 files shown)


dada0z:
> Could you show me the scenario that you want this `\n` ?

I want to show full file path in log, but AdapterFile show file name by default. I need to give a custom formatter like this,
`
package main

import (
    "github.com/beego/beego/v2/core/logs"
)

func main() {

    f := &logs.PatternLogFormatter{
        Pattern:    "%F:%n|%w%t>> %m",
        WhenFormat: "2006-01-02",
    }
    logs.RegisterFormatter("pattern", f)

   logs.SetLogger(logs.AdapterFile, `{"filename":"test.log", "formatter": "pattern"}`)

    logs.Info("hello, world")
}
`

But I find no \n at the end of line.

flycash:
I think this is not a common case, why don't you do it by yourself? I mean, you could implement formatter interface by yourself. It's very simple.

PR #4744, opened 2021-08-20
Add sameparty and priority for cookie

zh199225:
Add sameparty and priority for cookie.

### DIFF
--- a/server/web/context/output.go
+++ b/server/web/context/output.go
@@ -87,7 +87,7 @@ func (output *BeegoOutput) Body(content []byte) error {
 }
 
 // Cookie sets a cookie value via given key.
-// others: used to set a cookie's max age time, path,domain, secure and httponly.
+// others: used to set a cookie's max age time, path,domain, secure, httponly, samesite, sameparty and priority.
 func (output *BeegoOutput) Cookie(name string, value string, others ...interface{}) {
 	var b bytes.Buffer
 	fmt.Fprintf(&b, "%s=%s", sanitizeName(name), sanitizeValue(value))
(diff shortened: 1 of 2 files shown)


server/web/context/output.go
```diff
@@ -87,7 +87,7 @@ func (output *BeegoOutput) Body(content []byte) error {
 }
 
 // Cookie sets a cookie value via given key.
-// others: used to set a cookie's max age time, path,domain, secure and httponly.
+// others: used to set a cookie's max age time, path,domain, secure, httponly, samesite, sameparty and priority.
```
flycash, line 90: I  think this is not a good solution. I am thinking about deprecated this method and introduce new method which looks like:
```
type CookieOption func(ck *http.Cookie)
func (output *BeegoOutput) AddCookie(name string, value string, opts ... CookieOption) {}
```

zh199225: I use a function like this in my own code:

```golang
func SetCookie(c *beego.Controller, key string, val string, other map[string]interface{}) {
	var b bytes.Buffer
	fmt.Fprintf(&b, "%s=%s", sanitizeName(key), sanitizeValue(val))
	for each := range other {
		if strings.EqualFold(each, "max-age") {
			var maxAge int64
			switch v := other[each].(type) {
			case int:
				maxAge = int64(v)
                       case int8:
				maxAge = int64(v)
			case int16:
				maxAge = int64(v)
			case int32:
				maxAge = int64(v)
			case int64:
				maxAge = v
			}
			switch {
			case maxAge > 0:
				fmt.Fprintf(&b, "; Expires=%s; Max-Age=%d", time.Now().Add(time.Duration(maxAge)*time.Second).UTC().Format(time.RFC1123), maxAge)
			case maxAge <= 0:
				fmt.Fprintf(&b, "; Max-Age=0")
			}
		} else {
			switch other[each].(type) {
			case string:
				fmt.Fprintf(&b, "; %s=%s", each, other[each])
			case int, int8, int16, int32, int64:
				fmt.Fprintf(&b, "; %s=%d", each, other[each])
			case float32, float64:
				fmt.Fprintf(&b, "; %s=%f", each, other[each])
			case bool:
				if other[each].(bool) {
					fmt.Fprintf(&b, "; %s", each)
				}
			}
		}
	}
	c.Ctx.ResponseWriter.Header().Add("Set-Cookie", b.String())
}
```
I think it is more flexible. For example, if I just want to set "Path" and "Secure", I can use like this:
```golang
SetCookie(&c.Controller, "test", "TestVal", map[string]interface{}{
		"path":      "/test",
		"secure":    true,
	})
```
And I think it can also work well if there are some other properties added to cookie in the future, like this
```golang
SetCookie(&c.Controller, "test", "TestVal", map[string]interface{}{
		"FutureProperty1": "test",
                "FutureProperty2": 300,
                "FutureProperty3": true,
	})
```

flycash: It still not a good practice. I think using `Option` pattern is better.

flycash: Using map has several disadvantages:
1. not type safe.
2. users must know acceptable key.

When we use `Option` pattern, they must pass  correct parameter, and we don't need to do any type conversion.

PR #4759, opened 2021-09-01
fix numeric notation of permissions

nichtsen:
Create a directory without _execute_ permission  is problematic , since we can not access that directory.

For example, the following case 
`logs.SetLogger(logs.AdapterFile, `{"filename":"log/test.log"}`)`

we will get a **permission denied** error

more detail :
[https://en.wikipedia.org/wiki/File-system_permissions#Numeric_notation](wiki)

### DIFF
--- a/core/logs/file.go
+++ b/core/logs/file.go
@@ -136,7 +136,7 @@ func (w *fileLogWriter) Init(config string) error {
 	if len(w.Formatter) > 0 {
 		fmtr, ok := GetFormatter(w.Formatter)
 		if !ok {
-			return errors.New(fmt.Sprintf("the formatter with name: %s not found", w.Formatter))
+			return fmt.Errorf("the formatter with name: %s not found", w.Formatter)
 		}
 		w.formatter = fmtr
 	}
--- a/core/logs/file.go
+++ b/core/logs/file.go
@@ -66,7 +66,7 @@ type fileLogWriter struct {
 	Level int `json:"level"`
 	// Permissions for log file
 	Perm string `json:"perm"`
-	// Permissions for directory if it is spcified in FileName
+	// Permissions for directory if it is specified in FileName
 	DirPerm string `json:"dirperm"`
 
 	RotatePerm string `json:"rotateperm"`
--- a/core/logs/file.go
+++ b/core/logs/file.go
@@ -64,9 +64,9 @@ type fileLogWriter struct {
 	hourlyOpenTime time.Time
 
 	Level int `json:"level"`
-
+	// Permissions for log file
 	Perm string `json:"perm"`
-
+	// Permissions for directory if it is spcified in FileName
 	DirPerm string `json:"dirperm"`
 
 	RotatePerm string `json:"rotateperm"`
--- a/core/logs/file.go
+++ b/core/logs/file.go
@@ -97,7 +97,7 @@ func newFileWriter() Logger {
 	return w
 }
 
-func (w *fileLogWriter) Format(lm *LogMsg) string {
+func (*fileLogWriter) Format(lm *LogMsg) string {
 	msg := lm.OldStyleFormat()
 	hd, _, _ := formatTimeHeader(lm.When)
 	msg = fmt.Sprintf("%s %s\n", string(hd), msg)
--- a/core/logs/file.go
+++ b/core/logs/file.go
@@ -124,7 +124,7 @@ func (w *fileLogWriter) Init(config string) error {
 	if err != nil {
 		return err
 	}
-	if len(w.Filename) == 0 {
+	if w.Filename == "" {
 		return errors.New("jsonconfig must have filename")
 	}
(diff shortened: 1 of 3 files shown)


flycash:
It seems that users could use `Perm` to specific the mode, and it works well. So my question is that, there already is a `Perm`, why we need more `DirPem`?

nichtsen:
> It seems that users could use `Perm` to specific the mode, and it works well. So my question is that, there already is a `Perm`, why we need more `DirPem`?

https://github.com/beego/beego/blob/9f12707deebd483553bea36ee5536d6de359551c/core/logs/file.go#L213

It does not always work well actually, even  a  user explicitly assign the `Perm`, especially  when creating a log file with prefix path,   it  will use PERM both as permmision mode of  the file and its prefix  directory,  in the case of 0660 we do not have permmision to directory, in the case of 0770 the log file will  be grant *execute* permmision which is also not satisfying
Infact , I did not see  any testing method  concern the `filename` with prefix path

flycash:
Sorry for late reply. I got your point. Could you please add more comments to both `Perm` and `DirPerm` so that users could understand the difference. 
Although those Deepsource issues are not introduced by you, please help to fix them, it will be very appriciate.

nichtsen:
i have added more tests for `DirPerm`

core/logs/file.go
```diff
@@ -217,8 +220,13 @@ func (w *fileLogWriter) createLogFile() (*os.File, error) {
...
 	}
 
+	dirperm, err := strconv.ParseInt(w.DirPerm, 8, 64)
+	if err != nil {
+		return nil, err
+	}
+
 	filepath := path.Dir(w.Filename)
-	os.MkdirAll(filepath, os.FileMode(perm))
+	os.MkdirAll(filepath, os.FileMode(dirperm))
```
flycash, line 229: In order to keep compatibility, I think we should check `dirperm` and if it's empty, we use `perm` to create the directories

nichtsen: I think compatibility has been kept 
https://github.com/beego/beego/blob/d01f2dddcc40fa8b134114143297f21f83880374/core/logs/file.go#L433
   when we register  `fileLogWriter` to `Adapter`  we have assigned a default "0770" to dirperm, 
   could you please tell any cases that user can get a  `fileLogWriter` that having a perm value meanwhile dirperm is empty
The only case I can currently think of is an explicitly configuring like:
```go
logs.SetLogger(logs.AdapterFile, `{"filename":"log/test.log", "dirperm": ""}`)
```

flycash: Now the current users may use this function like:
```
logs.SetLogger(logs.AdapterFile, `{"filename":"log/test.log", "perm": 0770}`)
```
and we use this `perm` to create directories.

nichtsen: > Now the current users may use this function like:
> 
> ```
> logs.SetLogger(logs.AdapterFile, `{"filename":"log/test.log", "perm": 0770}`)
> ```
> 
> and we use this `perm` to create directories.

that solution seems  a  compromise  and  sacrifice configurability of log file

flycash: Yes. I think you raised a good question that we may need two `perm` fields. One is used to create directories and another is used to create log files. But now we only have one `perm` and we use it to create both directories and files, so we must avoid to introduce breaking changes.

flycash: In fact, I think we need two `perm` fields according to above discussion. One is for creating prefix directories, which is the `dirperm` in this PR. Another is for creating and writing log files, here is `perm`. 
But, now Beego uses `perm` to create prefix directories and log files, they don't pass `dirperm`. So if we don't do this:
```go
if dirperm == ""{
    dirperm = perm
}
```
Current users may got error because they could not create the prefix directories. 

It looks not good. Or did you have any better idea?

nichtsen: In the current version of this PR
if you run 
```go
logs.SetLogger(logs.AdapterFile, `{"filename":"log/test.log", "perm": 0770}`)
```
you could see that `dirperm` is absent and only `perm` is assigned, but user will not get any error and the prefix directory will be perfectly created with "0770", that beacuse when beego unmarshal user's config, DIrPerm will remain it's  default value .

only in the following case , will user get a error

``` go
logs.SetLogger(logs.AdapterFile, `{"filename":"log/test.log", "dirperm": ""}`)
```

flycash:
Tests failed. Please have a check.

flycash:
it still failed :(

nichtsen:
I found `syscall.Mkdir` in golang will not grant permisson that  current directory ( noted by ".")  do not have,
which is slightly different from shell `mkdir -m <mode>`.  whatever, have fixed it.
P.S
it seems that there still a fail in package `adapter/toolbox` when tests run in `1.16`,  the package itself have been deprecated in documentation, why not remove these test

flycash:
Thank you very much

PR #4785, opened 2021-10-01
minLength and maxLength in renderform parsing

DarkFighterLuke:
I added this two attributes in renderform's input tags parsing. 
They are both very useful when working with forms so that you can enstablish a first limit to the user input.

Tests have been edited to cover the modifications done.


DarkFighterLuke:
No one valued that Pull Request...

PR #4824, opened 2021-11-30
fix unexpected result on 'One' Function in orm

roy19831015:
'One' Funcition was expected to act like returning ErrMultiRows, but if we put limit 1 in sql script, it seems that it will never return ErrMultiRows but return the first one record instead.
It seems a bug.

### DIFF
--- a/client/orm/orm_queryset.go
+++ b/client/orm/orm_queryset.go
@@ -287,7 +287,7 @@ func (o *querySet) One(container interface{}, cols ...string) error {
 }
 
 func (o *querySet) OneWithCtx(ctx context.Context, container interface{}, cols ...string) error {
-	o.limit = 1
+	//o.limit = 1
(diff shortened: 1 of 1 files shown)


flycash:
I think that is the expected behaviour. You can check other ORM frameworks. If they have similar functions, most of them return the first raw. 

The name `One` is not very clear, it should be `First`

roy19831015:
> I think that is the expected behaviour. You can check other ORM frameworks. If they have similar functions, most of them return the first raw.
> 
> The name `One` is not very clear, it should be `First`

Please check the code below:

if num > 1 {
		return ErrMultiRows
	}


'One' is different from 'First',but the 'if' below shows your expectation for the result.
In your opinion, then the 'if' below it should be deleted, because it should 
have never returned an error called ErrMultiRows, it's also unexpected.

anyway, it seems adding a func called 'First' to replace One's func  is the better choice.

It depends on you.

PR #4841, opened 2022-01-05
refactor: SecureCookie: validate timestamp

catroll:
Secure Cookie 中的时间戳还是加个校验吧，要不然就是个摆设


catroll:
我本地的代码好像老了一点点，有点冲突，不好意思，没有先 pull

catroll:
Hello,
Can't read chinese?
I thought it is a chinese project, so ...

Secure Cookie need validate the timestamp, if not, what role does the timestamp should play?

catroll:
Is the project no longer active?

catroll (commented):
merge develop

catroll (commented):
ok

flycash:
This is a break change. And I don't think Beego should do this for two reasons:
- The browser won't carry the cookie to server if it's expired
- Beego provide this function for convenient, so if users want to have this feature, they can design their own "secure cookie"
- In theory, the `timestamp` here was not designed to check if it's expired. It was designed to check the signature.

catroll:
> 

i don't think so
check the timestamp can make system more secure, so it is necessary
the secury can not based on behavior of browser
if we don't want to break anything, maybe we can find a way to avoid it

PR #4862, opened 2022-02-12
Start response automatically when hijacked

de1acr0ix:
Beego router will try to find and render the template automatically after executing the controller, if the response writer (in fact a response object) hasn't been started. 

https://github.com/beego/beego/blob/b3e78c97d7e4eb076d8dc05724e3fee2f73c9b34/server/web/router.go#L1210-L1215

This is reasonable for usual HTTP request because the response writer is started automatically when controller writes any bytes. But if controller hijacks the connection, e. g. WebSocket case, it will use the underlying connection directly and the response writer does not start automatically. When request handling is completed, Beego router tries to render template but there usually aren't be any as hijacking means controller would takeover the connection completely. Eventually a panic happens.

This change starts the request writer automatically when it is hijacked to avoid such issue.

### DIFF
--- a/server/web/context/context.go
+++ b/server/web/context/context.go
@@ -373,9 +373,11 @@ func (r *Response) Hijack() (net.Conn, *bufio.ReadWriter, error) {
 		return nil, nil, errors.New("webserver doesn't support hijacking")
 	}
 	conn, rw, err := hj.Hijack()
-	//if err == nil {
-	//	r.Started = true
-	//}
+	if err == nil {
+		// If the response writer isn't started, router attempts to find a template to render but panics as there isn't
+		// any in most of the cases.
+		r.Started = true
+	}
 	return conn, rw, err
 }
 
--- a/server/web/context/context.go
+++ b/server/web/context/context.go
@@ -373,9 +373,9 @@ func (r *Response) Hijack() (net.Conn, *bufio.ReadWriter, error) {
 		return nil, nil, errors.New("webserver doesn't support hijacking")
 	}
 	conn, rw, err := hj.Hijack()
-	if err == nil {
-		r.Started = true
-	}
+	//if err == nil {
+	//	r.Started = true
+	//}
 	return conn, rw, err
 }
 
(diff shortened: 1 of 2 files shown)


de1acr0ix:
@flycash I am not sure if I am mentioning the right person but I do see you are the primary contributor recently. Can you please help approve the relevant workflows and review this PR? I hope I can get this merged in `master` and then backport it  to 1.x. Thanks!

flycash:
I can not get what you mean, so if you can provide a demo to reproduce the issue you mentioned, that would be very helpful.

de1acr0ix:
> I can not get what you mean, so if you can provide a demo to reproduce the issue you mentioned, that would be very helpful.

Sure. Let me create a minimum example to reproduce the issue.

de1acr0ix:
> I can not get what you mean, so if you can provide a demo to reproduce the issue you mentioned, that would be very helpful.

Sorry I was busy this week. I have just created a very basic unit test showcasing the issue. With the changes in `context.go` the test case would fail.

de1acr0ix:
> I am not sure if we need to do this because I think users can mark the writer started manually. But I will keep this merge request open, so if some one got similar problem, I can learn more about this.
> 
> Thank you!

Sure. Again I am trying to see if I can convince you by explaining my mindset. :-)

The most realistic use case for hijacking is WebSocket. A WebSocket implementation such as Gorilla WebSocket, would hijack the response writer (see https://github.com/gorilla/websocket/blob/69d0eb9187b6dead8fe84b2423518475e5cc535c/server.go#L180). Technically speaking a hijacked response writer does not belong to Beego anymore, because the codes hijacks it is going to read/write the underlying connection directly. Or I would it is already "started". Of course user can still set that field manually. But "user needs to start the response writer manually if he/she uses hijacking" isn't an explicit contract documented anywhere. More importantly I would rather consider that it is more of Beego implementation details as to find out why it panics when I use WebSocket requires some in-depth understanding of Beego router internals. Therefore I created this PR. :-)

PR #4882, opened 2022-03-12
Feature: Add custom TLSConfig 增加一个自定义参数，用于tls配置

dada0z:
CusdomTLSConfig is advanced configuration for tls. it is nil by default and compatible with the current state.
CusdomTLSConfig是一个高阶配置，用于tls配置，一般情况下用不到。它默认是nil，和当前的状态是兼容的。

I want to configure the CipherSuites and MinVersion about tls through this configuration.
我想通过这个配置，来配置tls的加密算法和最小版本。

### DIFF
--- a/server/web/config.go
+++ b/server/web/config.go
@@ -242,6 +242,10 @@ type Listen struct {
 	// The default value is tls.RequireAndVerifyClientCert
 	// @Default 4
 	ClientAuth int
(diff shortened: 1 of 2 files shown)


flycash:
It's a good idea that we need to support fully custom TLS configure. But I think simply adding a new configure item is not elegant. And let me think more about this.
The reason is that users must understand there are two ways to specify the TLS config which is not good for them.

PR #4887, opened 2022-03-18
feat: httplib add WithContext()

h1z3y3:
Support users to give customized context when using builder pattern.
Like: 
```
httplib.Get("your_url").WithContext(your_context).ToJson(&result)
```

But now we can only use like following:

```
req := httplib.Get("your_url")
resp, err := req.DoRequestWithCtx(c.Request().Context())

if err != nil {
    return err
}

defer resp.Body.Close()

// read from resp.Body
body, _ := ioutil.ReadAll(resp.Body)

// and then unmarshal json

json.Unmarshal(body, &result)

```

### DIFF
--- a/client/httplib/httplib.go
+++ b/client/httplib/httplib.go
@@ -87,6 +87,7 @@ func NewBeegoRequest(rawurl, method string) *BeegoHTTPRequest {
 		files:   map[string]string{},
 		setting: defaultSetting,
 		resp:    &resp,
+		context: context.Background(),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


h1z3y3:
@flycash  Could you please help me to review this PR?

flycash (commented):
It seems that if users want to use context as input, they can use `DoRequestWithCtx`.

h1z3y3:
Right. But DoRequestWithCtx returns Response and error. We can't use like this: httplib. DoRequestWithCtx.ToJSON()

h1z3y3:
> I check the source code and found that Beego actually does not support the context parameter even if there is a new function named DoRequestWithCtx. So I think we should design a new function:
> 
> ```go
> NewBeegoRequestWithCtx(ctx context.Context, rawurl, method string) *BeegoHttpRequest {
>    // inside we call http.NewRequestWithCtx
> }
> ```
> 
> In this way, we can mark `DoRequestWithCtx` as deprecated.

The truth is that httplib doesn't use the Context but the filters in the FilterChain does.

I send a new pull request to support you mentioned. 
Please review https://github.com/beego/beego/pull/4895

PR #4895, opened 2022-03-30
feat: add NewBeegoRequestWithCtx

h1z3y3:
More information: https://github.com/beego/beego/pull/4887#issuecomment-1080781354

### DIFF
--- a/client/httplib/httplib.go
+++ b/client/httplib/httplib.go
@@ -67,26 +67,23 @@ var doRequestFilter = func(ctx context.Context, req *BeegoHTTPRequest) (*http.Re
 // I think if we don't return error
 // users are hard to check whether we create Beego request successfully
 func NewBeegoRequest(rawurl, method string) *BeegoHTTPRequest {
-	var resp http.Response
-	u, err := url.Parse(rawurl)
+	return NewBeegoRequestWithCtx(context.Background(), rawurl, method)
+}
+
+// NewBeegoRequestWithCtx returns a new BeegoHTTPRequest given a method, URL
+func NewBeegoRequestWithCtx(ctx context.Context, rawurl, method string) *BeegoHTTPRequest {
+	req, err := http.NewRequestWithContext(ctx, method, rawurl, nil)
 	if err != nil {
-		logs.Error("%+v", berror.Wrapf(err, InvalidUrl, "invalid raw url: %s", rawurl))
+		logs.Error("%+v", berror.Wrapf(err, InvalidURLOrMethod, "invalid raw url or method: %s %s", rawurl, method))
 	}
-	req := http.Request{
-		URL:        u,
-		Method:     method,
-		Header:     make(http.Header),
-		Proto:      "HTTP/1.1",
-		ProtoMajor: 1,
-		ProtoMinor: 1,
-	}
+
 	return &BeegoHTTPRequest{
 		url:     rawurl,
-		req:     &req,
+		req:     req,
 		params:  map[string][]string{},
 		files:   map[string]string{},
 		setting: defaultSetting,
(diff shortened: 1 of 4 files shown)


client/httplib/httplib.go
```diff
@@ -58,36 +58,44 @@ import (
...
+	req, err := NewBeegoRequestWithContext(context.Background(), rawurl, method)
+	if err != nil {
+		return &BeegoHTTPRequest{}
+	}
+
+	return req
+}
+
+// NewBeegoRequestWithContext returns a new BeegoHTTPRequest given a method, URL
+func NewBeegoRequestWithContext(ctx context.Context, rawurl, method string) (*BeegoHTTPRequest, error) {
```
flycash, line 79: NewBeegoRequestWithCtx

client/httplib/httplib.go
```diff
@@ -409,7 +422,7 @@ func (b *BeegoHTTPRequest) handleFiles() {
 	b.Header("Transfer-Encoding", "chunked")
 }
 
-func (b *BeegoHTTPRequest) handleFileToBody(bodyWriter *multipart.Writer, formname string, filename string) {
+func (_ *BeegoHTTPRequest) handleFileToBody(bodyWriter *multipart.Writer, formname string, filename string) {
 	fileWriter, err := bodyWriter.CreateFormFile(formname, filename)
```
flycash, line 426: func (*BeegoHTTPRequest)

client/httplib/httplib.go
flycash: Because the BeegoRequest is the implementation of Builder pattern. So this function we should not return error. 
But I do agree with you that here we should tell users what happen.
So I think the possible solution is we add a new field named "err" in structure.
And when we call the terminative function like "DoRequest", we check the `if b.err != nil return b.err`

client/httplib/httplib.go
```diff
@@ -131,6 +139,11 @@ func (b *BeegoHTTPRequest) GetRequest() *http.Request {
 	return b.req
 }
 
+// Context returns the request context
+func (b *BeegoHTTPRequest) Context() context.Context {
```
flycash, line 143: I think we don't need this function. I can not image that why users want to get the context.

client/httplib/httplib.go
```diff
@@ -58,36 +58,44 @@ import (
 const contentTypeKey = "Content-Type"
 
 // it will be the last filter and execute request.Do
-var doRequestFilter = func(ctx context.Context, req *BeegoHTTPRequest) (*http.Response, error) {
-	return req.doRequest(ctx)
+var doRequestFilter = func(_ context.Context, req *BeegoHTTPRequest) (*http.Response, error) {
```
flycash, line 61: don't need to change this.

flycash (commented):
you still get some deepsource issues, please help to fix them.

h1z3y3:
Force push to fix: https://github.com/beego/beego/actions/runs/2065517230

Do we need to fix this lint problem: https://deepsource.io/gh/beego/beego/run/bda966d9-e29d-41ce-90d5-d93cb248db25/go/CRT-P0003

flycash:
> Force push to fix: https://github.com/beego/beego/actions/runs/2065517230
> 
> Do we need to fix this deepsource problem? I didn't change the function. Shall we open a new pull request to fix it? https://deepsource.io/gh/beego/beego/run/bda966d9-e29d-41ce-90d5-d93cb248db25/go/CRT-P0003

ignore that, it's a tiny break change

flycash (approved):

h1z3y3:
Finally！
Thanks for your patience and kindness during the review.

PR #4896, opened 2022-03-31
upgrade Go version to 1.16

flycash:


flycash:
For these reasons:
- Almost every new Go version will get some performance improvement and add more useful API
- Many third party require higher Go version

And why I don't choose the newest version:
- For users, they may not want to take the risk using newest version

So I choose Go1.16. I know the Go1.18 is released, but I think we don't need to use Go 1.17 or Go 1.18.

flycash:
For these reasons:
- Go1.18 has generic and we plan to use the generic to rewrite or provide mode functions
- The newest Go version always has some performance improvement
- Third party dependencies require higher version than Go 1.14

PR #4914, opened 2022-04-17
fix 4911: make the argument work

### LINKED ISSUE #4911: PatternLogFormatter not handled args
// ToString 'w' when, 'm' msg,'f' filename，'F' full path，'n' line number
// 'l' level number, 't' prefix of level type, 'T' full name of level type
func (p *PatternLogFormatter) ToString(lm *LogMsg) string {
	s := []rune(p.Pattern)
	m := map[rune]string{
		'w': lm.When.Format(p.getWhenFormatter()),
		'm': lm.Msg,
		'n': strconv.Itoa(lm.LineNumber),
		'l': strconv.Itoa(lm.Level),
		't': levelPrefix[lm.Level],
		'T': levelNames[lm.Level],
		'F': lm.FilePath,
	}
	_, m['f'] = path.Split(lm.FilePath)
	res := ""
	for i := 0; i < len(s)-1; i++ {
		if s[i] == '%' {
			if k, ok := m[s[i+1]]; ok {
				res += k
				i++
				continue
			}
		}
		res += string(s[i])
	}
	return res
}
lm.Msg no use lm.Args

flycash:
Closes #4911

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 - [upgrade redisgo to v1.8.8](https://github.com/beego/beego/pull/4872)
 - [fix prometheus CVE-2022-21698](https://github.com/beego/beego/pull/4878)
 - [upgrade to Go 1.18](https://github.com/beego/beego/pull/4896)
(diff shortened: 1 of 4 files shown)


flycash:
Caused by #4804

PR #4915, opened 2022-04-18
fix: inject propagator header

h1z3y3:
Forgot to inject propagator to request header. ||- - 

fix: https://github.com/beego/beego/pull/4888


h1z3y3:
@flycash  Please review and merge. 
Inject the request header to the upstream.

PR #4918, opened 2022-04-21
support LifeCycleCallback

robberphex:
* Run graceful mode by default.
  Go's `net.http.Server` cannot process in-transit requests when the server is shuting down, so graceful is set to the default Server implementation. `net.http.Server` is used only if the user manually turn off (`graceful=false`).
* Support `GetAllControllerInfo`. `GetAllControllerInfo` can be used to debug and manage routing information conveniently in debugging and other extension mechanisms.
* Support `LifeCycleCallback`. Adding 'AfterStart' and 'BeforeShutdown' extension points in the Beego framework lifecycle to provide more extensibility.

---

* **默认使用graceful模式**。Go的[net.http.Server](https://pkg.go.dev/net/http#Server) 在下线流量时会导致无法处理在途请求，所以将graceful设置为默认的Server实现。如果用户手动关闭的话（`graceful=false`），才使用`net.http.Server`。
* **支持路由信息内省**（`GetAllControllerInfo`）。在调试或者其他扩展机制中，路由信息内省可以很方便的调试、管理路由信息。
  比如在应用启动时，列出所有的route信息，有助于业务开发者查看路由是否生效。 
* **支持生命周期回调**。在beego框架生命周期中，增加`AfterStart`和`BeforeShutdown`扩展点，能够提供更多的扩展性。
  比如在应用就绪后，将应用注册到consul等注册中心。


jianzhiyao:
DeepSource test cases can not pass. Pls take a look!
@robberphex

flycash:
why we need this feature? can you provide some scenarios?

server/web/config.go
```diff
@@ -535,7 +535,7 @@ func newBConfig() *Config {
 		EnableErrorsShow:   true,
 		EnableErrorsRender: true,
 		Listen: Listen{
-			Graceful:      false,
+			Graceful:      true,
```
flycash, line 538: This should keep false because it's a break change. Users need to specify using true manually

server/web/router_test.go
```diff
@@ -1097,3 +1098,22 @@ func TestRouterAddRouterPointerMethodPanicNotImplementInterface(t *testing.T) {
 	handler := NewControllerRegister()
 	handler.AddRouterMethod(method, "/user", (*TestControllerWithInterface).PingPointer)
 }
+
+func TestGetAllControllerInfo(t *testing.T) {
```
flycash, line 1102: you still need to add more test cases to cover `composeControllerInfos` all path

server/web/server.go
```diff
@@ -76,6 +77,22 @@ func NewHttpServerWithCfg(cfg *Config) *HttpServer {
...
 type MiddleWare func(http.Handler) http.Handler
 
+// LifeCycleCallback configures callback.
+type LifeCycleCallback interface {
+	AfterStart(app *HttpServer)
+	BeforeShutdown(app *HttpServer)
+	mustEmbedUnimplementedLifeCycleCallback()
+}
+
+type UnimplementedLifeCycleCallback struct {
```
flycash, line 87: I think we don't need this because if we add more function to `LifeCycleCallback` interface, it's better for users to know this. They can easily get compiler error.
And it's not possible for users to have many callbacks, so it's easy for them to change it.
This is different from gRPC because gRPC is a RPC framework so it must keep both client side and server side compatible. But Beego doen't need this.

server/web/server.go
```diff
@@ -182,7 +199,22 @@ func (app *HttpServer) Run(addr string, mws ...MiddleWare) {
 				if app.Cfg.Listen.ListenTCP4 {
 					server.Network = "tcp4"
 				}
-				if err := server.ListenAndServe(); err != nil {
+				ln, err := net.Listen(server.Network, app.Server.Addr)
```
flycash, line 202: if you want to support the callbacks, I think you need to think about `app.Cfg.Listen.EnableFcgi` and `app.Cfg.Listen.EnableHTTPS || app.Cfg.Listen.EnableMutualHTTPS`, all cases.

server/web/grace/grace.go
```diff
@@ -107,6 +107,10 @@ func init() {
```
jianzhiyao, line 107: Pls use Options mode rathan making anthor method
like
```golang
func NewServer(addr string, handler http.Handler,opts ... ServerOpt) (srv *Server) {
```

jianzhiyao, line 107: Pls use Options mode rathan making anthor method
like
```golang
func NewServer(addr string, handler http.Handler,opts ... ServerOpt) (srv *Server) {
```

server/web/server.go
```diff
@@ -76,6 +77,12 @@ func NewHttpServerWithCfg(cfg *Config) *HttpServer {
 // MiddleWare function for http.Handler
 type MiddleWare func(http.Handler) http.Handler
```
jianzhiyao, line 79: Why there is no any implement of LifeCycleCallback ?

server/web/grace/grace.go
robberphex: `func NewServer(addr string, handler http.Handler,opts ... ServerOpt)` is adopted.

server/web/server.go
robberphex: `LifeCycleCallback` will be used for other extension or plugin, to custom logic into beego server lifecycle.

jianzhiyao:
lgtm

xuing:
这个合并，导致我无法开启graceful模式，开启后执行到ServeWithListener时，就不继续向下执行了..请问大家居然都没有遇到类似问题吗？

PR #4933, opened 2022-05-07
[bugfix] fix orm generate sql err on postgres

shanghs:
postgres comment syntax like this
```sql
-- --------------------------------------------------
--  Table Structure for `train/internal/model.Test`
-- --------------------------------------------------
CREATE TABLE IF NOT EXISTS "test" (
    "id" serial NOT NULL PRIMARY KEY
);
COMMENT ON COLUMN test.id is 'record id';
```
so make a string array to cache field comments and append to the end of raw sql


flycash:
It has been fixed, but thank you!

PR #4940, opened 2022-05-14
Fix: Append column comments to `create table` sentence when using postgres

### LINKED ISSUE #4932: [ORM BUG] create table failed when datasource is postgres
**go version go1.17.3 darwin/arm64
bee version v2.0.2**

**GOOS="darwin"
GOARCH="arm64"**

3. What did you do?
**create table with orm on postgres failed, when there has comment on orm tags**

If this is ORM issue, please provide the DB schemas.
just a sample struct like this 
```golang
type Test struct {
	Id int64 `orm:"auto;description(record id)"`
}

func init() {
	orm.RegisterModel(new(Test))
}
```

create table success, and we can find it in our database

5. What did you see instead?
when we startup our project, use orm.RunSyncdb to create table, there will have an error occur, err is `pq: syntax error at or near "COMMENT"`

this issue is occur only PostgreSQL. In https://github.com/beego/beego/blob/develop/client/orm/models.go line 478
```golang
if fi.description != "" && al.Driver != DRSqlite {
	column += " " + fmt.Sprintf("COMMENT '%s'", fi.description)
}
```
so the generate SQL will like this
```sql 
-- --------------------------------------------------
--  Table Structure for `train/internal/model.Test`
-- --------------------------------------------------
CREATE TABLE IF NOT EXISTS "test" (
    "id" serial NOT NULL PRIMARY KEY COMMENT 'record id'
);
```
but in postgres the correct sql like this
```sql 
-- --------------------------------------------------
--  Table Structure for `train/internal/model.Test`
-- --------------------------------------------------
CREATE TABLE IF NOT EXISTS "test" (
    "id" serial NOT NULL PRIMARY KEY
);
COMMENT ON COLUMN test.id is 'record id';
```

greenhandatsjtu:
Fix: #4932
For now, creating table with column comments when using Postgres will fails with error like:
```bash
create table `user` 
    -- --------------------------------------------------
    --  Table Structure for `main.User`
    -- --------------------------------------------------
    CREATE TABLE IF NOT EXISTS "user" (
        "id" serial NOT NULL PRIMARY KEY COMMENT 'user id',
        "name" text NOT NULL DEFAULT ''  COMMENT 'user name',
        "profile_id" integer NOT NULL UNIQUE
    );
panic: pq: syntax error at or near "COMMENT"
```
It's because, in Postgres, column comments should be appended to `create table` sentence.

Below is how I fix this issue:
When using Postgres, store indexes of columns which have comments, and append comments to `create table` sentence.
Now it works well:
```bash
create table `user` 
    -- --------------------------------------------------
    --  Table Structure for `main.User`
    -- --------------------------------------------------
    CREATE TABLE IF NOT EXISTS "user" (
        "id" serial NOT NULL PRIMARY KEY,
        "name" text NOT NULL DEFAULT '' ,
        "profile_id" integer NOT NULL UNIQUE
    );
    COMMENT ON COLUMN "user"."id" is 'user id';
    COMMENT ON COLUMN "user"."name" is 'user name';
```


flycash:
You can add a new file named "model_test".
I check the code, so I think you can test the private function `getDbCreateSQL` directly

flycash:
You can learn more about the go test 😃

greenhandatsjtu:
Hi, I've add test for `getDbCreateSQL()`, but I have no idea why the test CI fails at step `Run tests on mysql` in `github.com/beego/beego/v2/adapter/toolbox` and `github.com/beego/beego/v2/core/logs`:
https://github.com/greenhandatsjtu/beego/runs/6502090588?check_suite_focus=true
image

image
I didn't modify code in these pacakges.

client/orm/model_test.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+type ModelWithComments struct {
+	Id       int    `orm:"description(user id)"`
+	UserName string `orm:"size(30);unique;description(user name)"`
+	Email    string `orm:"size(100)"`
+	Password string `orm:"size(100)"`
+}
+
+func TestGetDbCreateSQLWithComment(t *testing.T) {
```
flycash, line 30: nice, but you can improve the unit tests.
Firstly, use table driven to write the unit tests, for example
```go
testCases := []struct{
    name string
    wantSQL string
    wantErr error
}
for _, tc := range testCases {
   t.Run(tc.name, func(t) {
    // ...
})
}
```
Secondly, do not use `modelCache`. The unit tests should be independent, but if you use the `modelCahe`, other tests may be affected by your test. So you can create a new `_modelCache`.

Thirdly, you need to think about the scenarios, for example, no comment, empty comment...

Finally, write the expected SQL directly, it's not necessary to use symbol "+" to construct the expected SQL

flycash (changes requested):

client/orm/model_test.go
greenhandatsjtu: Thanks for your advice! I'll try to fix these issues.

flycash (approved):

PR #4951, opened 2022-05-20
Update tree.go

runner361:
fix issues "https://github.com/beego/beego/issues/4946"
when `pattern == "" ` , it means ` seg ` is the last element，so we can treat aa.xml as aa , and set `:ext` as `xml`.


flycash:
Please add unit tests

PR #4968, opened 2022-05-30
Fix: Graceful mode can't running in http

xuing:
We don't need ServeWithListener to start server. use `ListenAndServe` is better choice, they do same things. and it can fix issue that http server be blocked in graceful mode.

### DIFF
--- a/server/web/server.go
+++ b/server/web/server.go
@@ -215,17 +215,11 @@ func (app *HttpServer) Run(addr string, mws ...MiddleWare) {
 				if app.Cfg.Listen.ListenTCP4 {
 					server.Network = "tcp4"
 				}
-				ln, err := net.Listen(server.Network, app.Server.Addr)
-				if err != nil {
-					logs.Critical("Listen for HTTP[graceful mode]: ", err)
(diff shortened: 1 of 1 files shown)


xuing:
I only fix it on http mode, but there are similar issue in https mode. perhaps should revert `ListenTLS` to  `ListenAndServeTLS` function.

flycash:
> We don't need ServeWithListener to start server. use `ListenAndServe` is better choice, they do same things. and it can fix issue that http server be blocked in graceful mode.

I am sorry, what do you mean "http server be blocked in graceful mode"? can you provide a demo?

xuing:
> > We don't need ServeWithListener to start server. use `ListenAndServe` is better choice, they do same things. and it can fix issue that http server be blocked in graceful mode.
> 
> I am sorry, what do you mean "http server be blocked in graceful mode"? can you provide a demo?

https://github.com/xuing/beego_demo01

xuing:
image
then,I turn off graceful
image

robberphex:
`AfterStart` should be called before `serve`, after `listen`.

PR #4978, opened 2022-06-09
fix: cache random time offset expired.

auula:
1. fix: [https://github.com/beego/beego/issues/4966](https://github.com/beego/beego/issues/4966)
2. add a random offset to the expire time.


auula:
哥这个，能不能合并啊，检测工具检测都是lint错误，我代码应该没啥问题啊

PR #4984, opened 2022-06-12
fix: random expired time cache.

auula:
1. fix [https://github.com/beego/beego/issues/4966](https://github.com/beego/beego/issues/4966)
2. add random offset, range from 3 to 8 seconds.


flycash (approved):

auula:
LGTM！！！！！！

PR #4990, opened 2022-06-18
bugfix: Csrf token should be Secure and httpOnly, but not now

dada0z:
Issue: https://github.com/beego/beego/issues/4989


flycash:
Oh!

PR #4999, opened 2022-06-27
Feat: add get all tasks function

qwganker:


task/task.go
```diff
@@ -653,6 +654,22 @@ func (m *taskManager) ClearTask() {
 	}
 }
 
+// GetAllTasks get all tasks
+func (m *taskManager) GetAllTasks() []*Task {
+	m.taskLock.RLock()
+
+	var l []*Task
+
+	for _, tasker := range m.adminTaskList {
```
flycash, line 663: why do you use reflection here?

task/task_test.go
```diff
@@ -206,3 +206,28 @@ func wait(wg *sync.WaitGroup) chan bool {
...
+	tk2 := NewTask("task2", "0/40 * * * * *", func(ctx context.Context) error {
+		return nil
+	})
+
+	m.AddTask("task1", tk)
+	m.AddTask("task2", tk2)
+
+	l := m.GetAllTasks()
+	taskNum := 0
+	for _, t := range l {
```
flycash, line 227: use assertion in tests, and do not use any printing function, including `fmt.Println`

flycash (changes requested):

qwganker:
@flycash 

pls, review again,

qwganker:
> I mean, why you use reflection in your code? you want to make a copy? but why we need a copy?

ok, now, i use another way to implment it that add GetTask() function into Tasker{} interface

qwganker:
`DeepSource`  scan rule is changed?

task/task.go
```diff
@@ -103,6 +103,7 @@ type TaskFunc func(ctx context.Context) error
 
 // Tasker task interface
 type Tasker interface {
+	GetTask(ctx context.Context) *Task
```
flycash, line 106: don't need this function

task/task.go
```diff
@@ -653,6 +659,20 @@ func (m *taskManager) ClearTask() {
 	}
 }
 
+// GetAllTasks get all tasks
+func (m *taskManager) GetAllTasks() []*Task {
```
flycash, line 663: it should return the Tasker, and the users do the type assertion by themselves.
And the `taskManager` is a private structure, how do users call this function?

qwganker:
@flycash  I updated the implement,  GetAllTasks() function will export to public, user  can use it like this
```golang
import beego "github.com/beego/beego/v2/task"

...

beego.AddTask("test1", t1)
beego.AddTask("test2", t2)

...

tasks := beego.GetAllTasks()
for _, task := range tasks {
  rv := reflect.ValueOf(task)
  t := rv.Interface().(*Task)
  fmt.Println(t)
}



```

task/task.go
```diff
@@ -653,6 +658,20 @@ func (m *taskManager) ClearTask() {
 	}
 }
 
+// GetAllTasks get all tasks
+func (m *taskManager) GetAllTasks() []Tasker {
+	m.taskLock.RLock()
+
+	var l []Tasker
```
flycash, line 665:
```
l := make([]Tasker, 0, len(m.adminTaskList))
```
We can specify the capacity to avoid the memory allocation

flycash (changes requested):

flycash:
And do remember add one change log into `.CHANGELOG` file like this:
image

qwganker:
@flycash  I has been updated, pls review again

flycash (approved):

PR #5001, opened 2022-06-28
fix: nil offset function of RandomExpireCache instance

ktalg:
1. when creating an RandomExpireCache instance, use the default random time generation function and allow external definitions
2. avoid using global rand instances to improve concurrency performance

### DIFF
--- a/client/cache/random_expired_cache.go
+++ b/client/cache/random_expired_cache.go
@@ -17,6 +17,7 @@ package cache
 import (
 	"context"
 	"math/rand"
+	"sync/atomic"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


client/cache/random_expired_cache.go
```diff
@@ -27,12 +28,12 @@ type RandomExpireCacheOption func(*RandomExpireCache)
 // Cache random time offset expired
 type RandomExpireCache struct {
 	Cache
-	offset func() time.Duration
+	Offset func() time.Duration
```
flycash, line 31: keep it private and add one more function:
```go
func WithOffsetFunc(fn func() time.Duration) RandomExpireCacheOption {
}
```

client/cache/random_expired_cache.go
```diff
@@ -43,10 +44,26 @@ func NewRandomExpireCache(adapter Cache, opts ...RandomExpireCacheOption) Cache
 	for _, fn := range opts {
 		fn(&rec)
 	}
+	if rec.Offset == nil {
```
flycash, line 47:
```go
	rec := RandomExpireCache {
            Cache: adapter,
            offset: defaultOffset()
        }
	rec.Cache = adapter
	for _, fn := range opts {
		fn(&rec)
	}
```

So you can remove this if block. 

I think using typical option pattern looks good

flycash (changes requested):

PR #5071, opened 2022-09-30
Add:Get response with media type

Feizekai:
In my work, there is a business situation: the client needs content-type in the server's return to be in other formats by setting the content-type of the request header to the json format. So I want to add a http request method that allows the client to specify the content-type of the return body while sending the request.

### DIFF
--- a/client/httplib/filter.go
+++ b/client/httplib/filter.go
@@ -21,4 +21,8 @@ import (
 
 type FilterChain func(next Filter) Filter
 
+type FilterChainWithMedia func(next FilterWithMedia) FilterWithMedia
+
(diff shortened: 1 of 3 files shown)


Feizekai:
The fault from DeepSource Analysis is native

flycash:
I don't get your point, you can specify the media type when you construct the http request

Feizekai:
> I don't get your point, you can specify the media type when you construct the http request

But sometimes we need to define the media tpye as json in request and get a response with zip

> I don't get your point, you can specify the media type when you construct the http request


But sometimes we need to define the media tpye as json in request and get a response with zip

flycash:
I mean, it looks like you can use filter directory? no need to have a new
structure ?

PR #5102, opened 2022-11-22
fix: revise the body wrapper to handle empty body case

chlins:
Fix the router.go serverHttp method, wrap the body if the request body is empty, which can avoid panic when calling the CopyBody method.

Signed-off-by: chlins <chenyuzh@vmware.com>

### DIFF
--- a/server/web/router.go
+++ b/server/web/router.go
@@ -15,8 +15,10 @@
 package web
 
 import (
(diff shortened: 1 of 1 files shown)


wy65701436 (approved):
lgtm

flycash (approved):

PR #5116, opened 2022-12-09
feature extend readthrough for cache module

Stone-afk:


client/cache/read_through.go
```diff
@@ -0,0 +1,73 @@
...
+	"strings"
+	"sync"
+	"time"
+)
+
+// readThroughCache is a decorator
+// add the read through function to the original Cache function
+type readThroughCache struct {
+	Cache
+	mutex      sync.RWMutex
```
flycash, line 16: you don't need the mutex. Just let the Cache to protect its resources.

client/cache/read_through.go
```diff
@@ -0,0 +1,67 @@
...
+		return nil, berror.Error(InvalidLoadFunc, "loadFunc cannot be nil")
+	}
+	return &readThroughCache{
+		Cache:      cache,
+		Expiration: expiration,
+		LoadFunc:   loadFunc,
+	}, nil
+}
+
+// Get cache from readThroughCache
```
flycash, line 32: Get will try to call the LoadFunc to load data if the Cache returns value nil or non-nil error.

client/cache/read_through.go
```diff
@@ -0,0 +1,67 @@
...
+		err = c.Cache.Put(ctx, key, val, c.Expiration)
+		if err != nil {
+			return val, err
+		}
+	}
+	return val, nil
+}
+
+// GetMulti cache from readThroughCache
+func (c *readThroughCache) GetMulti(ctx context.Context, keys []string) ([]any, error) {
```
flycash, line 50: You should call the c.Cache.GetMulti, it looks like:
```go
vals, err := c.Cache.GetMulti()
if err != nil {
    // call load functions
}
```

flycash: And then the comment should be:
GetMulti will call the LoadFunc to load data if the Cache.GetMulti returns non-nil error. You should check the concrete type of Cache to learn the cases that the GetMulti function will returns non-nil error.

client/cache/error_code.go
```diff
@@ -123,6 +123,10 @@ var InvalidSsdbCacheValue = berror.DefineCode(4002022, moduleName, "InvalidSsdbC
 SSDB cache only accept string value. Please check your input.
 `)
 
+var InvalidLoadFunc = berror.DefineCode(4002023, moduleName, "InvalidLoadFunc", `
+Invalid loadFunc.
```
flycash, line 127: Invalid load function for read-through pattern decorator.
You should pass a valid(non-nil) load function when initiate the decorator instance

client/cache/read_through.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+// Get cache from readThroughCache
+func (c *readThroughCache) Get(ctx context.Context, key string) (any, error) {
+	val, err := c.Cache.Get(ctx, key)
+	if val == nil || err != nil {
+		val, err = c.LoadFunc(ctx, key)
+		if err != nil {
+			return nil, berror.Wrap(
+				err, KeyNotExist, "cache unable to load data")
```
flycash, line 39: Define a new error instead of using KeyNotExit, i.e FailedToLoadData. As you can see, the key may be exist.

client/cache/read_through.go
```diff
@@ -0,0 +1,67 @@
...
+	"strings"
+	"time"
+)
+
+// readThroughCache is a decorator
+// add the read through function to the original Cache function
+type readThroughCache struct {
+	Cache
+	Expiration time.Duration
+	LoadFunc   func(ctx context.Context, key string) (any, error)
```
flycash, line 16: you can make them as private

client/cache/read_through.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+// NewReadThroughCache create readThroughCache
+func NewReadThroughCache(cache Cache, expiration time.Duration,
```
flycash, line 20: How about using a flag to indicate if the GetMulti will reuse the LoadFunction? looks like:
```go
if c.applyForGetMulti {
    // call load func
}
```
And the default value is false. I want to make the users to pay more attention to the GetMulti function.
They must call the NewReadThroughCache(..., ApplyForGetMulti), so they can understand we will call the LoadFunc by iterating the keys.

ApplyForGetMulti should be option pattern.

client/cache/read_through.go
flycash: The error code shout start with 4 because it's user's input error(the load func is invalid)

flycash (changes requested):

client/cache/read_through.go
Stone-afk: 这里直接使用 GetMulti 有种情况就是 err 为nil， 但是 vals 里的值为空接口，所以  GetMulti  本质上是无解的，所以考虑放弃支持 GetMulti

flycash (approved):

flycash:
@Stone-afk add this MR into CHANGLOG.md

flycash:
and please fix the golang-ci issues

PR #5117, opened 2022-12-10
feature: add write though for cache mode

hookokoko:
feature: add write though  cache mode


client/cache/write_through.go
```diff
@@ -0,0 +1,24 @@
...
+
+type WriteThoughCache struct {
+	Cache
+	StoreFunc func(ctx context.Context, key string, val any) error
+}
+
+func (w *WriteThoughCache) Set(ctx context.Context, key string, val any, expiration time.Duration) error {
+	if w.StoreFunc == nil {
+		return berror.Error(InvalidStoreFunc, "storeFunc can not be nil")
+	}
```
flycash, line 18: It looks not very elegant. The alternative solutions are:
- remove this validation and we do nothing. Users will get panic when we call StoreFunc. But it's not a big deal because users can easily find the panic if they write any tests. 
- make StoreFunc as private and add one more NewWriteThroughCache function. In this function check if the users pass valid StoreFunc

flycash (changes requested):
Just kindly reminds that here is a decoration pattern, so actually you don't need to write tests fro memcache, redis, ssdb because this decorator knows nothing about underlying Cache type. It doesn't care

client/cache/write_through.go
hookokoko: I have chosen the second solution,  I think it is clearer to use

hookokoko:
> Just kindly reminds that here is a decoration pattern, so actually you don't need to write tests fro memcache, redis, ssdb because this decorator knows nothing about underlying Cache type. It doesn't care

client/cache/error_code.go
```diff
@@ -123,6 +123,14 @@ var InvalidSsdbCacheValue = berror.DefineCode(4002022, moduleName, "InvalidSsdbC
 SSDB cache only accept string value. Please check your input.
 `)
 
+var InvalidInitParameters = berror.DefineCode(4002024, moduleName, "InvalidInitParameters", `
+Invalid init cache parameters.
```
flycash, line 127: Invalid init cache parameters.
You can check the related function to confirm that if you pass correct parameters or configure to initiate a Cache instance.

client/cache/error_code.go
```diff
@@ -123,6 +123,14 @@ var InvalidSsdbCacheValue = berror.DefineCode(4002022, moduleName, "InvalidSsdbC
...
+`)
+
+var PersistCacheFailed = berror.DefineCode(4002025, moduleName, "PersistCacheFailed", `
+StoreFunc execute failed.
```
flycash, line 131: Failed to execute the StoreFunc.
Please check the log to make sure the StoreFunc works for the specific key and value

flycash:
please add this MR into CHANGELGO and fix the golangci lint issues.

client/cache/error_code.go
flycash: @hookokoko please change this

client/cache/error_code.go
flycash: @hookokoko please change this.

client/cache/error_code.go
hookokoko: All right, I know what do you mean

client/cache/error_code.go
hookokoko: All right

client/cache/error_code.go
hookokoko: Do you mean that there is no need to define error codes？

client/cache/error_code.go
hookokoko: Sorry, I don’t understand what you mean.  Do you mean that there is no need to define error codes？

flycash:
install  the golangci on your computer, and run golang ci

flycash:
I think you can install the goimport to format the code. It can solve many issues

client/cache/error_code.go
```diff
@@ -123,6 +123,16 @@ var InvalidSsdbCacheValue = berror.DefineCode(4002022, moduleName, "InvalidSsdbC
 SSDB cache only accept string value. Please check your input.
 `)
 
+var InvalidInitParameters = berror.DefineCode(4002024, moduleName, "InvalidInitParameters", `
+Invalid init cache parameters.
```
flycash, line 127: The read through pattern already use the code 4002024, so you need to use a new one. 😄

flycash (commented):
Just skip the golangci action, it looks something wrong with that

PR #5119, opened 2022-12-11
feature add singleflight cache

Stone-afk:
add singleflight cache, test case not supplemented


flycash:
please  rebase to origin/develop and fix the conflicts

Stone-afk:
目前针对 ssdb 的并发测试，不管设置的 value 是啥最终都会会持续返回 “1”

Stone-afk:
`
func testSingleflightCacheConcurrencyGet(t *testing.T, bm cache.Cache) {
	key, value := "key3", "value3"
	db := &MockOrm{keysMap: map[string]int{key: 1}, kvs: map[string]any{key: value}}
	c, err := cache.NewSingleflightCache(bm, 10*time.Second,
		func(ctx context.Context, key string) (any, error) {
			val, er := db.Load(key)
			if er != nil {
				return nil, er
			}
			return val, nil
		})
	assert.Nil(t, err)

	var wg sync.WaitGroup
	wg.Add(10)
	for i := 0; i < 10; i++ {
		go func() {
			defer wg.Done()
			val, err := c.Get(context.Background(), key)
			if err != nil {
				t.Error(err)
			}
			assert.Equal(t, value, val)
		}()
	}
	wg.Wait()
}
`

Stone-afk:
```
func testSingleflightCacheConcurrencyGet(t *testing.T, bm cache.Cache) {
	key, value := "key3", "value3"
	db := &MockOrm{keysMap: map[string]int{key: 1}, kvs: map[string]any{key: value}}
	c, err := cache.NewSingleflightCache(bm, 10*time.Second,
		func(ctx context.Context, key string) (any, error) {
			val, er := db.Load(key)
			if er != nil {
				return nil, er
			}
			return val, nil
		})
	assert.Nil(t, err)


	var wg sync.WaitGroup
	wg.Add(10)
	for i := 0; i < 10; i++ {
		go func() {
			defer wg.Done()
			val, err := c.Get(context.Background(), key)
			if err != nil {
				t.Error(err)
			}
			assert.Equal(t, value, val)
		}()
	}
	wg.Wait()
}

```

flycash (changes requested):
remove other tests. Just keep singleflight_test

client/cache/singleflight.go
```diff
@@ -0,0 +1,66 @@
...
+	"golang.org/x/sync/singleflight"
+)
+
+// SingleflightCache
+// This is a very simple decorator mode
+type SingleflightCache struct {
+	Cache
+	group      *singleflight.Group
+	Expiration time.Duration
+	LoadFunc   func(ctx context.Context, key string) (any, error)
```
flycash, line 32: make them private

client/cache/singleflight.go
```diff
@@ -0,0 +1,63 @@
...
+		expiration: expiration,
+		loadFunc:   loadFunc,
+	}, nil
+}
+
+// Get In the Get method, single flight is used to load data and write back the cache.
+func (s *SingleflightCache) Get(ctx context.Context, key string) (any, error) {
+	val, err := s.Cache.Get(ctx, key)
+	if val == nil || err != nil {
+		val, err, _ = s.group.Do(key, func() (interface{}, error) {
```
flycash, line 53: I think you need to delete the key after loading data. So when the key is expired again, goroutine will load data one more time

flycash (approved):

PR #5120, opened 2022-12-11
feature: add preload cache

hookokoko:
support preload cache when key is expired by sentinel memory cache


flycash (commented):
suspends it until write-back was merged. Or you can split the preload cache to another merge request.

flycash:
you can split this merge request into two merge request, one for evicted callback, one for preload

hookokoko:
I have no good idea to solve the problem mention in write-back, so closed temporarily...

PR #5126, opened 2022-12-17
Bloom filter cache

hookokoko:
feature: add bloom filter cache

### DIFF
--- a/client/cache/bloom_filter_cache.go
+++ b/client/cache/bloom_filter_cache.go
@@ -20,17 +20,21 @@ import (
 	"time"
 
 	"github.com/beego/beego/v2/core/berror"
-	"github.com/bits-and-blooms/bloom/v3"
 )
 
 type BloomFilterCache struct {
 	Cache
-	*bloom.BloomFilter
+	BloomFilter
 	loadFunc   func(ctx context.Context, key string) (any, error)
 	expiration time.Duration // set cache expiration, default never expire
 }
 
-func NewBloomFilterCache(cache Cache, ln func(context.Context, string) (any, error), blm *bloom.BloomFilter,
+type BloomFilter interface {
+	Test(data string) bool
+	Add(data string)
+}
+
+func NewBloomFilterCache(cache Cache, ln func(context.Context, string) (any, error), blm BloomFilter,
 	expiration time.Duration,
 ) (*BloomFilterCache, error) {
 	if cache == nil || ln == nil || blm == nil {
@@ -51,7 +55,7 @@ func (bfc *BloomFilterCache) Get(ctx context.Context, key string) (any, error) {
 		return nil, err
 	}
 	if errors.Is(err, ErrKeyNotExist) {
-		exist := bfc.BloomFilter.TestString(key)
+		exist := bfc.BloomFilter.Test(key)
 		if exist {
 			val, err = bfc.loadFunc(ctx, key)
 			if err != nil {
--- a/client/cache/bloom_filter_cache.go
+++ b/client/cache/bloom_filter_cache.go
@@ -31,8 +31,8 @@ type BloomFilterCache struct {
 }
 
 func NewBloomFilterCache(cache Cache, ln func(context.Context, string) (any, error), blm *bloom.BloomFilter,
-	expiration time.Duration) (*BloomFilterCache, error) {
-
+	expiration time.Duration,
+) (*BloomFilterCache, error) {
 	if cache == nil || ln == nil || blm == nil {
 		return nil, berror.Error(InvalidInitParameters, "missing required parameters")
 	}
--- a/client/cache/bloom_filter_cache.go
+++ b/client/cache/bloom_filter_cache.go
@@ -1,3 +1,17 @@
+// Copyright 2014 beego Author. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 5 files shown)


client/cache/bloom_filter_cache.go
```diff
@@ -0,0 +1,84 @@
...
+	Cache
+	*bloom.BloomFilter
+	loadFunc       func(ctx context.Context, key string) (any, error)
+	expiration     time.Duration // set cache expiration, default never expire
+	notUpdateBloom bool          // update bloom key after put cache, default update
+}
+
+type BloomFilterCacheOption func(bfc *BloomFilterCache)
+
+func WithExpirationOpt(t time.Duration) BloomFilterCacheOption {
```
flycash, line 22: Maybe we can add some prefix to avoid potential conflicts, i.e. `BFCWithExpiration()` ...
Expiration should be a common configure for other cache implementation.

client/cache/bloom_filter_cache.go
```diff
@@ -0,0 +1,84 @@
...
+	return func(bfc *BloomFilterCache) {
+		bfc.expiration = t
+	}
+}
+
+func WithUpdateBloomOpt(update bool) BloomFilterCacheOption {
```
flycash, line 28: BFCWithUpdateBloom, or you can use other name if you have any other good idea

client/cache/bloom_filter_cache.go
```diff
@@ -0,0 +1,84 @@
...
+			err = bfc.Put(ctx, key, val, bfc.expiration)
+			if err != nil {
+				return val, err
+			}
+		}
+	}
+	return val, nil
+}
+
+func (bfc *BloomFilterCache) Put(ctx context.Context, key string, val any, expiration time.Duration) error {
```
flycash, line 75: What about let the users update the bloom filter? I mean, if the cache is a local cache, users must initiate the bloom filter by query all data from database. In this case, it's users' responsible to make sure the bloom filter always is correct.
Another concern is that, if users delete data, they must delete data from bloom filter.

So I think we don't need to update the bloom filter, just let the users do it.

client/cache/bloom_filter_cache_test.go
```diff
@@ -0,0 +1,139 @@
...
+	bfc.BloomFilter.ClearAll()
+	for i := 1; i <= 100; i++ {
+		bfc.AddString(fmt.Sprintf("key%d", i))
+	}
+	wg1 := sync.WaitGroup{}
+	wg1.Add(100)
+	for i := 0; i < 100; i++ {
+		go func() {
+			defer wg1.Done()
+			_, err = bfc.Get(context.Background(), fmt.Sprintf("key%d", i))
```
flycash, line 113: you can not use i here due to race condition. the value of i when execute this statement is unpredictable.

client/cache/bloom_filter_cache_test.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/bits-and-blooms/bloom/v3"
+	"github.com/stretchr/testify/assert"
+)
+
+type Mock struct {
+	Db      map[string]any
+	loadCnt int64
+}
+
+func TestBloomFilterCache(t *testing.T) {
```
flycash, line 20: refactor the test to be table driven, like:
```go
testCases = []struct {
    name string
    mock func() // can return something or accept something
    clean func()
    wantVal any
    wantErr error
}
```
So the test cases are independent and more readable

flycash (changes requested):
Please help to fix the golanci issues and deepsource issues

client/cache/bloom_filter_cache.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"github.com/beego/beego/v2/core/berror"
+	"github.com/bits-and-blooms/bloom/v3"
+)
+
+type BloomFilterCache struct {
+	Cache
+	*bloom.BloomFilter
+	loadFunc       func(ctx context.Context, key string) (any, error)
+	expiration     time.Duration // set cache expiration, default never expire
```
flycash, line 16: I think users must pass the expiration when they call the NewBloomFilterCache because we don't know how to handle the expiration. Using default value looks not elegant

client/cache/bloom_filter_cache_test.go
```diff
@@ -0,0 +1,204 @@
...
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/beego/beego/v2/core/berror"
+	"github.com/bits-and-blooms/bloom/v3"
+	"github.com/stretchr/testify/assert"
+)
+
+type MockDB struct {
```
flycash, line 31: can you use the memory cache implementation instead of MockDB?

flycash (commented):
Good design for test cases. But you still need to fix the github action. You can skip the golan-ci action

client/cache/bloom_filter_cache_test.go
hookokoko: No problem.
I just worry about two cache may cause some misunderstanding. Normally, the data should load from persistent storage when it not exist in cache.

flycash:
image

hookokoko:
> image

Because of the bloom package is NOT safe for concurrent use, so I remove this test cases. And if someone want safe for concurrent use, they can change bloom by decorator mode. So I think we should supply an interface of bloom filter, rather than specific implementation.

```go
type BloomFilter struct {
	*bloom.BloomFilter
	lock       *sync.RWMutex
	concurrent bool
}

func (b *BloomFilter) Exist(data []byte) bool {
	if b.concurrent {
		b.lock.Lock()
		defer b.lock.Unlock()
	}
	return b.BloomFilter.Test(data)
}
```

flycash (approved):

PR #5132, opened 2022-12-21
fix 5079: only log msg when the channel is not closed

flycash:


flycash:
I think I have address the issue, but not sure.

PR #5171, opened 2023-03-06
Provision to reset database cache

msiripurapu05:
**Reason to Update:**
I am trying to write UT for DB connections by mocking db using github.com/DATA-DOG/go-sqlmock. I want to recreate mock db and set to orm just before execution each test case. However the current implementation of orm.AddAliasWthDB is not allowing to reset db driver once it is initialized. It will throw the error fmt.Errorf("DataBase alias name %s already registered, cannot reuse", aliasName)

### DIFF
--- a/orm/db_alias.go
+++ b/orm/db_alias.go
@@ -467,3 +467,7 @@ func newStmtDecoratorLruWithEvict() *lru.Cache {
 	})
 	return cache
 }
+
+func ResetDataBaseCache() {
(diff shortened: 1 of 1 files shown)


msiripurapu05:
@flycash Thank you for merging the changes. Could you please tag it to a release. So, we can start consuming the method.

flycash:
okay

flycash:
check it! hope you can help to improve Beego v1.x in the future

PR #5190, opened 2023-04-16
bugfix: beegoAppConfig String and Strings function has bug

Stone-afk:


flycash:
close it and then you can reopen it

PR #5191, opened 2023-04-18
bugfix: CopyRequestBody behave

Panlibin2015:
1.CopyRequestBody default value: false -> true
2.fix BeegoInput.Bind* func is not available when CopyRequestBody=false

### DIFF
--- a/server/web/config.go
+++ b/server/web/config.go
@@ -528,7 +528,7 @@ func newBConfig() *Config {
 		ServerName:          "beegoServer:" + beego.VERSION,
 		RecoverPanic:        true,
 
-		CopyRequestBody:    false,
+		CopyRequestBody:    true,
 		EnableGzip:         false,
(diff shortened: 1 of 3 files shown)


flycash:
sorry, you can not change the default value. And can you rebase beego/develop? sorry for missing this merge request for a long time.

flycash:
In fact, we quest the users to specify the CopyValue as true. So I think it's not a bug, and if you still believe that this is a BUG, you can reopen the merge request.

Panlibin2015:
yes, it's not a bug
in fact, i think beego to specify CopyValue as true is right. because i believe that is users expect.
second, when CopyValue=false , the function which BeegoInput.Bind* is not available, that‘s a BUG.

CopyValue‘s value cannnot affect function's behave  which BeegoInput.Bind*

PR #5195, opened 2023-04-30
support db sharding.

ethanvc:
If we have many sharding database or databases in one data source, current solution to access the data is by register as many database as the sharding db or databases count. This is not efficient because of the connection can not be shared between databases.

In this PR, I give an easy solution to specify the full qualified table name so users can set database name in the under ground sql, which solve the problem above.

I have write a demo here for your reference: https://github.com/ethanvc/quickstart/blob/main/beego_orm/sharding_support/main.go

### DIFF
--- a/client/orm/db.go
+++ b/client/orm/db.go
@@ -280,7 +280,7 @@ func (d *dbBase) PrepareInsert(ctx context.Context, q dbQuerier, mi *modelInfo) 
 	sep := fmt.Sprintf("%s, %s", Q, Q)
 	columns := strings.Join(dbcols, sep)
 
-	query := fmt.Sprintf("INSERT INTO %s%s%s (%s%s%s) VALUES (%s)", Q, mi.table, Q, Q, columns, Q, qmarks)
+	query := fmt.Sprintf("INSERT INTO %s (%s%s%s) VALUES (%s)", d.getQuotedTableName(mi.table), Q, columns, Q, qmarks)
 
 	d.ins.ReplaceMarks(&query)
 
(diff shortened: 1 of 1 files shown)


ethanvc:
@flycash hello, can you please review this pr? thanks.

ethanvc:
> Sorry, I am afraid that I can not accept this merger request. Sharding is a big topic. Sharding indicates things:
> 
> * rewrite SQL. That's just what you did here. But actually things are complicated, because you can image that there are several kinds sharding rules, like hash, range, or region. For example, $region.mycompany.com:3306/xxxx_db_$id%1000 and the table is xxx_tab_%uid[0, 100]. You can take shardingsphere as reference
> * executes SQL. Step1 may generate several SQL and you need to route them to different data sources. And then collect the data
> * merge the results. for example, you need to support pagination, order by, aggregate functions.
> 
> So I really appreciate your effort, but I think supporting the sharding in Beego ORM is a hard job and I still have no plan for it.

My initial problem is how to specify the database in my query explicitly, i do not want to register database for every database in one data source. but i have not find the method to do it.
For you concerns:
- rewrite SQL item. I think my pr already solved the problem. Users can define there own sharding policy by write their own TableName member function of modal. Users also have to write a function to chose the right data source.
- executes SQL. Currently in my thinking, I can write a function to choose the alias name for my querys. Because currently, beego orm need to select database first to do following steps.
- merge results. Maybe in the first step, we do not support queries span data sources, users can do that outside the beeorm.

But my key problem now is i can not specify database name in my sql query, if i missed something, just let me know.

PR #5211, opened 2023-05-13
bugfix: protect field access with lock to avoid possible data race

Stone-afk:


server/web/session/couchbase/sess_couchbase.go
```diff
@@ -105,8 +106,9 @@ func (cs *SessionStore) SessionID(context.Context) string {
 // SessionRelease Write couchbase session with Gob string
 func (cs *SessionStore) SessionRelease(ctx context.Context, w http.ResponseWriter) {
 	defer cs.b.Close()
-
+	cs.lock.Lock()
```
flycash, line 109:
```go
cs.lock.RLock()
values := cs.values
cs.Lock.RUnlock()
session.EncodeGob(values)
```
Take this as an example

flycash (changes requested):
1. Change to use the RWMutex
2. Reduce the scope of the lock
3. Remove useless files

flycash (approved):

PR #5245, opened 2023-06-14
feat: add write-double-delete cache mode

Stone-afk:


client/cache/write_delete.go
```diff
@@ -47,3 +49,39 @@ func (w *WriteDeleteCache) Set(ctx context.Context, key string, val any) error {
...
+func (c *WriteDoubleDeleteCache) Set(ctx context.Context, key string, val any) error {
+	err := c.storeFunc(ctx, key, val)
+	if err != nil && !errors.Is(err, context.DeadlineExceeded) {
+		return berror.Wrap(err, PersistCacheFailed, fmt.Sprintf("key: %s, val: %v", key, val))
+	}
+	go func() {
+		timer := time.NewTimer(c.interval)
+		defer timer.Stop()
+		<-timer.C
+		_ = c.Cache.Delete(ctx, key)
```
flycash, line 83: What about using the time.AfterFunc here?
and when you call the c.Cache.Delete, you can not use the ctx and please create a new one with 1s expiration.

client/cache/write_delete.go
```diff
@@ -47,3 +49,39 @@ func (w *WriteDeleteCache) Set(ctx context.Context, key string, val any) error {
...
+	}()
+	return c.Cache.Delete(ctx, key)
```
flycash, line 85: wrap this error and tell the users that Write-Double-Delete failed to delete the key from cache

client/cache/write_delete_test.go
```diff
@@ -19,13 +19,189 @@ import (
...
+				return
+			}
+
+			err = c.Set(tt.ctx, tt.key, tt.value)
+			if err != nil {
+				assert.EqualError(t, tt.wantErr, err.Error())
+				return
+			}
+
+			err = cache.Put(tt.ctx, tt.key, tt.value, tt.interval)
```
flycash, line 116: before Put, you need to call Get to make sure the c.Set already delete the key. In general:
```go
c := NewWriteDouble
cache.Set()
c.Set()
c.Get()
// first round assertion
c.Put()
sleep()
c.Get()
// second round assertion
```

client/cache/write_delete.go
```diff
@@ -47,3 +49,39 @@ func (w *WriteDeleteCache) Set(ctx context.Context, key string, val any) error {
 	}
 	return w.Cache.Delete(ctx, key)
 }
+
+// WriteDoubleDeleteCache creates write double delete cache pattern decorator.
```
flycash, line 53: it will delete the key from cache when you call Set function, and wait for `interval`, it will delete the key from cache one more time.
This pattern help to reduce the possibility of data inconsistencies, but it's still possible to be inconsistent among database and cache.

client/cache/write_delete_test.go
Stone-afk: mark

Stone-afk:
mark

Stone-afk:
mark

PR #5257, opened 2023-06-20
closes 5254: %COL% should be a common placeholder

flycash:


flycash:
I don't know why the postgresql using the special %COL% placeholder, but I think the best way is to replace it in defer function.

PR #5261, opened 2023-06-23
fix: use of ioutil package

yagikota:
`io/ioutil` package has been deprecated since Go 1.16 as described [here](https://go.dev/doc/go1.16#ioutil).
So,I have fixed the relevant parts of the file except for [this auto-generated file](https://github.com/beego/beego/blob/develop/test/bindata.go).


yagikota:
Screenshot 2023-06-23 at 23 19 07

According to SonarCloud, there are several duplicated lines in the following files
- `core/logs/alils/log_project.go`
- `core/logs/alils/log_store.go`
- `core/logs/alils/machine_group.go`

But, I think It can't be helped. 
what do you think?

flycash (approved):

PR #5272, opened 2023-07-04
Fix/refactor update sql

uzziahlin:
refactor update sql and add test


flycash:
please help to solve the conflicts

Issue #5106, opened 2022-11-22
embed.FS with Beego


Nigel2392:
1. What version of Go and beego are you using (`bee version`)?
2.0.5

2. What operating system and processor architecture are you using (`go env`)?
Windows, amd64

3. What did you expect to see?
A way to embed templates when building the application

4. What did you see instead?
No clear documentation on this subject.

Basically, I have a question.
Is there a way to embed the HTML files with the executable for distribution?
Could a feature like this be implemented by users of beego?
I did not see any clear documentation on this subject, but I might have missed it.

Nigel2392:
```
//go:embed core/views/*
//go:embedassets/*
var ViewFS embed.FS

func init(){
    ....
    web.SetTemplateFSFunc(
		func() http.FileSystem {
			return http.FS(ViewFS)
		},
	)
    web.SetViewsPath("core/views")
    ....
}
```
Panics that a few directories inside the embed cannot be found.
```
panic: open core\views\admin: file does not exist
```
But I can use `ViewFS.ReadDir("core\views\admin")` and see its contents perfectly fine.

deatil:
@Nigel2392 this is work:

~~~go
//go:embed core/views/*
//go:embed assets/*
var ViewFS embed.FS

type FileSystem struct{
    fs http.FileSystem
}

func (this FileSystem) Open(name string) (http.File, error) {
    name = strings.Replace(name, "\\", "/", -1)

    return this.fs.Open(name)
}

func init(){
    ....
    web.SetTemplateFSFunc(
		func() http.FileSystem {
			return FileSystem{http.FS(ViewFS)}
		},
	)
    web.SetViewsPath("core/views")
    ....
}
~~~

PR #5279, opened 2023-07-19
fixed: canot use embed template

deatil:


flycash:
can you provide more details about this fix?
including the issue you want to solve and how to reproduce it.

deatil:
> can you provide more details about this fix? including the issue you want to solve and how to reproduce it.

issue #5106

deatil:
@flycash hi, i find can not use, embed need wrap open and it can work.

PR #5295, opened 2023-08-10
fix: refactor UpdateBatch method

uzziahlin:
refactor UpdateBatch method and add test

### DIFF
--- a/client/orm/db.go
+++ b/client/orm/db.go
@@ -936,63 +936,6 @@ func (d *dbBase) getSetSQL(cols []string, values []interface{}) string {
 	return buf.String()
 }
 
-func (d *dbBase) UpdateBatchSQL1(columns []string, values []interface{}, mi *models.ModelInfo, specifyIndexes, join, where string) string {
-	var query, T string
-
-	Q := d.ins.TableQuote()
-
-	if d.ins.SupportUpdateJoin() {
-		T = "T0."
-	}
-
-	cols := make([]string, 0, len(columns))
-
-	for i, v := range columns {
-		col := fmt.Sprintf("%s%s%s%s", T, Q, v, Q)
-		if c, ok := values[i].(colValue); ok {
-			switch c.opt {
-			case ColAdd:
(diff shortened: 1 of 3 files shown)


client/orm/db.go
```diff
@@ -819,63 +819,121 @@ func (d *dbBase) UpdateBatch(ctx context.Context, q dbQuerier, qs *querySet, mi
...
+				_, _ = buf.WriteString(" >> ?")
 			case ColBitLShift:
-				cols = append(cols, col+" = "+col+" << ?")
+				_, _ = buf.WriteString(" << ?")
 			case ColBitXOR:
-				cols = append(cols, col+" = "+col+" ^ ?")
+				_, _ = buf.WriteString(" ^ ?")
 			case ColBitOr:
-				cols = append(cols, col+" = "+col+" | ?")
+				_, _ = buf.WriteString(" | ?")
```
flycash, line 928: please add test cases to cover all these branches

uzziahlin: Okay.

client/orm/db.go
```diff
@@ -819,63 +819,121 @@ func (d *dbBase) UpdateBatch(ctx context.Context, q dbQuerier, qs *querySet, mi
...
+	}
+
+	query := buf.String()
+
+	d.ins.ReplaceMarks(&query)
+
+	return query
+}
+
+func (d *dbBase) getSetSQL(cols []string, values []interface{}) string {
```
flycash, line 881: refactor this function:
```go
func (d *dbBase) buildSetSQL(bs Buffer, cols []string, values []any) {

}
```
we want to reuse the buffer to reduce the memory allocation.
And you can still keep passing an empty buffer into your unit tests.
After that, rename the T to `owner` and Q to `quote`, T and Q are not good names.

flycash: if you don't like the name `owner`, you can give it a new name.

uzziahlin: Okay.

PR #5300, opened 2023-08-24
fix: refactor Count method

uzziahlin:
refactor Count method and add test


client/orm/db.go
```diff
@@ -1372,39 +1379,42 @@ func (d *dbBase) readSQL(tables *dbTables, tCols []string, cond *Condition, qs *
...
-	groupBy := tables.getGroupSQL(qs.groups)
-	tables.getOrderSQL(qs.orders)
-	join := tables.getJoinSQL()
-	specifyIndexes := tables.getIndexSql(mi.Table, qs.useIndex, qs.indexes)
+	buf := buffers.Get()
+	defer buffers.Put(buf)
 
-	Q := d.ins.TableQuote()
+	if len(qs.groups) > 0 {
+		_, _ = buf.WriteString("SELECT COUNT(*) FROM (")
```
flycash, line 1403: why do you need to write string "SELECT COUNT(*) FROM " here?
I found that function `readSQL`:
```go
	} else {
		_, _ = buf.WriteString(qs.aggregate)
	}
```

uzziahlin: Because if grouped, each group will generate one piece of data, and multiple pieces of data will be generated. A query should be nested outside to query the number of groups.

flycash (approved):

PR #5308, opened 2023-08-28
select statement currently support From , orderBy , Offset and Limit

smx-Morgan:
Using the Builder pattern to construct SQL,  currently select statement support From , orderBy , Offset and Limit


flycash:
Hi, please add this merge request to CHANGELOG.

flycash:
all actions failed ~~

PR #5404, opened 2023-09-09
support db_type in ddl

smx-Morgan:
When defining a model, users can consider using db_ Type to directly define specific types.


flycash:
all actions failed, please check them.

flycash:
In PostgreSQL, `model_with_d_b_types` should be "model_with_d_b_types" because the PostgreSQL uses the " as the quote.

flycash:
and then, please add this merge request to CHANELOG

PR #5415, opened 2023-09-17
orm: PostgreSQL change auto to bigserial

### LINKED ISSUE #5398: serial use bigserial type
current auto type mapping `serial NOT NULL PRIMARY KEY`, `serial` default is interge type, when we hope use bigint, `serial` should is `bigserial`.

flycash:


flycash:
#5404

PR #5420, opened 2023-09-19
refactor(server/web): remove redundant nil check

Juneezee:
From the Go specification (https://go.dev/ref/spec#For_range):

> "1. For a nil slice, the number of iterations is 0."
> "3. If the map is nil, the number of iterations is 0."

Therefore, an additional nil check for slice and map before the loop is unnecessary.


flycash (approved):

flycash:
Thank you!

PR #5538, opened 2023-11-29
Fix issue 5537 - Send missing app.conf ini error to STDERR

jeffdoubleyou:
Send "ignorable" error message for missing `conf/app.conf` to STDERR instead of using logger.


flycash:
ok, I think it make sense.

PR #5545, opened 2023-12-08
 fix issue 4936: retry not work, copy request body before retry

jptiancai:
reference：https://github.com/beego/beego/issues/4936


flycash:
thank you! let me check this pull request on this weekend

flycash:
#5546

flycash:
check that pull request, I think it's a better solution

PR #5548, opened 2023-12-11
Log file name generation bug

chenmingyong0423:
fix(logs): Fixed a bug where format was empty


flycash:
Some applications which will parse the log files may fail to look up the log files.

PR #5558, opened 2023-12-29
fix: refactor use global DefaultModelCache

Stone-afk:


client/orm/internal/models/models.go
```diff
@@ -22,37 +22,39 @@ import (
 	"sync"
 )
 
-// ModelCache info collection
-type ModelCache struct {
+var DefaultModelRegistry = newModelRegistry()
+
+// modelRegistry info collection
+type modelRegistry struct {
```
flycash, line 28: do not rename this, keep using the name ModelCache

client/orm/dialect/types.go
```diff
@@ -0,0 +1,20 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dialect
+
+type Dialect interface {
```
flycash, line 17: why do you need to introduce this structure? It looks like not necessary.

client/orm/internal/models/models_test.go
```diff
@@ -34,9 +34,9 @@ func (i *Interface) TableEngine() string {
 }
 
 func TestDbBase_GetTables(t *testing.T) {
-	c := NewModelCacheHandler()
```
flycash, line 37: do not change this. All tests should create its own ModelCache to decouple from the global cache.

client/orm/dialect/types.go
```diff
@@ -0,0 +1,20 @@
...
+	// Quoter 返回一个引号，引用列名，表名的引号
```
flycash, line 18: English

flycash (approved):

PR #5562, opened 2024-01-16
feature add deleter

Stone-afk:


client/orm/qb/builder.go
```diff
@@ -0,0 +1,103 @@
...
+
+import (
+	"github.com/valyala/bytebufferpool"
+
+	"github.com/beego/beego/v2/client/orm/internal/models"
+	"github.com/beego/beego/v2/client/orm/qb/errs"
+)
+
+type builder struct {
+	buffer *bytebufferpool.ByteBuffer
```
flycash, line 25: using the internal/buffer package

client/orm/qb/delete.go
```diff
@@ -0,0 +1,105 @@
...
+
+// Deleter builds DELETE query
+type Deleter[T any] struct {
+	builder
+	table interface{}
+	db    orm.Ormer
+	where []Predicate
+}
+
+// NewDeleter 开始构建一个 DELETE 查询
```
flycash, line 37: english

client/orm/qb/delete.go
```diff
@@ -0,0 +1,105 @@
...
+	registry := models.DefaultModelCache
+	d.model, _ = registry.GetByMd(d.table)
+	if d.model == nil {
+		err = registry.Register("", true, d.table)
+		if err != nil {
+			return nil, err
+		}
+		d.model, _ = registry.GetByMd(d.table)
+	}
+	if d.model.Table == "" {
```
flycash, line 63: I think it can not be empty. So you only need to use d.model.Table directly

client/orm/qb/delete.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	registry := models.DefaultModelCache
+	d.model, _ = registry.GetByMd(d.table)
+	if d.model == nil {
+		err = registry.Register("", true, d.table)
+		if err != nil {
+			return nil, err
+		}
+		d.model, _ = registry.GetByMd(d.table)
+	}
```
flycash, line 62: can you provide a new function which will register the model if model not found?
```go
func (m model) GetOrRegister() {
   // ...
}
```
I remember that Beego already has a similar function, you can check it.

client/orm/qb/select.go
```diff
@@ -62,130 +66,87 @@ func (s *Selector[T]) Build() (*Query, error) {
...
 
+func (s *Selector[T]) buildTable() {
+	if s.tableName != "" {
+		if s.tableName[0] == '`' && s.tableName[len(s.tableName)-1] == '`' {
+			s.writeString(s.tableName)
+		} else {
+			s.writeByte('`')
+			s.writeString(s.tableName)
+			s.writeByte('`')
+		}
```
flycash, line 117: if uses specify the tableName, using the tableName without `

flycash (approved):

PR #5606, opened 2024-02-29
upgrade to go-redis/v9

carlhopf:
updating to go-redis/v9, to support up to redis 7 (which f.ex. support new cluster modes, like announce-hostname)

some calls like `provider.SessionInit` pass `nil` as the context
 https://github.com/beego/beego/blob/219051dbc8cad7d92667dff7d69fb6e2f54c868b/server/web/session/session.go#L130

hence had to add some checks in f.ex. `sessions/redis/sess_redis.go` to initalize `ctx = context.Background()` if `ctx == nil`, as `ctx` is required by go-redis/v9

### DIFF
--- a/server/web/session/redis/sess_redis.go
+++ b/server/web/session/redis/sess_redis.go
@@ -41,7 +41,7 @@ import (
 	"sync"
 	"time"
 
-	"github.com/go-redis/redis/v7"
+	"github.com/redis/go-redis/v9"
 
 	"github.com/beego/beego/v2/server/web/session"
 )
@@ -109,7 +109,7 @@ func (rs *SessionStore) SessionRelease(ctx context.Context, w http.ResponseWrite
 		return
 	}
 	c := rs.p
-	c.Set(rs.sid, string(b), time.Duration(rs.maxlifetime)*time.Second)
+	c.Set(ctx, rs.sid, string(b), time.Duration(rs.maxlifetime)*time.Second)
 }
 
 // Provider redis session provider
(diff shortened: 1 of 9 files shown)


server/web/session/redis/sess_redis.go
```diff
@@ -240,34 +251,46 @@ func (rp *Provider) SessionRead(ctx context.Context, sid string) (session.Store,
...
 	}
 	return true, nil
 }
 
 // SessionRegenerate generate new sid for redis session
 func (rp *Provider) SessionRegenerate(ctx context.Context, oldsid, sid string) (session.Store, error) {
+	if ctx == nil {
+		ctx = context.Background()
+	}
+
```
flycash, line 271: I think we don't need this checkpoint because I think when the users call those methods with `context.Context` as the parameter, they should never pass nil.

So we don't need to be a babysitter.

flycash (changes requested):

server/web/session/redis/sess_redis.go
carlhopf: thanks for pointing this out! with the `nil` check removed, tests still pass.

but there's some code that calls `SessionRegenerate(nil, ...)` (same for `SessionRead`, `SessionExist`, `SessionDestroy` and `SessionInit`)

https://github.com/beego/beego/blob/219051dbc8cad7d92667dff7d69fb6e2f54c868b/server/web/session/session.go#L313C35-L313C47

does this need refactoring?

flycash: change to `SessionRegenerate(context.Backgroud())` 😄

carlhopf:
removed the `nil` checks and added some `context.Background`s 👍

PR #5621, opened 2024-03-31
fix 5620: ensure cookie always use the config

flycash:


flycash:
@Soringg please help to confirm if this merge request solve your problem.

flycash:
I will release a new version once you confirmed.

Soringg:
@flycash I have tested only the updated code for function `func (manager *Manager) SessionRegenerateID(w http.ResponseWriter, r *http.Request) (Store, error)`  and it works as expected. 

Thank you

PR #5631, opened 2024-04-06
remove unneccesary elements in keys in Signature()

seiyab:
I guess length is mistakenly specified confused with capacity.
This change should slightly improve performance. Fortunately, this mistake has not caused any bugs, I suppose.


flycash:
Hi, can you add one unit test?

seiyab:
This function already has an unit test.
https://github.com/beego/beego/blob/cca1f2f6e6976cf8fad44e22e3e1b4fa9423406b/server/web/filter/apiauth/apiauth_test.go#L8-L20

And this change shouldn't change output. Feel free to correct me if I misunderstand.

flycash (approved):

PR #5633, opened 2024-04-07
resolve #5604: using double instead of single hyphen argument 

### LINKED ISSUE #5604: Using --graceful instead of -graceful since single hyphen is treaded as short version of a flag
   the latest

   irrelevent to the problem
3. What did you do?
   I have a project that is using both cobra and beego, when beego's graceful restart functionality is enabled, cobra treats "- graceful" as "-g" + "raceful"
    beego should restart the program using "--graceful" instead of "-graceful"

5. What did you see instead?
   cobra now treats -graceful as "-g" + "raceful", and will return an error.

6. Why
   The single hyphen is used with "short" argument like this:
   `./bin/abc -h`
   While the double hyphen is used with "long" argument like this:
   `./bin/abc --help`

IMHO, we should follow the common wisdom and make a small change:
image

bo-er:
resolve #5604

### DIFF
--- a/server/web/grace/server.go
+++ b/server/web/grace/server.go
@@ -338,15 +338,15 @@ func (srv *Server) fork() (err error) {
 	var args []string
 	if len(os.Args) > 1 {
 		for _, arg := range os.Args[1:] {
-			if arg == "-graceful" {
+			if strings.TrimLeft(arg, "-") == "graceful" {
 				break
 			}
 			args = append(args, arg)
 		}
 	}
-	args = append(args, "-graceful")
+	args = append(args, "--graceful")
 	if len(runningServers) > 1 {
-		args = append(args, fmt.Sprintf(`-socketorder=%s`, strings.Join(orderArgs, ",")))
+		args = append(args, fmt.Sprintf(`--socketorder=%s`, strings.Join(orderArgs, ",")))
 		log.Println(args)
 	}
 	cmd := exec.Command(path, args...)


flycash:
thank you!

flycash:
I will test it this week.

flycash:
I found that using "-graceful" and using "--graceful" are valid, and the help message still print:
```shell
  -graceful
    	listen on open fd (after forking)
  -socketorder string
    	previous initialization order - used when more than one listener was started
```
And then I found that Go does not follow the unix pattern that using double dash, so I think it's not very necessary to make this change. 😄

bo-er:
> I found that using "-graceful" and using "--graceful" are valid, and the help message still print:
> 
> ```shell
>   -graceful
>     	listen on open fd (after forking)
>   -socketorder string
>     	previous initialization order - used when more than one listener was started
> ```
> 
> And then I found that Go does not follow the unix pattern that using double dash, so I think it's not very necessary to make this change. 😄

Yes, if only beego is used there would be no problem.
If a project using both third party(like cobra) command-line parser and beego, then there is a problem. Currently I need to modify the code, using "--graceful" instead of "-graceful" AND adding "graceful" to cobra's flag set to make cobra and beego work together.

Thank you flycash for taking a glimsp into this issue, and I support either closing or merging this pull request. Keep up the great work! 👍

flycash:
Thank you

PR #5641, opened 2024-04-09
Feature: support custom accesslog format

moonsn:


flycash:
The beego already provide the feature `Formater`:
```go
type LogFormatter interface {
	Format(lm *LogMsg) string
}
```
I think users should use the interface. What scenario do you think that users may need this new feature?

moonsn:
the "access log" and "beego log" are different,  some times we don't need all the fields in `logs.AccessLogRecord`.

moonsn:
the "access log" and "beego log" are different. users cann't use the `LogFormatter` to custom the "access log"

flycash:
Thank you, I got your point. I will merge this new feature. 

But actually I think current design of `AccessLog` is terrible. I think we need to refactor this part.

At least, the AccessLog should belong to module `web`.

PR #5679, opened 2024-07-12
feat(update): update text at server running in localhost

ryuji-cre8ive:
The text output when running with bee run is as follows, and we felt it was not appropriate.
スクリーンショット 2024-07-12 18 17 54


flycash:
Thanks for your contribution. And please provide the result(i.e. screenshot or something) after this change.

PR #5685, opened 2024-07-29
Fix session concurrent

### LINKED ISSUE #5681: the session store has a write conflict issue Under concurrent requests
**English Only**. Please use English because others could join the discussion if they got similar issue!

1. What did you do?
I was building an SSO system with Casdoor, I found that user logins often did not take effect. After investigating the source code and testing, I discovered that Casdoor uses Beego’s session to cache the user’s login status. When the user logs out, the user data in the session is deleted, but if there are multiple concurrent requests at this time, the request that finishes later will rewrite the deleted data during SessionRelease, causing the user’s login status to be restored.

2. What did you expect to see?
If a session store has already deleted a key in one request, the key-value pair read by other requests should not be rewritten during the release phase.
3. What did you see instead?
> please provide log or error information.

4. How to reproduce the issue?

> or you can provide a reproduce demo.

5. What version of Go and beego are you using (`bee version`)?
v1.12.12
6. What operating system and processor architecture are you using (`go env`)?

XdpCs:
resolve #5681


XdpCs:
finish file, mem, cookie and mock.

XdpCs:
SessionReleaseIfPresent save session values to ledis when key is present
it is not supported now, because ledis has no feature like SETXX or atomic operation.
https://github.com/ledisdb/ledisdb/issues/251
https://github.com/ledisdb/ledisdb/issues/351

flycash (commented):
I think we can replace the mock/response with `httptest` package. And then we can remove this file.

flycash (approved):

PR #5688, opened 2024-08-03
feat(session): support v1

XdpCs:
support v1


flycash (approved):

flycash:
@lucklyt can you help to confirm this change if actually solve your problem?

PR #5691, opened 2024-08-07
Added support for `select` with `options` tag for `templatefun.RenderForm`

fahadnaeemkhan:
Added support for `select` with `options` tag for `templatefun.RenderForm`

# Example
```go
type Vnf struct {
	id       int
	Name     string `form:"name" required:"true" class:"w-full h-4 sm:h-9 border-b-2 border-gray-300 focus:border-blue-300 outline-none"`
	PublicIp string `form:"publicIp" required:"true" class:"w-full h-4 sm:h-9 border-b-2 border-gray-300 focus:border-blue-300 outline-none"`
	Version  string `form:"version,select" required:"true" class:"w-full h-4 sm:h-9 border-b-2 border-gray-300 focus:border-blue-300 outline-none"`
	Flavor   string `form:"flavor" required:"true" class:"w-full h-4 sm:h-9 border-b-2 border-gray-300 focus:border-blue-300 outline-none"`
}

vnf := Vnf{Flavor: "small", Version: "v0.0.1,v0.0.2,v2.0.0"}
c.Data["Form"] = beego.RenderForm(&vnf)
```
Screenshot 2024-08-07 at 7 11 24 PM

### DIFF
--- a/server/web/templatefunc.go
+++ b/server/web/templatefunc.go
@@ -26,6 +26,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/beego/beego/v2/core/logs"
 	"github.com/beego/beego/v2/server/web/context"
 )
 
@@ -290,10 +291,10 @@ func RenderForm(obj interface{}) template.HTML {
 	}
 	objT = objT.Elem()
 	objV = objV.Elem()
-
 	var raw []string
 	for i := 0; i < objT.NumField(); i++ {
 		fieldV := objV.Field(i)
+
 		if !fieldV.CanSet() || unKind[fieldV.Kind()] {
 			continue
 		}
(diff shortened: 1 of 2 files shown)


flycash:
sorry, I do think beego should not support any template function and I plan to remove this feature. In modern web application, you should use Vue, React or others to build your web application rather than Beego.

fahadnaeemkhan:
Hey Ming,

Thanks for getting back to me. I do agree that in most cases we should use frontend frameworks for this but this is very handy for small projects where we can quickly setup forms based on structs defined (just like `ORM`).

I'm coming from `Elixir/Phoenix` framework, there we have something very similar. This really increase the productivity as we don't have to define each form field and when we are reading back form data we can simply pass same struct to unmarshall it.

https://hexdocs.pm/phoenix_live_view/form-bindings.html

flycash:
can you add unit tests for this new feature?

fahadnaeemkhan:
Added tests as requested !!!

Issue #5695, opened 2024-08-19
Add Enum to validators


hamidrabedi:
add enum or valid values to validators, something like:
```go
type MFA struct {
	MFAType string `valid:"Required;Enum(sms,email,code)"`
}
```

so only those values are acceptable and any other values should return errors.

hamidrabedi:
if we could make it like it accepts functions, that would be amazing like:
```
type MFA struct {
	MFAType string `valid:"Required;Values(GetValidValues)"`
}

func (c *MFA) GetValidValues() []string{
    // now we can fetch data from everywhere
}

```

flycash:
@hamidrabedi hi, can you help to confirm if the #5697 meet your requirement. I think this design looks good.

The key point is that using different implementation to support different enum, so we will introduce two more validator:
- Enum
- IntEnum

PR #5697, opened 2024-08-19
add Enum string to validators

zhyclouds:
#5695 
Because "," will be split, I use "|" to split.
And the number of parameters will be compared in parseFunc, so I accept the whole snippet with string.
example: `valid:"Enum(sms|email|code)"`

### DIFF
--- a/core/validation/README.md
+++ b/core/validation/README.md
@@ -140,6 +140,7 @@ Struct Tag Functions:
 	Tel
 	Phone
 	ZipCode
+    Enum
 
 ## LICENSE
 
--- a/core/validation/validation.go
+++ b/core/validation/validation.go
@@ -245,6 +245,11 @@ func (v *Validation) ZipCode(obj interface{}, key string) *Result {
 	return v.apply(ZipCode{Match{Regexp: zipCodePattern}, key}, obj)
 }
 
+// Enum Test that the obj is in the specified enumeration if type is string
+func (v *Validation) Enum(obj interface{}, vals string, key string) *Result {
+	return v.apply(Enum{vals, key}, obj)
+}
+
 func (v *Validation) apply(chk Validator, obj interface{}) *Result {
 	if nil == obj {
 		if chk.IsSatisfied(obj) {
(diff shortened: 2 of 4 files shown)


flycash:
thank you!

hamidrabedi:
I think this has some problems @flycash @zhyclouds :
- I tested this with this piece of code that I use for all my validations:
```go
func ValidateStruct(input interface{}) (bool, []map[string]interface{}) {
	var errorDetails []map[string]interface{}

	passed, err := valid.Valid(input)
	if err != nil {
		logs.Error("Internal server error during validation" + err.Error())
		errorDetails = append(errorDetails, map[string]interface{}{
			"message": "Internal server error during validation",
			"error":   err.Error(),
		})
		return false, errorDetails
	}

	if !passed {
		for _, err := range valid.Errors {
			errorDetails = append(errorDetails, map[string]interface{}{
				"message": err.Message,
				"name":    err.Name,
				"field":   err.Label,
			})
		}
	}

	return passed, errorDetails
}
```
Normally it does not raise any errors and only `passed` var will be false and I can catch errors, but with Enums I'm getting this Error: `[E] [util.go:113]  Internal server error during validation doesn't exists Enum valid function`

- second thing, I don't think the error is appropriate, better to add the acceptable values to the output, something like: 
`auth_code doesn't exists Enum valid function message. allowed values are: sms|email|code`

- and if we had a way to define a function to get the acceptable values would be amazing. in many use cases you need to hit to database to fetch the acceptable values.

zhyclouds:
@hamidrabedi 
I apologise for not replying in time.
- I verified with your code snippet and did not get this error: `[E] [util.go:113]  Internal server error during validation doesn't exists Enum valid function`, I'm guessing it's possible that the source code you're using isn't pulling from my branch causing this error.
- I apologise for the definition of my error message, I've changed it like this: `MFAType Must be a string value in "sms|email|code"`, thank you very much for the suggestion!
- I can't define a function to get the method for acceptable values, this would modify existing source code instead of adding code to implement this, and could lead to some unintended risks, for which I apologise. I hope that one day developers will be able to implement this feature in an elegant way!

hamidrabedi:
@zhyclouds all good. this looks good to me. thank you.

flycash (approved):

flycash:
I will release it next month.

Issue #5698, opened 2024-08-22
add feature for custom validation 


hamidrabedi:
let users define custom functions and do custom validations:

```go
type MFA struct {
	MFAType string `valid:"Func(IsmfaValid)"`
}

func (c *MFA) IsmfaValid(input string) error {
	validMfa := []string{"email", "sms"}
	if slices.Contains(validMfa, input) {
		return nil
	}
	return fmt.Errorf("MFAType is not valid")
}
```

PR #5699, opened 2024-08-22
feat(validation): add Func validation to have the ability to do customized validations #5698

hamidrabedi:
#5698
I tried to do this with the least amount of code, and small changes on the main code

### DIFF
--- a/core/validation/util.go
+++ b/core/validation/util.go
@@ -298,3 +298,40 @@ func parseParam(t reflect.Type, s string) (i interface{}, err error) {
 func mergeParam(v *Validation, obj interface{}, params []interface{}) []interface{} {
 	return append([]interface{}{v, obj}, params...)
 }
+
+func GetMethod(i interface{}, methodName string) *reflect.Value {
+	var ptr reflect.Value
+	var value reflect.Value
+	var finalMethod reflect.Value
+
+	value = reflect.ValueOf(i)
+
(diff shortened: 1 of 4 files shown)


hamidrabedi:
I have no idea why tests are failing. and for deepSource, they are not even my code, I'm confused.

flycash:
> I have no idea why tests are failing. and for deepSource, they are not even my code, I'm confused.
you can ignore this.

But what issue does this merge request aim to solve?

hamidrabedi:
this one:
https://github.com/beego/beego/issues/5698

flycash:
I have another idea that we allow users define the specific function on the type:
```go
type User struct {

}

func (u User) Validate() error {
// users write the code here
}
```
And for Beego:
```go
// validate the field based on tag
validator.Validate()
// check if the specific instance has the function Validate()
if val, ok := obj.(interface{Validate() error}; ok{
   err = val.Validate()
} 
```

hamidrabedi:
i don't get it, you want to do only one validation for the whole struct? and not per field? or you want to have like both?

flycash:
never mind, I will find someone to implement this new feature and then you can help to review it and confirm if it solves your problem.

flycash:
https://github.com/beego/beego/pull/5703 we already support custom functions, you can find the example.

PR #5710, opened 2024-10-04
use sync.Once to replace lock

luxcgo:


client/orm/internal/models/models.go
```diff
@@ -28,7 +29,7 @@ type ModelCache struct {
 	orders          []string
 	cache           map[string]*ModelInfo
 	cacheByFullName map[string]*ModelInfo
-	done            bool
+	done            atomic.Bool
```
flycash, line 32: I think we can remove the sync.RWMutex. And can you try to use the `sync.Once`?

PR #5713, opened 2024-10-31
fixed #5573 ignore print to load conf/app.conf when not run web server

### LINKED ISSUE #5573: Cannot ignore failure to load conf/app.conf
Hello,
I used Beego and Cobra to build a command-line tool similar to Consul, which can be either a server or a client. This command will be registered in the system environment. When it acts as a server, it will read a configuration file under a fixed path to start the Beego web framework. But as a client, it does not require configuration files in a fixed directory, but when I execute a command, I always print  
````txt
[D] init global config instance failed' on the first line If you don't use this, just ignore it. open conf/app. conf: no such file or directory 
````
**I want to remove this log information for printing.**
I have noticed ISSU5537 and MR5538, which move the information originally printed in the log to STDERR. **However, when the command line displays both STDOUT and STDERR on the console, this command will still be printed before the output information, and it cannot be removed**. Therefore, there must be a mechanism that can be used by the caller to control the information printed in similar INIT functions.

lengpucheng:
Fix # 5573 , Although this problem was fixed to some extent in MR5538 by displaying it in STE, the content in STE will also be displayed at certain times, which will lead to some ambiguity in use.

For example, I quoted beego to open a command line tool. The web server will be enabled only when its startup command is server or agent. At other times, it will be used as a command line tool and registered in PATH and used anywhere in the system. At this time, the following will appear:

_The user executes a command in a folder in the system. The command is executed successfully and the content is displayed in the window. However, since there is no config file in the directory, the beego error will be printed in the first line._

_This will cause ambiguity in the user's use. Similarly, when the configuration file is placed in the etc directory and specified in the code, an error will still occur when it is started at other locations._

Therefore, the error is modified and cached this time, and it is only displayed when the web server is running.

### DIFF
--- a/core/config/ini.go
+++ b/core/config/ini.go
@@ -19,7 +19,6 @@ import (
 	"bytes"
 	"context"
 	"errors"
-	"fmt"
 	"io"
 	"os"
 	"os/user"
@@ -30,6 +29,12 @@ import (
 
 	"github.com/mitchellh/mapstructure"
 )
+
+// InitDefaultGlobalInstanceErr cache default ini config init error
+// when it as cli or only import some mod and not run web server or manually specify the configuration file in the code
+// The current default load will print errors directly in the console resulting in an unfriendly display
+// Here the error will be saved and lazy loaded, displayed if and only when it is run web server
+var InitDefaultGlobalInstanceErr error
 
 var (
 	defaultSection = "default"   // default section means if some ini items not in a section, make them in default section,
@@ -518,7 +523,7 @@ func init() {
 
 	err := InitGlobalInstance("ini", "conf/app.conf")
 	if err != nil {
-		fmt.Fprint(os.Stderr, "init global config instance failed. If you do not use this, just ignore it. ", err)
+		InitDefaultGlobalInstanceErr = err
 	}
 }
 
--- a/server/web/beego.go
+++ b/server/web/beego.go
@@ -15,6 +15,8 @@
 package web
 
 import (
+	"fmt"
+	"github.com/beego/beego/v2/core/config"
 	"os"
 	"path/filepath"
 	"sync"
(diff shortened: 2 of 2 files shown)


lengpucheng:
#5573

lengpucheng:
This is not a good solution because when you are executing console commands, such as wanting to view the current status, this exception will be printed on the first line, which will affect subsequent operations> I think it's not a good solution. How about changing to this?
> ```go
> slog.Debug(os.Stderr, "init global config instance failed. If you do not use this, just ignore it. ", err)
> ```

lengpucheng:
@flycash This issue should be fixed, and it would be unacceptable for  a CLI tool to if import the Beego library and display this message on the first line at runtime

flycash:
no, Beego needs to print the error information because many users rely on this behavior to confirm if they load the configure file correctly. 

For your scenarios, I don't think it's a big problem.

lengpucheng:
Thank you for your reply, but i deem that the error message printing of loading the default configuration file can be  **delayed until the web server is running**, so that the error message printing can be retained and the additional display caused by importing the beego library can be avoided. 
Sincerely hope you can adopt

flycash:
I do consider remove the default behavior but it's a break change, so let me think more about this.
I believe that all `init` function should never be used in Beego.

lengpucheng:
> 我确实考虑删除默认行为，但这是一个中断更改，所以让我更多地考虑一下。我认为 Beego 中永远不应该使用所有函数。`init`



> I do consider remove the default behavior but it's a break change, so let me think more about this. I believe that all `init` function should never be used in Beego.

I think the init function related to Register can be retained, but all the content of the init function that affects the execution should be moved to function of Web Run. For this example, this modification is used in a temporary workaround, of course, it is not a perfect modification method.

lengpucheng:
> I do consider remove the default behavior but it's a break change, so let me think more about this. I believe that all `init` function should never be used in Beego.

I think the init function related to Register can be retained, but all the content of the init function that affects the execution should be moved to function of Web Run. For this example, this modification is used in a temporary workaround, of course, it is not a perfect modification method.

PR #5725, opened 2024-11-30
fix #5724 ORM: support WhereRaw and WhereMap

### LINKED ISSUE #5724: ORM: qb Where support WhereRaw and WhereMap
As someone may notice that we plan to provide more APIs for ORM.

Currently, we have `Selector`, `Updater` and `Deleter`. Both of these three structures have the `Where` function：
```go
func (s *Selector[T]) Where(ps ...Predicate) *Selector[T] {
	s.where = ps
	return s
}

// example 
// SELECT * FROM `test_model` WHERE `id` < ?; [100]
NewSelector[TestModel](db).Where(C("Id").As("my_id").LT(100))
```
And you can image that this API looks better but it's not convenient.

The design of this API  was inspired by two frameworks:
- GORM, but most of its API take `any` as input looked not good;
- gdbc, the internal ORM framework developed by Shopee. But I don't think it's a good design and it's full of some inconvenient and terrible design.

So I prefer to combine these two frameworks and then design a better one.

Currently, I think we can add two more functions:
```go
function (s *Selector) WhereMap(conds map[string]any) {
    return s.Where(toPs(conds))
}

// WhereRaw("a=? AND b = ?", 1, 2)
function (s *Selector) WhereRaw(raw string, args ...any) {
    return s.Where(toPs(raw, args))
}
```

But I will keep the `Predicate` since I want to support the sharding feature in the future.

a1270107629:
fix #5724 ORM: support WhereRaw and WhereMap

### DIFF
--- a/client/orm/qb/select.go
+++ b/client/orm/qb/select.go
@@ -244,3 +244,18 @@ func (s *Selector[T]) Get(ctx context.Context) (*T, error) {
 	err = s.db.ReadRaw(ctx, t, q.SQL, q.Args...)
 	return t, nil
 }
+
+func (s *Selector[T]) WhereMap(conds map[string]any) *Selector[T] {
+	psArr := make([]Predicate, 0, len(conds))
+	for k, v := range conds {
+		ps := C(k).EQ(v)
+		psArr = append(psArr, ps)
+	}
+	return s.Where(psArr...)
+}
+
+// WhereRaw("a=? AND b = ?", 1, 2)
+func (s *Selector[T]) WhereRaw(raw string, args ...any) *Selector[T] {
+	return s.Where(Raw(raw, args...).AsPredicate())
+	//return s.Where(toPs(raw, args))
+}
--- a/client/orm/qb/select.go
+++ b/client/orm/qb/select.go
@@ -198,7 +198,7 @@ func (s *Selector[T]) From(table string) *Selector[T] {
 }
 
 func (s *Selector[T]) Where(ps ...Predicate) *Selector[T] {
-	s.where = ps
+	s.where = append(s.where, ps...)
 	return s
 }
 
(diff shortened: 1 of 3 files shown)


client/orm/qb/select.go
```diff
@@ -244,3 +244,18 @@ func (s *Selector[T]) Get(ctx context.Context) (*T, error) {
...
+		ps := C(k).EQ(v)
+		psArr = append(psArr, ps)
+	}
+	return s.Where(psArr...)
+}
+
+// WhereRaw("a=? AND b = ?", 1, 2)
+func (s *Selector[T]) WhereRaw(raw string, args ...any) *Selector[T] {
+	return s.Where(Raw(raw, args...).AsPredicate())
+	//return s.Where(toPs(raw, args))
```
flycash, line 260: removed

client/orm/qb/select.go
```diff
@@ -244,3 +244,18 @@ func (s *Selector[T]) Get(ctx context.Context) (*T, error) {
...
 	return t, nil
 }
+
+func (s *Selector[T]) WhereMap(conds map[string]any) *Selector[T] {
+	psArr := make([]Predicate, 0, len(conds))
+	for k, v := range conds {
+		ps := C(k).EQ(v)
+		psArr = append(psArr, ps)
+	}
+	return s.Where(psArr...)
```
flycash, line 254: Please help to make the change:
```go
func (s *Selector[T]) Where(ps ...Predicate) *Selector[T] {
	s.where = ps // s.where = append(s.where, ps)
	return s
}
```

flycash: I think this change makes the Where more convenient. And please help to add test for this change.

flycash: After making this change, you can change to:
```go
	for k, v := range conds {
		ps := C(k).EQ(v)
		s.Where(ps) // avoid memory allocation
	}
```

a1270107629: I solved the above problem. Add  test example for WhereRaw and WhereMap and Where, but find that WhereMap test might fail because the traversal of the map is unordered.

flycash: use one key-value pair. It's a simple function

flycash:
@a1270107629 can you help to add these functions to Deleter and Updater?

a1270107629:
yes i'd like to, I find the deleter, but I can't find the updater

flycash:
ok, you can just add the functions to Deleter.

a1270107629:
I can try to write an updater based on the existing code

flycash:
thank you !

PR #5734, opened 2024-12-22
WIP: refactor modify alias to DB

Stone-afk:


flycash (commented):
I think we should reuse the existing `DB` or `alias` instead of defining new `DB`

PR #5766, opened 2025-04-04
feature: add stream  event resp

Stone-afk:


server/web/context/context.go
```diff
@@ -77,6 +78,91 @@ type Context struct {
...
 }
 
+type StreamEvent struct {
+	Content          string
+	ReasoningContent string
+	Error            error
+	Done             bool
+}
+
+func (ctx *Context) EventStreamResp() chan<- StreamEvent {
```
flycash, line 88: return chan <- []byte
```go
select {
    case event, ok := dataCh:
         if !ok {
               return
         }
        // write to the resp and flush.
}
```
I think not only AI but also other scenarios may use this function.

server/web/context/context.go
```diff
@@ -77,6 +78,91 @@ type Context struct {
...
+					endEvent(ctx)
+					return
+				}
+
+				if event.Error != nil {
+					errorEvent(ctx, event.Error)
+					return
+				}
+				msgEvent(ctx, event)
+			case <-ctx.Request.Context().Done():
```
flycash, line 104: Please confirm whether this code path will be reached if the caller forgot to close the channel. if so, I think we don't need to require the caller must close the channel.

flycash (commented):
Don't forget to add tests.

server/web/context/context.go
```diff
@@ -77,6 +78,33 @@ type Context struct {
...
+				close(eventCh)
+				return
+			}
+		}
+	}()
+	return eventCh
+}
+
+func sendEvent(ctx *Context, data string) {
```
flycash, line 100: `data []byte`

server/web/context/context.go
```diff
@@ -77,6 +78,33 @@ type Context struct {
...
+	buf := bytes.Buffer{}
+	buf.WriteString(data)
+	buf.WriteByte('\n')
```
flycash, line 103: remove these

PR #5771, opened 2025-04-08
Added plugin support for generic types at the web.

junwense:
as title

### DIFF
--- a/server/web/webx/type.go
+++ b/server/web/webx/type.go
@@ -4,7 +4,7 @@ import (
 	"github.com/beego/beego/v2/server/web/context"
 )
 
-// WrapperFromJson is a wrapper function for handling JSON in request's body.
+// WrapperFromJson is a internalWrapper function for handling JSON in request's body.
 // usage:
 //
 //	web.Post("/hello", WrapperFromJson(func(ctx *context.Context, param T) (any, error) {
@@ -15,12 +15,12 @@ import (
 // See test cases for details
 func WrapperFromJson[T any](
 	biz func(ctx *context.Context, param T) (any, error)) func(ctx *context.Context) {
-	return wrapper(biz, func(ctx *context.Context, params *T) error {
+	return internalWrapper(biz, func(ctx *context.Context, params *T) error {
 		return ctx.BindJSON(params)
 	})
 }
 
-// WrapperFromForm is a wrapper function for handling form data in request.
+// WrapperFromForm is a internalWrapper function for handling form data in request.
 // usage:
 //
 //	web.Post("/hello", WrapperFromForm(func(ctx *context.Context, param T) (any, error) {
@@ -31,7 +31,7 @@ func WrapperFromJson[T any](
 // See test cases for details
 func WrapperFromForm[T any](
 	biz func(ctx *context.Context, param T) (any, error)) func(ctx *context.Context) {
-	return wrapper(biz, func(ctx *context.Context, params *T) error {
+	return internalWrapper(biz, func(ctx *context.Context, params *T) error {
 		return ctx.BindForm(params)
 	})
 }
@@ -47,12 +47,13 @@ func WrapperFromForm[T any](
 // See test cases for details
 func Wrapper[T any](
 	biz func(ctx *context.Context, param T) (any, error)) func(ctx *context.Context) {
-	return wrapper(biz, func(ctx *context.Context, params *T) error {
+	return internalWrapper(biz, func(ctx *context.Context, params *T) error {
 		return ctx.Bind(params)
 	})
 }
 
-func wrapper[T any](
+// internalWrapper is a core helper function that wraps the business logic and the binding logic.
+func internalWrapper[T any](
 	biz func(ctx *context.Context, param T) (any, error),
 	wf func(ctx *context.Context, params *T) error) func(ctx *context.Context) {
 	return func(ctx *context.Context) {
--- a/server/web/webx/type.go
+++ b/server/web/webx/type.go
@@ -1,70 +1,74 @@
 package webx
 
 import (
+	"github.com/beego/beego/v2/core/logs"
 	"github.com/beego/beego/v2/server/web/context"
+	"github.com/beego/beego/v2/server/web/session"
 )
 
+type BizFunc[T any] func(ctx *context.Context, param T) (any, error)
+type ExtractFunc[T any] func(ctx *context.Context) (params T, err error)
+
 // WrapperFromJson is a internalWrapper function for handling JSON in request's body.
-// usage:
-//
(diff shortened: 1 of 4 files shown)


server/web/webx/type.go
```diff
@@ -0,0 +1,76 @@
...
+// usage:
+//
+//	web.Post("/hello", WrapperFromForm(func(ctx *context.Context, param T) (any, error) {
+//		 return param, nil
+//	}))
+//
+// It binds the form data to the specified type T
+// See test cases for details
+func WrapperFromForm[T any](
+	biz func(ctx *context.Context, param T) (any, error)) func(ctx *context.Context) {
```
flycash, line 33: we don't need these two functions. I think users can use the Wrapper

server/web/webx/type.go
```diff
@@ -0,0 +1,76 @@
...
+	})
+}
+
+// Wrapper is use by beego ctx.Bind(any) api
+// usage:
+//
+//	web.Post("/hello", Wrapper(func(ctx *context.Context, param T) (any, error) {
+//		 return param, nil
+//	}))
+//
```
flycash, line 45: remove these lines. Currently you can write a ExampleXXX in test files to generate the docs.

server/web/webx/type.go
```diff
@@ -0,0 +1,76 @@
...
+
+// internalWrapper is a core helper function that wraps the business logic and the binding logic.
+func internalWrapper[T any](
+	biz func(ctx *context.Context, param T) (any, error),
+	wf func(ctx *context.Context, params *T) error) func(ctx *context.Context) {
+	return func(ctx *context.Context) {
+		var params T
+		err := wf(ctx, &params)
+		if err != nil {
+			ctx.Abort(400, err.Error())
```
flycash, line 63: Log error. And I think we should Abort(400, res) while serialize the res based on the response's `Content-Type` or Request's `Accept` header

flycash: For example, in most cases, the developers wants return the value:
```go
return {
   Code: 100010, // bussinese error code
   Msg: "Internal server error" and please retry,
  ...
}, errors.New("more detail but don't want the users know")
```

server/web/webx/type.go
```diff
@@ -0,0 +1,76 @@
...
+func Wrapper[T any](
+	biz func(ctx *context.Context, param T) (any, error)) func(ctx *context.Context) {
+	return internalWrapper(biz, func(ctx *context.Context, params *T) error {
+		return ctx.Bind(params)
+	})
+}
+
+// internalWrapper is a core helper function that wraps the business logic and the binding logic.
+func internalWrapper[T any](
+	biz func(ctx *context.Context, param T) (any, error),
```
flycash, line 57: here is a problem: how the develop get the session in biz?

server/web/webx/type.go
flycash: never mind, just keep these functions.

server/web/webx/type.go
junwense: > remove these lines. Currently you can write a ExampleXXX in test files to generate the docs.

this i dont know how to do it , like generate the docs. - -

server/web/webx/type.go
junwense: > For example, in most cases, the developers wants return the value:
> 
> ```go
> return {
>    Code: 100010, // bussinese error code
>    Msg: "Internal server error" and please retry,
>   ...
> }, errors.New("more detail but don't want the users know")
> ```

whether to return  this struct

server/web/webx/type.go
```diff
@@ -0,0 +1,102 @@
+package webx
+
+import (
+	"github.com/beego/beego/v2/core/logs"
+	"github.com/beego/beego/v2/server/web/context"
+	"github.com/beego/beego/v2/server/web/session"
+)
+
+type BizFunc[T any] func(ctx *context.Context, param T) (any, error)
+type ExtractFunc[T any] func(ctx *context.Context) (params T, err error)
```
flycash, line 10: make this type private

server/web/webx/type.go
```diff
@@ -0,0 +1,102 @@
...
+
+// WrapperFromJson is a internalWrapper function for handling JSON in request's body.
+// It binds the JSON request body to the specified type T
+// See test cases for details
+func WrapperFromJson[T any](
+	biz BizFunc[T],
+	opts ...Option[T]) func(ctx *context.Context) {
```
flycash, line 17: remove the opts... params, uses can do it manually

server/web/webx/type.go
```diff
@@ -0,0 +1,102 @@
...
+			holder.setSession(store)
+		}
+		return nil
+	}
+}
+
+type SessionHolder interface {
+	GetSession() session.Store
+	setSession(s session.Store)
+}
```
flycash, line 102: remove these, looks not good. may be we should think about refactor the Session module.

server/web/webx/type.go
```diff
@@ -0,0 +1,84 @@
...
+		return
+	})
+}
+
+// Wrapper is use by beego ctx.Bind(any) api
+// It binds the data to the specified type T
+// See test cases for details
+func Wrapper[T any](
+	biz bizFunc[T],
+	opts ...Option[T]) func(ctx *context.Context) {
```
flycash, line 44: remove `opts` from all functions. we don't need it. and then move the files to `server/web`, rename to `generic_wrapper`

server/web/webx/type.go
```diff
@@ -0,0 +1,84 @@
...
+	return internalWrapper(biz, func(ctx *context.Context) (params T, err error) {
+		err = ctx.Bind(&params)
+		return
+	}, opts...)
+}
+
+func internalWrapper[T any](
+	biz bizFunc[T],
+	ef extractFunc[T],
+	opts ...Option[T]) func(ctx *context.Context) {
```
flycash, line 54: remove

server/web/webx/type.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+// bizFunc is a function that wraps a business logic function with a context and parameter extraction function.
+type bizFunc[T any] func(ctx *context.Context, param T) (any, error)
+
+// extractFunc is a function that extracts parameters from the context.
+type extractFunc[T any] func(ctx *context.Context) (params T, err error)
+
+// Option options to T
+type Option[T any] func(ctx *context.Context, t *T) error
```
flycash, line 15: remove this type

flycash (commented):
write examples for these three functions.

flycash (commented):
Nice work!
And could you please check the docs [https://go.dev/blog/examples](https://go.dev/blog/examples) and write example for these three functions.

flycash (approved):

flycash:
image

PR #5807, opened 2025-06-17
WIP: add updater

Stone-afk:


client/orm/qb/update.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+func (u *Updater[T]) Build() (*Query, error) {
+	defer buffers.Put(u.buffer)
+	var err error
+	t := new(T)
+	if u.table == nil {
+		u.table = t
+	}
+	u.model, err = u.registry.GetOrRegisterByMd(&t)
```
flycash, line 34: t is a pointer, you don't need the & operation.

client/orm/qb/update.go
```diff
@@ -0,0 +1,195 @@
...
+		if has {
+			u.comma()
+		}
+		switch a := assign.(type) {
+		case Column:
+			c, ok := u.model.Fields.Fields[a.name]
+			if !ok {
+				return errs.NewErrUnknownField(a.name)
+			}
+			refVal, _ := u.val.Field(a.name)
```
flycash, line 120: define a new function on Field:
```go
func (fs *Field) ColumnName() {
    // ... don't forget to keep consistent with Beego ORM naming
}
```

client/orm/qb/update.go
```diff
@@ -0,0 +1,195 @@
...
+
+func isNilValue(val reflect.Value) bool {
+	switch val.Kind() {
+	case reflect.Chan, reflect.Func, reflect.Map, reflect.Ptr, reflect.UnsafePointer, reflect.Interface, reflect.Slice:
+		return val.IsNil()
+	}
+	return false
+}
+
+func (u *Updater[T]) SkipZeroValue() *Updater[T] {
```
flycash, line 167: Add the comments to specify the behaviors:
- basic type like int
- pointer

flycash: and a pointer is not nil but point to "zero value", for example *int = &(0)

client/orm/qb/valuer/reflect.go
```diff
@@ -0,0 +1,45 @@
...
+		val:  reflect.ValueOf(t).Elem(),
+		meta: model,
+	}
+}
+
+func (r *reflectValue) SetColumns(rows *sql.Rows) error {
+	return nil
+}
+
+// Field 返回字段值
```
flycash, line 26: english

PR #5883, opened 2026-03-10
special workaround for our unique scenario - should not go upstream

mrigankakb:


flycash:
The main concern is that the current toInt (or Int()) has a clear and strict semantic: parsing a string into an integer. This change introduces implicit support for floating-point values (via ParseFloat and truncation), which makes the method’s behavior less predictable and its intent less clear.

With this change, callers can no longer easily tell whether the method performs strict integer parsing or a more permissive numeric conversion. This kind of implicit behavior may lead to confusion and potential data issues (e.g., "1.9" being converted to 1).

If there is a real need to support this behavior, it would be better to implement it as a separate method with explicit semantics (e.g., ToIntLoose or ParseNumberToInt) at the call site, rather than extending the behavior of the existing method.

PR #5889, opened 2026-04-12
Propagate context.Context and Fix Stability Issues in Session Manager

hamid17amu:
This PR addresses several critical issues in the` session.Manager` regarding context propagation, latent crash risks due to `nil `context usage, and suboptimal resource management.

### 1. Context Propagation
- **Request Context Usage:** Updated `SessionStart`, `SessionDestroy`, and `SessionRegenerateID `to use` r.Context()` instead of hardcoded` context.Background()`. This ensures that session operations (especially those involving remote backends like Redis or SQL) respect request cancellation and support distributed tracing.
- **Updated `GetSessionStore `Signature:** Added `context.Context` as the first argument to `GetSessionStore(ctx context.Context, sid string)` to ensure consistent propagation throughout the API.
- **Eliminated `nil` Contexts:** Replaced `nil` context arguments in `GC()` and `GetActiveSession()` with `context.Background()` to prevent runtime panics in provider implementations.

### 2. Stability and Reliability
- **Panic to Error Conversion:** Modified `NewManager` to return formal `error` values instead of using `panic` during configuration validation (e.g., checking `SessionNameInHTTPHeader` format).
- **GC Loop Improvement:** Replaced the recursive `time.AfterFunc` logic in `GC()` with a `time.Ticker` running in a dedicated goroutine. This prevents potential stack growth and ensures a more stable background process.
- Cookie Unescape Validation: Added explicit error checking when unescaping session IDs in `SessionDestroy`. If an ID is malformed, it now logs the error via `SLogger` instead of silently failing or passing corrupted data to the provider.

### 3. General Cleanup
- **Logger Modernization:** Updated `NewSessionLog` to use standard `log.LstdFlags` for more informative timestamps and consistency with standard Go logging.
- **Variable Retention:** Maintained all original exported variable names and comments to ensure backward compatibility for users of the Beego session package.

### Impact
- **Observability**: Distributed tracing (OpenTelemetry/Jaeger) will now correctly track session store latency as the trace context is preserved.
- **Resource Management:** Sessions will stop processing immediately if a client disconnects during a long-running provider operation.
- **Robustness:** Eliminated "latent crash" scenarios where a provider attempting to use a `nil` context would trigger a service-wide panic.

### Testing Done
- Verified that all `Provider `interface calls are guarded by non-nil contexts.
- Validated that `r.Context()` is preferred whenever a request object is available.

Resolves #5878

### DIFF
--- a/server/web/session/session.go
+++ b/server/web/session/session.go
@@ -118,13 +118,13 @@ func NewManager(provideName string, cf *ManagerConfig) (*Manager, error) {
 
 	if cf.EnableSidInHTTPHeader {
 		if cf.SessionNameInHTTPHeader == "" {
-			panic(errors.New("SessionNameInHTTPHeader is empty"))
+			return nil, errors.New("SessionNameInHTTPHeader is empty")
 		}
 
 		strMimeHeader := textproto.CanonicalMIMEHeaderKey(cf.SessionNameInHTTPHeader)
 		if cf.SessionNameInHTTPHeader != strMimeHeader {
 			strErrMsg := "SessionNameInHTTPHeader (" + cf.SessionNameInHTTPHeader + ") has the wrong format, it should be like this : " + strMimeHeader
-			panic(errors.New(strErrMsg))
+			return nil, errors.New(strErrMsg)
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


hamid17amu:
@flycash Please review

